PR #2, opened 2020-08-28
feat: implement detection mode

xopham:
The detection mode allows Connaisseur to run in non-strict mode, letting non-compliant images pass, but logging the error. As of Kubernetes version 1.19, a warning is issued to the user (https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/).

### DIFF
--- a/connaisseur/admission_review.py
+++ b/connaisseur/admission_review.py
@@ -3,7 +3,11 @@ import json
 
 
 def get_admission_review(
-    uid: str, allowed: bool, patch: list = None, status_id: int = None, msg: str = None
+    uid: str,
+    allowed: bool,
+    patch: list = None,
+    msg: str = None,
+    detection_mode: bool = False,
 ):
     """
     Get a standardized response object with patching instructions for the
@@ -19,8 +23,6 @@ def get_admission_review(
         A list with JSON patch instruction, that will modify the original
         request, send to the Admission Controller. The list is Base64
         encoded.
-    status_id : int (optional)
-        The HTTP status code send back with the response.
     msg : str (optional)
         The error message, which will be displayed, should allowed be
         'False'.
@@ -40,6 +42,7 @@ def get_admission_review(
                 "code": 200,
                 "message": "All gucci, my boi."
             },
+            "warnings": ["detection_mode ON"]
             "patchType": "JSONPatch",
             "patch":
                 "W3sib3AiOiAiYWRkIiwgInBhdGgiOiAiL3NwZWMvcmVwbGljYXMiLCAidmFsdWUiOiAzfV0="
@@ -49,16 +52,17 @@ def get_admission_review(
     review = {
         "apiVersion": "admission.k8s.io/v1beta1",
         "kind": "AdmissionReview",
-        "response": {"uid": uid, "allowed": allowed},
+        "response": {
+            "uid": uid,
+            "allowed": allowed or detection_mode,
+            "status": {"code": 202 if allowed or detection_mode else 403},
+        },
     }
 
-    if status_id or msg:
-        if not status_id:
-            status_id = 202 if allowed else 403
-        status = {"code": status_id}
-        if msg:
-            status["message"] = msg
-        review["response"]["status"] = status
+    if msg:
+        review["response"]["status"]["message"] = msg
+        if detection_mode and not allowed:
+            review["response"]["warnings"] = [msg]
 
     if patch:
         review["response"]["patchType"] = "JSONPatch"
--- a/connaisseur/exceptions.py
+++ b/connaisseur/exceptions.py
@@ -1,4 +1,7 @@
 # pylint: disable=dangerous-default-value
+import os
+
+
 class BaseConnaisseurException(Exception):
     """
     Base exception that can take an error message and context information as a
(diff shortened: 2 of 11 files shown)


connaisseur/admission_review.py
```diff
@@ -60,6 +62,9 @@ def get_admission_review(
             status["message"] = msg
         review["response"]["status"] = status
 
+    if warnings:
```
peterthomassen, line 65: `if warnings is not None:` (if an empty list is passed explicitly, we should honor it). Alternatively, the default should be the empty list (and then the `if` condition is fine).

connaisseur/exceptions.py
```diff
@@ -7,14 +10,28 @@ class BaseConnaisseurException(Exception):
...
 
     def __str__(self):
-        return str({"message": self.message, "context": self.context})
+        return str(
+            {
+                "message": self.message,
+                "context": self.context,
+                "detection_mode": self.detection_mode,
+            }
+        )
```
peterthomassen, line 28: consider `return str(self.__dict__)`

connaisseur/exceptions.py
```diff
@@ -7,14 +10,28 @@ class BaseConnaisseurException(Exception):
...
+
+    def get_user_msg(self):
+        return (
+            f"{self.message}"
+            f"{' (not denied due to DETECTION_MODE)' if self.detection_mode else ''}"
+        )
```
peterthomassen, line 34: overcomplicated string operations.

```
def get_user_msg(self):
    msg = self.message
    if self.detection_mode:
        msg += ' (not denied due to DETECTION_MODE)`
    return msg
```

connaisseur/exceptions.py
```diff
@@ -7,14 +10,28 @@ class BaseConnaisseurException(Exception):
...
-        return str({"message": self.message, "context": self.context})
+        return str(
+            {
+                "message": self.message,
+                "context": self.context,
+                "detection_mode": self.detection_mode,
+            }
+        )
+
+    def get_user_msg(self):
```
peterthomassen, line 30: should be a property:

```
@property
def user_msg(self):
    ...
```

connaisseur/flask_server.py
```diff
@@ -26,24 +28,29 @@ def mutate():
         response = admit(admission_request)
     except BaseConnaisseurException as err:
         logging.error(str(err))
-        msg = err.message
+        msg = err.get_user_msg()
```
peterthomassen, line 31: if `BaseConnaisseurException.get_user_msg()` is turned into a property as proposed above, this becomes `msg = err.user_msg` (no parentheses). Also, this assignment is unnecessary; you can fetch the message a few lines down where you actually need it.

connaisseur/admission_review.py
```diff
@@ -60,6 +62,9 @@ def get_admission_review(
...
+        review["response"]["warnings"] = warnings
```
peterthomassen, line 66: I think this list would be passed by reference, so whoever receivers the returned `review` can alter the list that is passed into this method!

Solution: make sure to make a copy of the list before you use it, e.g. `warnings = warnings[:]` at the beginning of the method. (This would also favor the default argument to be an empty list, so you don't need to worry about the `None` case.)

connaisseur/flask_server.py
```diff
@@ -26,24 +28,29 @@ def mutate():
...
+                warnings=[msg] if DETECTION_MODE else None
+            )
         )
     except UnknownVersionError as err:
         logging.error(str(err))
         return jsonify(
             get_admission_review(
-                admission_request["request"]["uid"], False, msg=str(err)
+                admission_request["request"]["uid"], DETECTION_MODE, msg=str(err),
+                warnings=[msg] if DETECTION_MODE else None
```
peterthomassen, line 43: is it intended that we pass `err` into `msg`, but `msg` into `warnings`?

peterthomassen (changes requested):
Very nice work! A few comments:

- I added some inline comments, mainly on code style.
- Tests for the new functionality are missing. We should include them to maintain code coverage.
- The two commits are part of the same "atomic" thing and should be squashed.

Also, I think that we should start using semantic commit titles at this point (makes life easier because the commit history style will look familiar to many; also facilitates generating release notes in the future). See:
- https://www.conventionalcommits.org/en/v1.0.0/
- https://nitayneeman.com/posts/understanding-semantic-commit-messages-using-git-and-angular/

peterthomassen:
... can you please add corrections as fixup commits (`git commit --fixup`), and squash everything once approved? That makes the next review iteration much easier. Also, let's consider signing all commits. :)

connaisseur/tests/test_admission_review.py
```diff
@@ -50,32 +60,34 @@ def adm_rev():
...
-        (1, True, None, None, "Well hello there.", admission_review_msg),
+        (1, True, None, None, None, False, admission_review_plain),
+        (1, True, None, None, "Well hello there.", False, admission_review_msg),
         (
             1,
             True,
             [{"op": "add", "path": "/spec/replicas", "value": 3}],
             None,
-            None,
+            None, False,
```
peterthomassen, line 72: looking around, it seems like there should be one item per line here, for consistency. Consider using the Black code formatter.

connaisseur/tests/test_admission_review.py
```diff
@@ -50,32 +60,34 @@ def adm_rev():
...
             admission_review_patch,
         ),
         (
             1,
             False,
             [{"op": "add", "path": "/spec/replicas", "value": 3}],
             None,
-            "Well hello there.",
+            "Well hello there.", False,
```
peterthomassen, line 80: looking around, it seems like there should be one item per line here, for consistency. Consider using the Black code formatter.

peterthomassen (changes requested):

peterthomassen (approved):
lgtm! (don't forget to squash and rebase)

peterthomassen (approved):

PR #5, opened 2020-09-11
fix: added subjectAltNames to TLS certificates

### LINKED ISSUE #4: TLS Certificates Fix
The TLS certificates for connaisseur have subjectAltNames defined for them in the CSR config file, but they stll don't appear in the resulting certificates. This is caused by a known OpenSSL [bug](https://www.openssl.org/docs/man1.1.0/man1/x509.html#BUGS). These subjectAltNames are mandatory in Kubernetes v1.19 and need to be added or connaisseur will break.

phbelitz:
Kubernetes v1.19 requires subjectAltNames to be present on the TLS certificates or otherwise will break. Even though the TLS config file already added those, an OpenSSL bug prevented them to be actually propagated to the actual certificates.

closes #4

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -16,5 +16,4 @@ testing-env/aks/k8s/deployment.yaml
 .scannerwork/
 samples/*
 demo/
-ca.conf
-tls.conf
+tls-*.conf
--- a/helm/certs/gen_certs.sh
+++ b/helm/certs/gen_certs.sh
@@ -4,40 +4,17 @@ SCRIPT_PATH=$(dirname $0)
 SERVICE_NAME=$(yq r $SCRIPT_PATH/../Chart.yaml name)
 
 cd $SCRIPT_PATH
-openssl genrsa -out ca.key 4096
-
-cat <<EOF > ca.conf
-[ req ]
-default_bits       = 4096
-default_md         = sha512
-default_keyfile    = ca.key
-prompt             = no
-encrypt_key        = yes
-
-# base request
-distinguished_name = req_distinguished_name
-
-# distinguished_name
-[ req_distinguished_name ]
-countryName            = "DE"
-stateOrProvinceName    = "Berlin"
-localityName           = "Berlin"
-organizationName       = "SSE"
-organizationalUnitName = "Engineering Department"
-commonName             = "connaisseur.internal"
-emailAddress           = "no-reply@securesystems.de"
-EOF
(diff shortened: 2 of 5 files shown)


.gitignore
```diff
@@ -17,4 +17,4 @@ testing-env/aks/k8s/deployment.yaml
 samples/*
 demo/
 ca.conf
```
peterthomassen, line 19: is this line still needed?

helm/certs/gen_certs.sh
```diff
@@ -47,5 +38,6 @@ DNS.2 = $SERVICE_NAME-svc.$SERVICE_NAME
 DNS.3 = $SERVICE_NAME-svc.$SERVICE_NAME.svc
 DNS.4 = $SERVICE_NAME-svc.$SERVICE_NAME.svc.cluster.local
 EOF
-openssl req -new -key tls-key.pem -subj "/CN=$SERVICE_NAME-svc.$SERVICE_NAME.svc" -out tls.csr -config tls.conf
-openssl x509 -req -in tls.csr -CA ca.crt -CAkey ca.key -CAcreateserial -out tls-crt.pem
+
+openssl req -new -key tls.key -sha256 -config tls-csr.conf -subj "/CN=$SERVICE_NAME-svc.$SERVICE_NAME.svc" -out tls.csr
+openssl x509 -req -days 365 -in tls.csr -signkey tls.key -sha256 -extensions v3_ca -extfile tls-crt.conf -out tls.crt
```
peterthomassen, line 43: Is 365 days enough / is there a rotation mechanism?

peterthomassen: Also, are we signing with the same key that the certificate is for? Please check with @Starkteetje if that's ok

peterthomassen (approved):
lgtm, modulo validity period and @Starkteetje's feedback on the signing key setting

helm/certs/gen_certs.sh
Starkteetje: It's supposed to be a self-signed cert, right? In that case that's fine

peterthomassen (approved):
lgtm, although Connaisseur will break unexpectedly after 10 years. Why not set it to 100 years?

Also, one conversation is still open, but it's not so important.

peterthomassen:
oh, and next time please add a fixup commit (`git commit --fixup=...`) so that can review the changes more easily, and then squash after approval.

peterthomassen (approved):

Issue #6, opened 2020-09-23
Support for Nexus 


gilaiz:
I am using Nexus instead of Harbor this means that when I am trying to pull an image the command will look like this:
"pull nexus.example.com:18079/hello-world:test"
I am getting the following error:
Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: "nexus.example.com:18079/hello-world:test" is not a valid image format. 
from what I understand the problem is with the validation of the path format. Is there a way to make Connaisseur work with Nexus?

annekebr:
Thanks a lot for your feedback! We try to keep Connaisseur as compatible as we possible, so such tests in other platforms help a lot!

It seems that the regex used for the [image name validation](https://github.com/sse-secure-systems/connaisseur/blob/2b4e5314d1a11174a069cf1e6345952af3162b26/connaisseur/image.py#L30) does currently not allow ports. We are working on an MR to fix the image validation and will make an update soon!

gilaiz:
Thank you for the update.
Do you know when the update will be available? this is very urgent.

annekebr:
I pushed a proposed fix that is neither fully tested nor fully reviewed yet. This will be merged monday at the earliest. 

If you want to test it in advance, you should be able to test it by pulling the `fix-image-validation` branch, building and pushing the new test docker image to your registry, changing the kubectl pull secret to point to your own registry and changing the reference [https://github.com/sse-secure-systems/connaisseur/blob/fix-image-validation/helm/values.yaml#L4](here).

We'd appreciate your feedback if it works before we can test ourselves!

annekebr:
Hey @gilaiz! We merged the proposed fix and the input validation should allow a custom port for your nexus registry by now. 

Feel free to reopen in case you face any issues!

PR #7, opened 2020-09-25
fix: allow registry with custom port

annekebr:
The regex validation for image names does not allow specifying a port in the registry name. This commit extends the domain_re regex to allow ports.

Relates to #6

### DIFF
--- a/connaisseur/image.py
+++ b/connaisseur/image.py
@@ -26,8 +26,8 @@ class Image:
     digest: str
 
     def __init__(self, image: str):
-        # e.g. example.com, super.example.com
-        domain_re = r"(?:[a-z0-9-]{1,63}\.){1,62}[a-z0-9-]{1,63}"
+        # e.g. example.com, super.example.com:3498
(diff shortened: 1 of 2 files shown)


peterthomassen (commented):
Neat!

The character group corresponding to `domain_re` is stored in `Image.registry`, which is again used in the following piece of code to construct the Notary `url` from which to fetch the trust data:

https://github.com/sse-secure-systems/connaisseur/blob/2b4e5314d1a11174a069cf1e6345952af3162b26/connaisseur/notary_api.py#L181-L190

Here, the registry is embedded in the URL path. Do we know that specifying the port there is supported by Notary?

---

Also, please change the commit title to begin with `fix:`, cf. https://www.conventionalcommits.org/en/v1.0.0/.

phbelitz (approved):
lgtm

peterthomassen:
@annekebr please also change the commit title (not only the PR title)

phbelitz (approved):
all good

PR #14, opened 2020-10-07
docs: Add guide for cluster with Docker Hub

Starkteetje:
The current setup guide is complicated and requires setup of Harbor and minikube. We assume the average developer looking at Connaisseur has a test cluster available and is fine with testing on Docker Hub for a start. Then this guide (for EKS in particular) is much shorter and focusses on the Connaisseur aspects of image integrity.


examples/eks/README.md
```diff
@@ -0,0 +1,226 @@
...
+## Starting and testing a deployment of Connaisseur
+
+### 1. Set up environment
+
+First off, we'll export the variables used in this tutorial to allow you to immediately use your own names/URLs instead of relying on our default names. Below, substitute your own values as appropriate:
+
+```bash
+REGISTRY_URL=docker.io
+REPOSITORY_NAME=testingconny
+IMAGE_NAME=testserver
```
xopham, line 20: I would recommend using `testimage` as default value. `testserver` felt confusing to me at first sight.

examples/eks/README.md
```diff
@@ -0,0 +1,226 @@
...
+LAST_LINE=$(grep -nF -- '-----END PUBLIC KEY-----' helm/values.yaml | sed s/:.*//)
+sed -i $((${FIRST_LINE} + 1)),${LAST_LINE}d helm/values.yaml
+
+cd ~/.docker/trust/private
+sed '/^role:\sroot$/d' $(grep -iRl "role: root" .) > root-priv.key
+openssl ec -in root-priv.key -pubout -out root-pub.pem
+sed -i "${FIRST_LINE}s/.*/  rootPubKey: |\n    $(sed ':a;N;$!ba;s/\n/\\n    /g' root-pub.pem)/" ${OLDPWD}/helm/values.yaml
+rm root-priv.key root-pub.pem
+cd -
+```
```
xopham, line 69: One could set Conny to detection mode which would make sure to not break anything, but might make seeing the outcome of using Conny less explicit (either warning or nothing).

xopham, line 69: Might make a comment that this great for testing conny on a test cluster with test creds. For production use, one should look into the other options to manage the notary credentials and consider the non-strict detection mode. Might also resolve the comment above.

examples/eks/README.md
```diff
@@ -0,0 +1,226 @@
...
+
+### 4. Deploy Connaiseur
+
+After having completed the configuration, deploying Connaisseur is easy. Run
+
+```bash
+make install
+```
+
+after setting your Kubernetes context to the correct cluster. Your output should look something like
```
xopham, line 81:
```suggestion
Deploying Connaisseur is easy once the configuration is completed. Set your Kubernetes context to the correct cluster and run:

```bash
make install
```

Your output should look something like
```

File: examples/eks/README.md [hunk 4]
```diff
@@ -0,0 +1,226 @@
...
+
+NOTARY_URL=notary.docker.io
+NOTARY_USER=<YOUR-NOTARY-USERNAME/DOCKER-HUB-ID>
+NOTARY_PASSWORD=<YOUR-NOTARY-PASSWORD>
+
+git clone https://github.com/sse-secure-systems/connaisseur.git
+cd connaisseur
+```
+
+### (Optional) 2. Prepare test images
```
xopham, line 31: I see you put this here, since we need the DCT root key later. Technically, it interrupts the flow and would be much better placed at the end before deploying the image. Might enforce creation of a DCT key at this point

examples/eks/README.md
```diff
@@ -0,0 +1,226 @@
...
+spec:
+  ports:
+  - name: port
+    port: 8080
+    targetPort: 5000
+  selector:
+    app: sample
+  type: LoadBalancer
+EOF
+```
```
xopham, line 208: this and the following could technically also be example files, right?

xopham (commented):
Really like the guide! Very concise :+1: This will be very helpful. I had 4 more questions/recommendations/comments:
1. Is there any EKS specifics? It might make sense to point those out.
2. Could you make a note referring to this guide? Either in the README.md or Setup guide. Might be the first bullet of a list...
3. Since you created the sample files (which I really like), should we update the central setup guide to use those as well?
4. Ultimately, in case we see that the central part is always the same, we could also consolidate the whole thing to avoid duplication and only indicate the differences.

examples/eks/README.md
Starkteetje: I wouldn't do that (by default). But since it is a nice feature, I would be inclined to mention it, now that you mention it (hehe). Like "hey if y'all wanna use non-blocking mode, you could also run "this command here", which is nice for checking at first but just doesnt give you as much sec, so y'all probably dont wanna do that in prod long-term.
That fine by you?

Starkteetje: Again, I would be hesitant to recommend non-strict for more than testing. Can add a comment around cred handling though. Do also have a different option to putting the key in this conf file, like deploying a secret to the namespace (haven't really gone deep into conny)? Maybe also @phbelitz can give a bit of guidance here for how to best manage these

xopham: Would go for smth like that. Like "this could be useful for testing or rolling out", but wont block the ima just give a log message/warning

examples/eks/README.md
Starkteetje: technically yes, but then you do need envsubst to substitute the IMAGE_PATH variable and envsubst is actually not installed by default and iirc needs quite a few dependencies. All the other tooling, git, make, kubectl, sed, grep, ... are either preinstalled on most systems or can be reasonably suspected to be on the machine of whoever tries the tutorial. Envsubst isnt, so I wouldnt do it

Starkteetje:
@xopham Since Github won't allow me to reply to your last comment (for whatever reason):
To 1: Not really, I also do not expect any differences for the different clouds. I've now made that a bit more transparent in a top-level README, but would only explicitly write that after testing it (and then change e.g. folder name...)
To 2: Done
To 3: Agreed. Would you do that with this PR or separately? I'd kinda go for separate, because I wanted to revisit the minikube setup anyways later-on
To 4: Some parts will be the same, but at least my vision for final state would be something like: local (minikube + harbor), "standard" k8s (should work on all clouds and local installs, "abstracts" registry and notary by going for DockerHub) and ACR (AKS + ACR, Azure specific). In that case, the differences in the guides would (hopefully?) be substantial enogh to warrant the different READMEs

examples/eks/README.md
Starkteetje: Also tbh I like the fact that we show what's being run on k8s, instead of `kubecl apply -f thisfilewhichmayrunarbitraryprivilegeddeployments.yaml`. For the Dockerfiles it's not that important, since it will by default be run in the very deployment defined in this piece of code and thus can be considered somewhat sandboxed

xopham:
> @xopham Since Github won't allow me to reply to your last comment (for whatever reason):
> To 1: Not really, I also do not expect any differences for the different clouds. I've now made that a bit more transparent in a top-level README, but would only explicitly write that after testing it (and then change e.g. folder name...)
> To 2: Done
> To 3: Agreed. Would you do that with this PR or separately? I'd kinda go for separate, because I wanted to revisit the minikube setup anyways later-on
> To 4: Some parts will be the same, but at least my vision for final state would be something like: local (minikube + harbor), "standard" k8s (should work on all clouds and local installs, "abstracts" registry and notary by going for DockerHub) and ACR (AKS + ACR, Azure specific). In that case, the differences in the guides would (hopefully?) be substantial enogh to warrant the different READMEs

100% agree to both goal and approach, e.g. consolidate later! Sounds like a good plan :+1:

Starkteetje:
Regarding 1: Since I now got the time to test it also with minikube, I've updated the description and folder names to be more general and then only stated that it was tested on EKS and minikube.

setup/README.md
```diff
@@ -1,419 +1,21 @@
-# A Connaisseur Setup Guide
+# Setup
 
-## Intro
+Setting up Connaisseur for the first time can be a bit of a hassle, especially if you are not familiar with Docker Content Trust (DCT). Hence, we've created multiple guides for various environments to help you on the way.
```
xopham, line 3:
```suggestion
Setting up Connaisseur for the first time can be a bit of a hassle, especially if you are not familiar with Docker Content Trust (DCT). Furthermore, Connaisseur requires/links several other services (Kubernetes cluster, container registry, notary server) for which many common providers exist. Hence, we've created multiple guides for various environments to help you on the way.

The guide below offers a simple default configuration for setting up Connaisseur using public infrastructure and is aimed at quick testing. It uses [Docker Hub](https://hub.docker.com/) as both container registry and notary, but is expected to work for other managed solutions as well. It has been tested for the following Kubernetes services:
- [x] [minikube](https://github.com/kubernetes/minikube)
- [x] [Azure Kubernetes Service (AKS)](https://docs.microsoft.com/en-us/azure/aks/)
- [x] [Amazon Elastic Kubernetes Service (EKS)](https://docs.aws.amazon.com/eks/)
- [ ] [Google Kubernetes Engine](https://cloud.google.com/kubernetes-engine/docs/)
- [ ] [SysEleven MetaKube](https://docs.syseleven.de/metakube/de)

Two further specialized guides exist:
1. **Local setup ([Minikube](https://github.com/kubernetes/minikube) & [Harbor](https://github.com/goharbor/harbor))**: Connaisseur can be run completely locally with a minikube instance paired with a local Harbor installation that combines registry and notary servers. This setup is a bit more cumbersome since it does not use existing public infrastructure such as PKI. However, it is entirely based on open-source solutions, does not require continuous access to the internet and you control every part of the installation. This makes it specifically suited e.g. for security researchers and users interested in the inner workings of DCT and Connaisseur. Head over to [the respective README](local/README.md) to try it out.
2. TODO: **Azure [AKS](https://docs.microsoft.com/en-us/azure/aks/) with [Azure Container Registry (ACR)](https://docs.microsoft.com/en-us/azure/container-registry/)**: Microsoft Azure is currently the only one of the big cloud providers with a managed Kubernetes and container registry that supports DCT.
```
Explains why we offer separate guides, which ones were tested and other approaches without making this too deep a guide. I would then drop the other headings below and put the 'EKS Guide' directly in here below. This makes the setup guide flatter.

setup/cluster/README.md
```diff
@@ -0,0 +1,228 @@
+# Connaisseur with existing Kubernetes cluster
```
xopham, line 1: this guide could go into `setup/README.md`

setup/cluster/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+### 1. Set up environment
+
+First off, we'll export the variables used in this tutorial to allow you to immediately use your own names/URLs instead of relying on our default names. Below, substitute your own values as appropriate:
+
+```bash
+REGISTRY_URL=docker.io
+REPOSITORY_NAME=testingconny
+IMAGE_NAME=testimage
+IMAGE_PATH=$(echo ${REGISTRY_URL}/${REPOSITORY_NAME}/${IMAGE_NAME})
```
xopham, line 21: I think you only ever use the `IMAGE_PATH` and thus I would only set that. Otherwise, it makes it seem overly complicated. Maybe make a comment to show a sample entry.

setup/cluster/README.md
```diff
@@ -0,0 +1,228 @@
...
+
+During the second push, you will be asked to create a root key password and a repository key password (unless you already signed images or pushed signed images to the repository, respectively).
+
+### 3. Configure Connaisseur
+
+Before you can finally set up Connaisseur, you will need to configure its connection to the notary server provided by Docker Hub (or your notary server of choice).
+
+> At this point, Connaisseur only supports BasicAuth for connecting to the notary.
+
+Below is a list of commands to programmatically do the changes to the `value.yaml` file in the `helm` folder. The openssl command will take a password as input, so you won't be able to copy+paste past it. However, you can also do these changes manually if you want. They are comprised of setting the notary server, notary credentials and the root publiy key for verification. If you did not do step 2, you will need to enter your registry's root key from a (potentially) different source.
```
xopham, line 52:
```suggestion
Below is a list of commands to programmatically do the changes to the `values.yaml` file in the `helm` folder. The openssl command will take a password as input, so you won't be able to copy+paste past it. However, you can also do these changes manually if you want. They are comprised of setting the notary server, notary credentials and the root publiy key for verification. If you did not do step 2, you will need to enter your registry's root key from a (potentially) different source.
```
I am not certain what you mean that I will not be able to 'copy paste past it'. Does that mean if I copy the full code and run it in my shell that it fails bc it does not wait for the pw for OpenSSL? If so, I would split it in 2 code sections to avoid confusion. Maybe mention afterwards that a `git diff` will help to understand the changes implemented at this point.

setup/cluster/README.md
```diff
@@ -0,0 +1,228 @@
...
+```
+
+You should see the deployment being rejected with an error similar to
+
+```bash
+service/sample created
+Error from server: error when creating "STDIN": admission webhook "connaisseur-svc.connaisseur.svc" denied the request: could not find signed digest for image "docker.io/testingconny/testimage:unsigned" in trust data.
+```
+
+> Note that the service was indeed created, because it does not reference an image and is thus not blocked by Connaisseur. You can clean it up by calling `kubectl delete service sample`.
```
xopham, line 166:
```suggestion
> Note that while the container is blocked Kubernetes still creates the service, because it does not reference an image and is thus not denied by Connaisseur. You can clean it up by calling `kubectl delete service sample`.
```
otherwise it reads a bit like Conny is not working

xopham (commented):
It is a very neat guide. Very concise! I like it a lot and would therefore explicitly make it THE central guide as by my comments.

setup/local/README.md
```diff
@@ -0,0 +1,419 @@
...
+
+What about image integrity, making sure that running applications don't do some bitcoin mining in addition to or instead of showing your *"Hello World!"* page? No? Well then let's change that!
+
+Entering the scene – **Connaisseur**. It runs as a Mutating Admission Webhook that intercepts incoming requests and changes all images to their signed versions, if present. Otherwise Connaisseur denies the request. To do so, Connaisseur requires a **Notary server**, which is an implementation of *The Update Framework*. Notary stores the actual signatures and relevant metadata of the tag-to-digest relation.
+
+All you need to do is set those two up and you get image integrity for your Kubernetes cluster. This write-up shall give you a practical setup guide that shows how to install Notary and Connaisseur in your cluster and how to sign all your images.
```
xopham, line 15:
```suggestion
# Local Setup Guide
This guide describes setting up a completely local installation of Connaisseur using minikube paired with a local Harbor installation that combines registry and notary servers. This setup is a bit more cumbersome since it does not use existing public infrastructure such as PKI. However, it is entirely based on open-source solutions, does not require continuous access to the internet and you control every part of the installation. This makes it specifically suited e.g. for security researchers and users interested in the inner workings of DCT and Connaisseur.

>  Note: For quick testing, we recommend using the simple main [setup guide](setup/README.md) using Docker Hub.
```

setup/cluster/README.md
Starkteetje: > Does that mean if I copy the full code and run it in my shell that it fails bc it does not wait for the pw for OpenSSL?
yes
Can change it

setup/cluster/README.md
```diff
@@ -0,0 +1,228 @@
...
+cd setup
+docker build -f Dockerfile.unsigned -t ${IMAGE_PATH}:unsigned .
+docker push ${IMAGE_PATH}:unsigned
+docker build -f Dockerfile.signed -t ${IMAGE_PATH}:signed .
+DOCKER_CONTENT_TRUST=1 docker push ${IMAGE_PATH}:signed
+cd -
+```
+
+During the second push, you will be asked to create a root key password and a repository key password (unless you already signed images or pushed signed images to the repository, respectively).
+
```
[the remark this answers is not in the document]
Starkteetje: > If you did not do step 2, you will need to enter your registry's root key from a (potentially) different source.

From step 3, isnt that enough?

Can additionally point out DCT export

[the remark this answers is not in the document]
Starkteetje: Would change it to  `If you did not do step 2, you may need to enter your registry's root key from a different source.` but would still claim that's enough for step 2 still being optional

setup/README.md
```diff
@@ -37,30 +37,22 @@ git clone https://github.com/sse-secure-systems/connaisseur.git
 cd connaisseur
 ```
 
-### 2. (Optional) Prepare test images
+### 2. Setting up Docker Content Trust
```
xopham, line 40:
```suggestion
### 2. Set up Docker Content Trust
```

setup/README.md
```diff
@@ -11,22 +11,233 @@ The guide below offers a simple default configuration for setting up Connaisseur
...
+TEST SUITE: None
+```
+
+Connaisseur was successfully deployed.
+
+### 5. (Optional) Verify Connaisseur is working
+
+If you were just trusting everything someone told you, you wouldn't be here looking for a tool that ensures image integrity, so don't take our word for Connaisseur working. Go ahead and try it:
+
+`kubectl get all -n connaisseur` will show you most* of what you deployed (ConfigMaps, Secrets and MutatingWebhookConfiguration are not shown).
```
xopham, line 122:
```suggestion
`kubectl get all -n connaisseur` will show you most* of what you deployed (ConfigMaps, Secrets and MutatingWebhookConfiguration are not shown).

### 1. Deploy (un-)signed images
```

setup/README.md
```diff
@@ -233,7 +240,7 @@ service/sample created
 
 ### 6. (Optional) Cleaning up
 
-If you want to remove the test deployment, run `kubectl delete all -lapp=sample`.
+If you want to remove the test deployment of step 5, run `kubectl delete all -lapp=sample`.
```
xopham, line 243:
```suggestion
If you want to remove the test deployment of step 1, run `kubectl delete all -lapp=sample`.
```

xopham (approved):
minor structural proposals! Like it :+1:

xopham (approved):
lgtm!

PR #15, opened 2020-10-08
fix: extend validation schemas to allow dates across all timezones and all possible tag values

### LINKED ISSUE #10: 'trust data has invalid format.'
I am using Dockerhub instead of Harbor and getting the following error:
`Error from server: error when creating "deploymentandservice.yaml": admission webhook "connaisseur-svc.connaisseur.svc" denied the request: trust data has invalid format.`
Below is the connaisseur pod log:
`[2020-09-30 00:59:37,182] ERROR: {'message': 'trust data has invalid format.', 'context': {'trust_data_type': 'root', 'user': 'minikube-user', 'operation': 'CREATE', 'kind': 'Deployment', 'namespace': 'default', 'name': 'sample-deployment', 'image': '272965041/signed-image:v1'}, 'detection_mode': False}`

annekebr:
The current schemas for trust data validation neither allow timezones with negative offsets nor tags containing dashes. This fixes both and adds additional tests.

Fix #10

### DIFF
--- a/connaisseur/tests/test_trust_data.py
+++ b/connaisseur/tests/test_trust_data.py
@@ -63,6 +63,63 @@ root_keys = {
         },
     },
 }
+root_keys_sample5 = {
+    "59752a99a56142b0d0af030ed78768f313946fcfe17f153b31f6a4c3e95ba778": {
+        "keytype": "ecdsa",
+        "keyval": {
+            "private": None,
+            "public": (
+                "MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEnvvEn/Y3CG9ZqPxKdnPPmM"
+                "LsLwc4d7+qIU6gYNC0782kyRZYm8rG2NMi45UsZm+fgrRolBezbPGrmTGT"
+                "2j5EbA=="
+            ),
+        },
+    },
+    "654647b9543690aafc97f608ec604f265f01fdf05d2d488d4b3cd332a2db9d43": {
+        "keytype": "ecdsa-x509",
+        "keyval": {
+            "private": None,
+            "public": (
+                "LS0tLS1CRUdJTiBDRVJUSUZJQ0FURS0tLS0tCk1JSUJqRENDQVRHZ0F3SU"
+                "JBZ0lRU3FrTnV5Zk91RGkybzZ6KzJNTVAvekFLQmdncWhrak9QUVFEQWpB"
+                "c01Tb3cKS0FZRFZRUURFeUZ5WldkcGMzUnllUzVsZUdGdGNHeGxMbXh2WT"
+                "JGc09qVXdNREF2YzI1aGEyVXdIaGNOTVRjeApNREF4TWpJd01UUTNXaGNO"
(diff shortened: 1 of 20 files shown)


connaisseur/tests/test_trust_data.py
```diff
@@ -260,7 +337,7 @@ def test_validate_schema_error(td, mock_schema_path):
         (trust_data("tests/data/sample_snapshot.json"), "snapshot"),
         (trust_data("tests/data/sample_timestamp.json"), "timestamp"),
         (trust_data("tests/data/sample_targets.json"), "targets"),
-        (trust_data("tests/data/sample_releases.json"), "targets/releases"),
+        (trust_data("tests/data/sample_releases.json"), "targets/releases")
```
peterthomassen, line 340: No need for this change. Also, you probably didn't run the Black code formatter; I think it puts in the comma?

connaisseur/tests/test_trust_data.py
```diff
@@ -306,12 +383,12 @@ def test_validate_hash_error(td, mock_schema_path, mock_keystore):
         trust_data_._validate_hash(ks)
     assert "failed validating trust data hash." in str(err.value)
 
-
 @pytest.mark.parametrize(
     "data, role",
     [
         (trust_data("tests/data/sample_snapshot.json"), "snapshot"),
         (trust_data("tests/data/sample_timestamp.json"), "timestamp"),
+        (trust_data("tests/data/sample4_targets.json"), "targets")
```
peterthomassen, line 391: comma in the end is better

connaisseur/tests/test_trust_data.py
```diff
@@ -434,3 +521,50 @@ def test_get_digest_error(td, mock_schema_path):
...
+def test_validate(td, mock_schema_path, mock_keystore, data: dict, role: str):
+    ks = KeyStore()
+    _trust_data = td.TrustData(data, role)
+    _trust_data.validate(ks)
+
+@pytest.mark.parametrize(
+    "data, role",
+    [
+        (trust_data("tests/data/sample_timestamp.json"), "timestamp"),
+        (trust_data("tests/data/sample5_targets.json"), "targets")
```
peterthomassen, line 540: comma

peterthomassen (changes requested):
Looks pretty good!

- Let's use the Black code formatter for consistency
- Make sure your commit title is short enough so that GitHub doesn't shorten it (something like 70 characters?)

After that, I'll approve.

In the future, it would be great to keep commits smaller, i.e. have one for the timezone issue, and one for the tag issue. No need to untangle this now.

peterthomassen (dismissed):

xopham (commented):
Really cool! Just one thing I am uncertain about: I think this may not actually check the `format:` specification, as you have to set `format_checker` parameter in `jsonschema.validate`. See: https://python-jsonschema.readthedocs.io/en/stable/validate/

_format_checker – an instance of FormatChecker whose FormatChecker.conforms method will be called to check and see if instances conform to each format property present in the schema. If unprovided, no validation will be done for format. Certain formats require additional packages to be installed (ipv5, uri, color, date-time). The required packages can be found at the bottom of this page._

phbelitz (approved):

PR #16, opened 2020-10-13
docs: add and adjust external and mailto links

xopham:
The reference to the external documentation of DCT has been added, since it is important for the underlying functionality. The mailto links for reaching out to the team were adjusted to use the .dev domain.


phbelitz:
Maybe consider changing the maintainer label in the Dockerfile as well.

https://github.com/sse-secure-systems/connaisseur/blob/befe9920fb7b896c57eadd9ff5fb48b967c085ac/docker/Dockerfile#L1-L12

xopham:
> Maybe consider changing the maintainer label in the Dockerfile as well.
> 
> https://github.com/sse-secure-systems/connaisseur/blob/befe9920fb7b896c57eadd9ff5fb48b967c085ac/docker/Dockerfile#L1-L12

I think we can keep it for now. Does also not directly fit into the PR

phbelitz (approved):

Issue #3, opened 2020-09-11
Connaisseur should not report ready if webhook not deployed during helm installation


Starkteetje:
If something breaks during the helm installation (or otherwise the webhook gets removed), but the connaisseur pods report a ready status, it is not apparent (e.g. when running `kubectl get all -n connaisseur`) that connaisseur is currently inactive and will not be blocking deployments with unsigned images or such with invalid signatures.

Connaisseur pods should report a non-ready status when there's no webhook using them.

PR #20, opened 2020-10-15
#3 helm deployment change

phbelitz:
As stated in #3, the connaisseur pods may show a `ready` state, even though the webhook is not running. This can happen, if during deployment something goes wrong.

The new readiness probe now checks for the notary server health status and the availability of the webhook. This has major implication on the deployment of connaisseur, as the readiness probe was previously used to determine, when to start the webhook. With the readiness probe waiting for the webhook, and the webhook waiting on the readiness probe, a new deployment had to be written, to solve the deadlock.

The new deployment starts up a bootstrap pod during installation and upgrade, which only runs for 30 seconds. During these 30 seconds, the connaisseur pods are automatically `ready`, so the webhook can be deployed. After the 30 seconds the readiness probe behaves as describe above.


README.md
```diff
@@ -114,7 +125,7 @@ The STRIDE threat model has been used as a reference for threat modeling. Each o
...
+| Tampering | An attacker could modify Connaisseur's image policy to bypass signature verification and inject malicious images. Alternatively, the public root key could be replaced, allowing fake trust data to pass as legit. Lastly, the admission controller could be simply deactivated by deleting the webhook. | **RBAC + Monitoring:** An RBAC system can prevent unauthorized changes to both the image policy and public root key. Additionally, the connaisseur readiness probe checks the availability of the webhook and will be set to `Not Ready`, should the webhook not be present. Monitoring should still be used to keep track of the admission controller's webhook availability status, as setting up a fake `connaisseur-bootstrap-sentinel` pod in the `connaisseur` namespace can bypass that readiness probe check. More on that in an upcoming architectural decision record. |
```
Starkteetje, line 128:
```suggestion
| Tampering | An attacker could modify Connaisseur's image policy to bypass signature verification and inject malicious images. Alternatively, the public root key could be replaced, allowing fake trust data to pass as legit. Lastly, the admission controller could be simply deactivated by deleting the webhook. | **RBAC + Monitoring:** An RBAC system can prevent unauthorized changes to both the image policy and public root key. Additionally, the Connaisseur readiness probe checks the availability of the webhook and will be set to `Not Ready` should the webhook not be present. Monitoring should still be used to keep track of the admission controller's webhook availability status, as setting up a fake `connaisseur-bootstrap-sentinel` pod in the `connaisseur` namespace can bypass that readiness probe check. More on that in an upcoming architectural decision record. |
```

connaisseur/flask_server.py
```diff
@@ -59,12 +60,60 @@ def mutate():
...
-    return ("", 200) if health else ("", 500)
+    return ("", 200)
+
+
+# readiness probe
+@APP.route("/ready", methods=["GET", "POST"])
+def readyz():
+    """
+    Handles the '/ready' endpoint and checks the health status of the configured notary
+    server + whether the webhook is installed or not. If the notary is available and the
```
Starkteetje, line 79:
```suggestion
    server and whether the webhook is installed or not. If the notary is available and the
```

connaisseur/flask_server.py
```diff
@@ -59,12 +60,60 @@ def mutate():
...
+    webhook installed it returns a 200 status code. Otherwise should one of them not be
```
Starkteetje, line 80:
```suggestion
    webhook is installed it returns a 200 status code. Otherwise should one of them not be
```

connaisseur/flask_server.py
```diff
@@ -59,12 +60,60 @@ def mutate():
...
+    reachable, 500 is returned. For installation purposes, the readiness probe first
+    checks whether a specific bootstrap pod is running in the namespace. If this pod can
+    be found and it is still running, the readiness probe automatically return 200. This
+    bootstrap pod should only run for the first 30 seconds after installation. After that
```
Starkteetje, line 84:
```suggestion
    bootstrap pod should only run for a short time interval after installation. After that
```

scripts/helm_hook.sh
```diff
@@ -7,21 +7,23 @@ fi
 if [ "$1" == "install" ]
 then
     echo "installing ..."
-    DEPLOYMENT=$(kubectl -n connaisseur get deployments.apps -lapp.kubernetes.io/instance=connaisseur -o=jsonpath='{.items[*].metadata.name}')
-    kubectl wait --for=condition=available --timeout=600s deployments.apps/$DEPLOYMENT -n connaisseur
+    DEPLOYMENT=$(kubectl -n $CONNAISSEUR_NAMESPACE get deployments.apps -lapp.kubernetes.io/instance=connaisseur -o=jsonpath='{.items[*].metadata.name}')
```
Starkteetje, line 10:
```suggestion
    DEPLOYMENT=$(kubectl -n ${CONNAISSEUR_NAMESPACE} get deployments.apps -lapp.kubernetes.io/instance=connaisseur -o=jsonpath='{.items[*].metadata.name}')
```

scripts/helm_hook.sh
```diff
@@ -7,21 +7,23 @@ fi
...
+    kubectl wait --for=condition=available --timeout=600s deployments.apps/$DEPLOYMENT -n $CONNAISSEUR_NAMESPACE
```
Starkteetje, line 11:
```suggestion
    kubectl wait --for=condition=available --timeout=600s deployments.apps/${DEPLOYMENT} -n ${CONNAISSEUR_NAMESPACE}
```

scripts/helm_hook.sh
```diff
@@ -7,21 +7,23 @@ fi
...
     kubectl apply -f /data/webhook.yaml
+    kubectl delete pod -lapp.kubernetes.io/service=bootstrap -n $CONNAISSEUR_NAMESPACE --force=true
```
Starkteetje, line 13:
```suggestion
    kubectl delete pod -lapp.kubernetes.io/service=bootstrap -n ${CONNAISSEUR_NAMESPACE} --force=true
```

scripts/helm_hook.sh
```diff
@@ -7,21 +7,23 @@ fi
...
     echo "done."
 elif [ "$1" == "upgrade" ]
 then
     echo "upgrading ..."
-    DEPLOYMENT=$(kubectl -n connaisseur get deployments.apps -lapp.kubernetes.io/instance=connaisseur -o=jsonpath='{.items[*].metadata.name}')
-    kubectl rollout status $DEPLOYMENT
+    DEPLOYMENT=$(kubectl -n $CONNAISSEUR_NAMESPACE get deployments.apps -lapp.kubernetes.io/instance=connaisseur -o=jsonpath='{.items[*].metadata.name}')
```
Starkteetje, line 18:
```suggestion
    DEPLOYMENT=$(kubectl -n ${CONNAISSEUR_NAMESPACE} get deployments.apps -lapp.kubernetes.io/instance=connaisseur -o=jsonpath='{.items[*].metadata.name}')
```

scripts/helm_hook.sh
```diff
@@ -7,21 +7,23 @@ fi
...
+    kubectl rollout status $DEPLOYMENT -n $CONNAISSEUR_NAMESPACE
```
Starkteetje, line 19:
```suggestion
    kubectl rollout status ${DEPLOYMENT} -n ${CONNAISSEUR_NAMESPACE}
```

scripts/helm_hook.sh
```diff
@@ -7,21 +7,23 @@ fi
...
     kubectl apply -f /data/webhook.yaml
+    kubectl delete pod -lapp.kubernetes.io/service=bootstrap -n $CONNAISSEUR_NAMESPACE --force=true
```
Starkteetje, line 21:
```suggestion
    kubectl delete pod -lapp.kubernetes.io/service=bootstrap -n ${CONNAISSEUR_NAMESPACE} --force=true
```

scripts/helm_hook.sh
```diff
@@ -7,21 +7,23 @@ fi
...
     echo "done."
 elif [ "$1" == "delete" ]
 then 
     echo "deleting ..."
-    WEBHOOK=$(kubectl -n connaisseur get mutatingwebhookconfigurations.admissionregistration.k8s.io -lapp.kubernetes.io/instance=connaisseur -o=jsonpath='{.items[*].metadata.name}')
+    WEBHOOK=$(kubectl -n $CONNAISSEUR_NAMESPACE get mutatingwebhookconfigurations.admissionregistration.k8s.io -lapp.kubernetes.io/instance=connaisseur -o=jsonpath='{.items[*].metadata.name}')
```
Starkteetje, line 26:
```suggestion
    WEBHOOK=$(kubectl -n ${CONNAISSEUR_NAMESPACE} get mutatingwebhookconfigurations.admissionregistration.k8s.io -lapp.kubernetes.io/instance=connaisseur -o=jsonpath='{.items[*].metadata.name}')
```

scripts/helm_hook.sh
```diff
@@ -7,21 +7,23 @@ fi
...
     kubectl delete mutatingwebhookconfigurations.admissionregistration.k8s.io $WEBHOOK
```
Starkteetje, line 27:
```suggestion
    kubectl delete mutatingwebhookconfigurations.admissionregistration.k8s.io ${WEBHOOK}
```

Starkteetje (commented):
Also: Did we agree on whether to make the 30s configurable? Would you wanna do that in a different PR?
Also: And shouldnt footer be Fix #3 ?

helm/templates/bootstrap-sentinel.yaml
```diff
@@ -0,0 +1,17 @@
+apiVersion: v1
+kind: Pod
```
Starkteetje, line 2: This should be a Job following last week's discussion, right?

phbelitz: If we change it to a Job, then the job won't disappear and we have the same problem, just with another resource. Additionally, the pod now gets deleted anyways by the webhook-creation-job, so no need to change it anymore.

phbelitz:
i haven't yet added the version upgrade! wil do after review.

connaisseur/flask_server.py
```diff
@@ -59,12 +60,61 @@ def mutate():
...
+    sentinel = os.environ.get("CONNAISSEUR_SENTINEL")
+    sentinel_ns = os.environ.get("CONNAISSEUR_NAMESPACE")
+    sentinel_path = "api/v1/namespaces/{ns}/pods/{name}".format(
+        ns=sentinel_ns, name=sentinel
+    )
+
+    sentinel_response = {}
+    try:
+        sentinel_response = api.request_kube_api(sentinel_path)
+    except Exception:  # nosec
```
peterthomassen, line 97: Is this supposed to catch the exception that's raised when the sentinel is not running? If so, we should be explicit about it. If not, what other exceptions do you have in mind?

connaisseur/flask_server.py
```diff
@@ -59,12 +60,61 @@ def mutate():
...
+
+# readiness probe
+@APP.route("/ready", methods=["GET", "POST"])
+def readyz():
+    """
+    Handles the '/ready' endpoint and checks the health status of the configured notary
+    server and whether the webhook is installed or not. If the notary is available and the
+    webhook is installed it returns a 200 status code. Otherwise should one of them not be
+    reachable, 500 is returned. For installation purposes, the readiness probe first
+    checks whether a specific bootstrap pod is running in the namespace. If this pod can
```
peterthomassen, line 82: Clarify that the "specific bootstrap pod" is what you're calling "sentinel" below

connaisseur/flask_server.py
```diff
@@ -59,12 +60,61 @@ def mutate():
...
+    webhook = os.environ.get("CONNAISSEUR_WEBHOOK")
+    webhook_path = (
+        "apis/admissionregistration.k8s.io/v1beta1/mutatingwebhookconfigurations/{name}"
+    ).format(name=webhook)
+
+    webhook_response = None
+    try:
+        webhook_response = api.request_kube_api(webhook_path)
+    except Exception:  # nosec
+        pass
```
peterthomassen, line 112: You can move `webhook_response = None` here.

Also, what are the expected exception cases? Can we be explicit about them?

peterthomassen (changes requested):

peterthomassen (approved):

peterthomassen (approved):
lgtm!

PR #21, opened 2020-10-20
docs/feat: Allow and document integration with Azure Container Registry

Starkteetje:
Since the Azure Container Registry has a slightly different API than other notaries, some Connaisseur internals had to be adapted. In addition, there's a guide to integrate the ACR starting differently than the general Kubernetes guide.

### DIFF
--- a/connaisseur/notary_api.py
+++ b/connaisseur/notary_api.py
@@ -23,6 +23,12 @@ def health_check(host: str):
     if not host:
         return False
 
+    # The ACR does not expose the notary health endpoint: https://github.com/Azure/acr/issues/475
+    # As a consequence, Connaisseur Pods will not appear as unhealthy if they cannot connect to the ACR
+    # This behavior differs from behavior for all other notaries
+    if is_acr():
+        return True
+
     url = f"https://{host}/_notary_server/health"
 
     request_kwargs = {"url": url}
@@ -31,6 +37,14 @@ def health_check(host: str):
     response = requests.get(**request_kwargs)
 
     return response.status_code == 200
+
+
+def is_acr():
+    """
+    Checks whether the notary server should be considered to be the
+    Azure Container Registry (ACR).
+    """
+    return os.environ.get("IS_ACR", "0") == "1"
 
 
 def is_notary_selfsigned():
@@ -295,7 +309,10 @@ def get_auth_token(url: str):
     response.raise_for_status()
 
     try:
-        token = response.json()["token"]
+        if is_acr():
+            token = response.json()["access_token"]
+        else:
+            token = response.json()["token"]
     except KeyError:
         raise NotFoundException(
             "no token in authentication server response.", {"auth_url": url}
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -1,7 +1,7 @@
 # configure connaisser deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v1.1.5
+  image: securesystemsengineering/connaisseur:v1.2.0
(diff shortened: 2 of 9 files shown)


helm/values.yaml
```diff
@@ -45,6 +45,9 @@ notary:
   rootPubKey: |
     -----BEGIN PUBLIC KEY-----
     -----END PUBLIC KEY-----
+  # whether the notary server will be the ACR
```
xopham, line 48: This seems insufficiently explicit. I would recommend to write it out:
`# if you use Azure Container Registry (ACR) for your notary changes ... . For compatibility with ACR set isAcr to `true`

setup/acr/README.md
```diff
@@ -0,0 +1,66 @@
...
+# Create a service principal with the Reader role on your registry
+SP_CREDENTIALS=$(az ad sp create-for-rbac --name "${CONNAISSEUR_SP_NAME}" --role Reader --scopes ${REGISTRY_ID})
+
+NOTARY_USER=$(echo ${SP_CREDENTIALS} | jq -r '.appId')
+NOTARY_PASSWORD=$(echo ${SP_CREDENTIALS} | jq -r '.password')
+```
+
+> If something goes wrong with the creation of the Service Principal, check that you actually have sufficient privileges to create a Service Principal.
+
+At this point, the adaptations specific to AKS and ACR are complete and you can continue with the [second step of the general Kubernetes guide](../README.md#2-set-up-docker-content-trust).
```
xopham, line 64: Could add a note that Azure also gives some guidance how to implement image signing in Azure pipelines which might be of interest to review: https://docs.microsoft.com/en-us/azure/devops/pipelines/ecosystems/containers/content-trust?view=azure-devops
like further reading for a user...

connaisseur/notary_api.py
```diff
@@ -23,6 +23,10 @@ def health_check(host: str):
     if not host:
         return False
 
+    # The ACR does not expose the notary health endpoint: https://github.com/Azure/acr/issues/475
+    if is_acr():
```
xopham, line 27: what is the caveat of this like ACR notary is actually not healthy....?

xopham (commented):
lgtm! Very nice work. Good find.

I made some minor comments.

connaisseur/notary_api.py
Starkteetje: This basically translates into us not checking whether ACR is healthy, so Conny would show `Ready` even if the ACR were down, whereas for all other notaries Conny pods would become unhealthy once the connection to notary doesn't work anymore.
Do you think, I should make this more explicit here? Kinda thought it would be obvious from the code

xopham: might make a comment in code or in the docs I guess...

connaisseur/notary_api.py
```diff
@@ -23,6 +23,12 @@ def health_check(host: str):
     if not host:
         return False
 
+    # The ACR does not expose the notary health endpoint: https://github.com/Azure/acr/issues/475
```
xopham, line 26: would really write out ACR as Azure Container Registry here and start with that:
`Set to true if you are using Azure Container Registry (ACR). (...)`

Otherwise, people may be confused by this option.

Starkteetje: You probably mean this: https://github.com/sse-secure-systems/connaisseur/blob/db06c5e834bd75a78e7567c5b76b8a0e7fcf728b/helm/values.yaml#L48
Whoever looks into the actual implementation instead of the config file can know what ACR is at latest from following the link

phbelitz (approved):

PR #23, opened 2020-10-21
fix: Default to Docker Hub notary

### LINKED ISSUE #13: Wrong documentation regarding null value in notary URL
https://github.com/sse-secure-systems/connaisseur/blob/master/helm/values.yaml#L24 state that the `domain to the notary server can be null to use the public docker hub notary server`. However, when set to `null` validation fails with

```bash
bash helm/certs/gen_certs.sh
Generating RSA private key, 4096 bit long modulus (2 primes)
.............................................................................................................................................................................++++
..............................++++
e is 65537 (0x010001)
Signature ok
subject=CN = connaisseur-svc.connaisseur.svc
Getting Private key
kubectl create ns connaisseur || true
namespace/connaisseur created
kubectl config set-context --current --namespace connaisseur
Context "**********" modified.
helm install connaisseur helm --wait
Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: unknown object type "nil" in ConfigMap.data.NOTARY_SERVER
make: *** [Makefile:20: install] Error 1
```

We should either correct the documentation or make the value actually default to `notary.docker.io`

Starkteetje:
Previously we incorrectly stated that a notary host value of null would default to the Docker Hub notary. The actual behavior now matches the stated one.

Fix #13

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
 NAMESPACE = connaisseur
 IMAGE = $(IMAGE_NAME):$(TAG)
 IMAGE_NAME = securesystemsengineering/connaisseur
-TAG = v1.1.3
+TAG = v1.1.4
 POD = not-smooth-app
 
 .PHONY: all docker certs install unistall upgrade annihilate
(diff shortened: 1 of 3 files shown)


xopham (approved):
lgtm

Issue #12, opened 2020-10-02
container hardening


phbelitz:
We should investigate, whether the security of the connaisseur container can be futher improved, in case of a container compromise.

Examples:
1. change the user inside the container from root, to a less privileged one
2. set the filesystem to be read-only

PR #24, opened 2020-10-22
feat: harden image and deploy with kubernetes security context

Starkteetje:
The Connaisseur image is based on an alpine image, which has now been slightly hardened. Furthermore the deployment to Kubernetes is now being controlled by a SecurityContext that is a lot stricter than the default configuration. Also Pods now use the runtime's default seccomp profile instead of none.

Fix: #12

### DIFF
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -2,10 +2,16 @@ FROM python:alpine
 
 WORKDIR /app
 
+# Harden image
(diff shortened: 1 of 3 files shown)


xopham (commented):
lgtm! Was it tested that we are not breaking any critical capabilities?

phbelitz (approved):

PR #30, opened 2020-10-23
docs: Architecture Decision Record regarding Sentinel

Starkteetje:
Adds folder for Architecture Decision Records with an explaination of why we use them. Adds first ADR regarding the use of the connaisseur-bootstrap-sentinel Pod.

### DIFF
--- a/adr/ADR-1_connaisseur-bootstrap-sentinel.md
+++ b/adr/ADR-1_connaisseur-bootstrap-sentinel.md
@@ -1,0 +1,37 @@
+# ADR 1: Connaisseur Bootstrap Sentinel
+
+## Status
+
+Accepted
+
+## Context
+
+Connaisseur's main components are a MutatingWebhookConfiguration and the Connaisseur Pods. The MutatingWebhookConfiguration intercepts requests to create or update Kubernetes resources and forwards them to the Connaisseur Pods tasked, on a high level, with verifying trust data. The order of deploying both components matters, since a blocking MutatingWebhookConfiguration without the Connaisseur Pods to answer its requests would also block the deployment of said Pods.
+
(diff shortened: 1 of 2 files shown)


xopham (commented):
generally lgtm!
- Was not involved in the ADR 1, so won't comment on that.
- Would recommend adding a minimal template for ADRs (Title, Status, Context, Considered Options, Decision)

adr/ADR 1: connaisseur-bootstrap-sentinel.md
```diff
@@ -0,0 +1,37 @@
...
+
+### Negative Consequences
+
+On the other hand, if an adversary can deploy a Pod named `connaisseur-bootstrap-sentinel` to Connaisseur's Namespace, the Connaisseur Pods will always show `Ready` regardless of the MutatingWebhookConfiguration. However, if an adversary can deploy to Connaisseur's Namespace, chances are Connaisseur can be compromised anyways. More importantly, if not a single Connaisseur Pod is successfully deployed or if the notary healthcheck fails during the initial period of 30 seconds, then the deployment will fail regardless of possible recovery at a later time. Another issue would be the `connaisseur-bootstrap-sentinel` Pod being left behind, however since it has a very limited use-case we can also clean it up during the deployment, so apart from the minimal additional complexity of the deployment this is a non-issue.
```
phbelitz, line 37: `More importantly, if not a single Connaisseur Pod is successfully deployed or if the notary healthcheck fails during the initial period of 30 seconds, then the deployment will fail regardless of possible recovery at a later time.`

What are you trying to say here? That if the deployment fails, an adversery can't recover it? Is this supposed to be a positive or negative consequece?

phbelitz (changes requested):
left some comments

adr/ADR 1: connaisseur-bootstrap-sentinel.md
Starkteetje: No, what I'm saying is that if notary is down for the first 30s of the deployment, the deployment will definetly fail, even if notary recovers afterwards

phbelitz: ah ok now i get it

phbelitz (dismissed):

phbelitz (approved):

PR #31, opened 2020-10-28
fix/docs: improve compatibility of sed expression to set root pub key

xopham:
The sed expression was using '/' for replacement. That is however part of the base64 character set and is therefore replaced with '?'


Starkteetje (approved):
Good find :+1:

PR #34, opened 2020-10-28
docs: use small caps for repo and image names

xopham:
Docker does not allow usage of capital letters in image names and repos. This seems to also cause problems with DCT.

### DIFF
--- a/setup/README.md
+++ b/setup/README.md
@@ -27,7 +27,7 @@ We assume you have a running Kubernetes cluster already. Since this tutorial wor
 
 ### 1. Set up environment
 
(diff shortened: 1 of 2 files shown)


setup/README.md
```diff
@@ -27,7 +27,7 @@ We assume you have a running Kubernetes cluster already. Since this tutorial wor
 
 ### 1. Set up environment
 
-First off, we'll export the variables used in this tutorial to allow you to immediately use your own names/URLs instead of relying on our default names. Below, substitute your own values as appropriate:
+First off, we'll export the variables used in this tutorial to allow you to immediately use your own names/URLs instead of relying on our default names. Below, substitute your own values as appropriate (use small caps for compatibility with Docker):
```
Starkteetje, line 30: should that be small letters instead of small caps?

Starkteetje (approved):

PR #35, opened 2020-10-29
docs: Move IMAGE_PATH variable to test section

Starkteetje:
Previously the IMAGE_PATH variable was defined inside the Connaisseur setup section although it was no longer used there. Its definition is now moved to the section dealing with testing Connaisseur.

### DIFF
--- a/setup/README.md
+++ b/setup/README.md
@@ -27,11 +27,9 @@ We assume you have a running Kubernetes cluster already. Since this tutorial wor
 
 ### 1. Set up environment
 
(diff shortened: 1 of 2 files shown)


phbelitz (approved):

xopham (changes requested):
Would it be ok if we moved this into #40 and drop this PR as we will otherwise have rebase nightmare?

Starkteetje:
Obsolute due to #40

PR #40, opened 2020-11-01
Simplify setup guide

xopham:


helm/values.yaml
```diff
@@ -33,7 +33,7 @@ notary:
   # if notary uses an authentication server, give a user and password
   # with pull rights
   auth:
-    enabled: true
+    enabled: false
```
annekebr, line 36: Why do we want to change the default here?

setup/Dockerfile
```diff
@@ -0,0 +1,3 @@
+FROM python:3.7-alpine
```
annekebr, line 1: We could keep the visual distinction between the signed and the unsigned image if we want to by using a build-argument e.g. 
``` 
FROM python:3.7-alpine
ARG MESSAGE="Hello World of signing! I am a simple image sleeping forever!"
RUN echo "echo ${MESSAGE}; sleep infinity;" >> hello.sh
CMD ["sh", "hello.sh"]
``` 
and then change the message during build by passing it as build-arg somehow like
`DOCKER_CONTENT_TRUST=0 docker build -f Dockerfile -t ${IMAGE_PATH}:signed  --build-arg MESSAGE="I am an unsigned image and won't be allowed to the party :-(" .

README.md
```diff
@@ -35,6 +37,35 @@ Enter [*Docker Content Trust (DCT)*](https://docs.docker.com/engine/security/tru
 
 **Connaisseur** solves this problem by bringing image signature verification into Kubernetes. It uses the already existing Docker Content Trust solution and adds it to the cluster as a mutating admission controller. It pulls trust data from *Notary* and verifies it before any commitment to Kubernetes happens.
 
+## Demo
+
+We provide two public sample images for testing. Setting the trust anchor of Connaisseur to our public key allows you to have a working demo in just a few minutes. Below you see a full setup starting [MicroK8s](https://microk8s.io/) as a test cluster, installation of Connaisseur and testing with a signed and unsigned image.
```
Starkteetje, line 42: Would put a comma after below. Afaik its necessary if there could be confusion, and it could read "below, you see" or "below you, see"

Starkteetje: "MicroK8s as a test cluster" sounds weird, why not "a Micro test cluster?

Starkteetje, line 42:
```suggestion
We provide two public sample images for testing. Setting the trust anchor of Connaisseur to our public key allows you to have a working demo in just a few minutes. Below, you see a full setup starting a [MicroK8s](https://microk8s.io/) test cluster, installing of Connaisseur and testing with a signed and unsigned image.
```

setup/README.md
```diff
@@ -124,134 +172,83 @@ If you were just trusting everything someone told you, you wouldn't be here look
...
 `kubectl get all -n connaisseur` will show you most* of what you deployed (ConfigMaps, Secrets and MutatingWebhookConfiguration are not shown).
 
-### 1. Deploy (un)signed images
+### Deploy (un)signed images
+
+Let's test if our configuration works. We need to prepare a signed and an unsigned image and push these to a remote repository. For simplicity, we define an image path variable (host, repository and image). Below, substitute `IMAGE_PATH` with your own the values as appropriate:
+
+```bash
+# Typically, IMAGE_PATH=<YOUR-REGISTRY>/<YOUR-REPOSITORY-NAME-/-DOCKER-HUB-ID>/<IMAGE-NAME>
+IMAGE_PATH=docker.io/connytest/testimage
```
annekebr, line 181: Who is the owner of this repo?

README.md
```diff
@@ -35,6 +37,35 @@ Enter [*Docker Content Trust (DCT)*](https://docs.docker.com/engine/security/tru
...
+
+![](img/connaisseur_demo.gif)
+
+Make sure `kubectl`, `helm`, `docker`, `git`, `yq` and `make` are installed. Your `kubectl` context should be set to your Kubernetes test cluster.
```
Starkteetje, line 46:
```suggestion
Make sure `docker`, `git`, `helm`, `kubectl`, `make` and `yq` are installed. Your `kubectl` context should be set to your Kubernetes test cluster.
```

setup/README.md
```diff
@@ -11,13 +11,15 @@ The guide below offers a simple default configuration for setting up Connaisseur
...
+
 2. **[Azure Kubernetes Services](https://docs.microsoft.com/en-us/azure/aks/) with [Azure Container Registry (ACR)](https://docs.microsoft.com/en-us/azure/container-registry/)**: Microsoft Azure is currently the only one of the big cloud providers with a managed Kubernetes and container registry that supports DCT. If you want to set up Connaisseur with ACR, head over to [the specialized setup guide](acr/README.md).
 
 
 ## Requirements
 
-We assume you have a running Kubernetes cluster already. Since this tutorial works with the CLI, we furthermore assume you have the `docker`, `git`, `helm`, `kubectl`, `make` binaries installed and usable, i.e. having run `docker login` and having switched to the appropriate `kubectl` context.
+We assume you have a running Kubernetes cluster already. Since this tutorial works with the CLI, we furthermore assume you have the `docker`, `git`, `helm`, `kubectl`, `make`, `openssl` and `yq` binaries installed and usable, i.e. having run `docker login` and having switched to the appropriate `kubectl` context.
```
Starkteetje, line 22:
```suggestion
We assume you have a running Kubernetes cluster already. Since this tutorial works with the CLI, we furthermore assume you have the `docker`, `git`, `helm`, `kubectl`, `make`, `openssl` and `yq` binaries installed and usable, i.e. having run `docker login` and having switched to the appropriate `kubectl` context.
```

helm/values.yaml
xopham: Mainly for the reason that the default creds do not work, so in case you need to authorize you have to touch this part anyways. In case, you use a publicly accessible repo, you do not even have to touch on this :shrug: 
Also, it makes the demo shorter ;-)

setup/Dockerfile
xopham: Good point. I, however, do not think that changing what is in the image is actually critical for testing it after initial setup and would for that reason opt for keeping things as simple as possible, as it is already quite complicated.
What do you think?

setup/README.md
```diff
@@ -44,45 +40,97 @@ cd connaisseur
...
+rm root-priv.key root.pub
 cd -
 ```
 
-> The `values.yaml` file in the `helm` folder is _the_ config file for Connaisseur. If you have some time to spare, have a look :)
->
-> For example, you can turn on detection mode by running `sed -i "s/detection_mode: false/detection_mode: true/" helm/values.yaml`, which will allow your deployments to pass, but will log the violation. This might be useful when rolling out Connaisseur to an existing Kubernetes installation in order to first get a grasp on remaining obstacles before turning on (the default) blocking mode.
-
-> What is being done at this point is to set your personal root key as Connaisseur's trust anchor. If used in production, you may wish to go for [setting up delegation keys](https://docs-stage.docker.com/engine/security/trust/trust_delegation/#creating-delegation-keys) and keep the root key away from everyday-signing.
+- The resulting `/helm/values.yaml` should look similar as for the first case above.
```
Starkteetje, line 133:
```suggestion
- The resulting `/helm/values.yaml` should have an excerpt looking similar to what is shown in the first case above.
```

setup/README.md
xopham: Me. It is also the one I use in the demo. It all ties together and you are safe if you use that value and the sample key below ;-)
Also created #41 about this. What is bad if we move it into SSE, is that it makes every command like this much longer...just from a UX perspective.
What do you think?

setup/README.md
```diff
@@ -124,134 +172,83 @@ If you were just trusting everything someone told you, you wouldn't be here look
...
 
-Your signed images were allowed through, in contrast to those unsigned ones with unknown content. DCT, not in a nutshell, but in Kubernetes!
+or if you pushed the unsigned image to another registry altogether:
+
+```bash
+Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: no trust data for image "docker.io/connytest/testimage:unsigned".
+```
 
-### 2. Cleanup
+> Note that while deployment of containers is blocked Kubernetes will still create services or other resources you might specify in a deployment.yaml, because those does not reference an image and are thus not denied by Connaisseur. You might have to clean up denied deployments.
```
Starkteetje, line 245:
```suggestion
> Note that while deployment of containers is blocked Kubernetes will still create services or other resources you might specify in a deployment.yaml, because those do not reference an image and are thus not denied by Connaisseur. You might have to clean up denied deployments.
```

helm/values.yaml
annekebr: I think I would add a comment above explaining that :) Just to point out that this is not an insecure default

setup/README.md
```diff
@@ -124,134 +172,83 @@ If you were just trusting everything someone told you, you wouldn't be here look
...
 
-If you want to remove the test deployment of step 1, run `kubectl delete all -lapp=sample`.
```
Starkteetje, line 250: Would still keep this deletion of test resources, maybe in the cleanup section

setup/README.md
annekebr: Oh, didn't see that :+1: I think it would be cleaner if it was moved into SSE. Also, I don't think the long path matters much. However, if we decide to move it we can do it independently of this PR I guess :)

setup/acr/README.md
```diff
@@ -19,48 +19,35 @@ For the most part, this tutorial will be the same as [for a regular Kubernetes c
...
-SP_CREDENTIALS=$(az ad sp create-for-rbac --name "${CONNAISSEUR_SP_NAME}" --role Reader --scopes ${REGISTRY_ID})
-
-NOTARY_USER=$(echo ${SP_CREDENTIALS} | jq -r '.appId')
-NOTARY_PASSWORD=$(echo ${SP_CREDENTIALS} | jq -r '.password')
+az ad sp create-for-rbac --name "<SERVICE-PRINCIPLE-NAME>" --role Reader --scopes ${REGISTRY_ID}
 ```
 
 > If something goes wrong with the creation of the Service Principal, check that you actually have sufficient privileges to create a Service Principal.
 
+Note down the `.appId` and `.password` values. These are your notary username and password, respectively. You will need them later in the setup when [configuring Notary](../README.md#configure-notary).
```
Starkteetje, line 49:
```suggestion
Note down the `.appId` and `.password` values. These are your notary username and password, respectively. You will need them later in the setup when [configuring Notary](../README.md#configure-notary). Use the `.appId` value as your Notary username and `.password` as password.
```

setup/Dockerfile
annekebr: Fair! Just wanted to propose how we could keep the visibility if we thought it was useful ;-)

helm/values.yaml
```diff
@@ -33,7 +33,7 @@ notary:
   # if notary uses an authentication server, give a user and password
   # with pull rights
```
annekebr, line 34:
```suggestion
  # with pull rights
  # The default we use is the notary instance from docker
  # which is publicly accessible and does not need authentication
```

helm/values.yaml
xopham: great idea!

setup/README.md
xopham: yes, that was my rationale as well...

setup/README.md
xopham: but there is no real resources anymore, just two pods. Will add that below though...

xopham: done!

Starkteetje (approved):

PR #42, opened 2020-11-02
ci/test: Add integration test

### LINKED ISSUE #25: Integration test
Would be nice if we had an integration test of at least roughly the setup guide. If it ran inside the pipeline that would be best, however as a start, local test script may be sufficient

Starkteetje:
Add integration test with KinD cluster in Github Actions. Tests building of docker images, installing Connaisseur, rejecting unsigned image, rejecting image signed under different key, accepting signed image and uninstallation.

Fix #25

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -132,4 +132,39 @@ jobs:
         with:
            name: trivy-report
            path: trivy-report.json
-  
+
+  integration-test:
+    runs-on: ubuntu-latest
+    needs: [bandit, safety, trivy]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install yq and bash
+        run: |
+          sudo snap install yq
+          sudo apt update
+          sudo apt install bash -y
+
+      - name: Build Connaisseur image
+        run: |
+          sed -i "s+TAG =.*+TAG = ${RELEASE_VERSION}+" Makefile
+          sed -i "s+IMAGE_NAME =.*+IMAGE_NAME = ${RELEASE_IMAGE}+" Makefile
+          make docker
+
+      - name: Create KinD cluster
+        run: |
+          GO111MODULE="on" go get sigs.k8s.io/kind
+          kind create cluster --wait 120s
+
+      - name: Check KinD cluster
+        run: |
+          kubectl get nodes
+
+      - name: Add images to KinD
+        run: |
+          kind load docker-image ${RELEASE_IMAGE}:${RELEASE_VERSION}
+          kind load docker-image ${RELEASE_IMAGE}:helm-hook
+
+      - name: Run actual integration test
+        run: |
+          bash connaisseur/tests/integration/integration-test.sh
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -68,13 +68,14 @@ This helps identify bugs in changes before pushing.
 
 > :information_source: **INFO** We believe that testing should not only ensure functionality, but also aim to test for expected security issues like injections and appreciate if security tests are added with new functionalities.
 
-Besides the unit testing and before any PR can be merged, an integration test should be carried out whereby:
+Besides the unit testing and before any PR can be merged, an integration test is carried out whereby:
 - Connaisseur is successfully installed in a test cluster
 - a non-signed image is deployed to the cluster and denied
+- an image signed with an unrelated key is denied
 - a signed image is deployed to the cluster and passed
 - Connaisseur is successfully uninstalled
 
-We hope to automate these tests in the future and also provide a test script for local testing with minikube.
+You can also run this integration test on a local cluster after setting the necessary environment variables.
 
 ### Signed Commits and Pull Requests
 All changes to the `master` branch must be signed which is enforced via [branch protection](https://docs.github.com/en/free-pro-team@latest/github/administering-a-repository/about-required-commit-signing). This can be achieved by only fast-forwarding signed commits or signing of merge commits by a contributor. Consequently, while we appreciate signed commits in PRs, we do not require it.
--- a/connaisseur/tests/integration/integration-test.sh
+++ b/connaisseur/tests/integration/integration-test.sh
@@ -1,0 +1,59 @@
+#! /bin/bash
+set -euo pipefail
+
+# This script is expected to be called from the root folder of Connaisseur
+
+if [[ -z "${RELEASE_IMAGE-}" ]] || [[ -z "${RELEASE_VERSION-}" ]]; then
+  echo "Missing environment variables."
+  exit 1
+fi
+
+echo 'Preparing Connaisseur config...'
+envsubst < connaisseur/tests/integration/update-config.yaml > update
+yq write --inplace --script update helm/values.yaml # Coincidentally, this also ensure all used env variables are set
+rm update
+echo 'Config set'
+
+echo 'Installing Connaisseur...'
+make install || { echo 'Failed to install Connaisseur'; exit 1; }
+echo 'Successfully installed Connaisseur'
+
+echo 'Testing unsigned image...'
+kubectl run pod --image=connytest/testimage:unsigned >output.log 2>&1 || true
+
+if [[ "$(cat output.log)" != 'Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: could not find signed digest for image "docker.io/connytest/testimage:unsigned" in trust data.' ]]; then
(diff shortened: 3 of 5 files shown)


.github/workflows/cicd.yaml
```diff
@@ -120,4 +120,42 @@ jobs:
...
+
+      - name: Add images to KinD
+        run: |
+          kind load docker-image ${RELEASE_IMAGE}:${RELEASE_VERSION}
+          kind load docker-image ${RELEASE_IMAGE}:helm-hook
+
+      - name: Run actual integration test
+        env:
+          NOTARY_USER: ${{ secrets.DOCKER_USERNAME }}
+          NOTARY_PASSWORD: ${{ secrets.DOCKER_PASSWORD }}
```
xopham, line 159: these are not used/needed? Probably, there due to the originally required auth to notary...

xopham (commented):
Really neat work! Have not reviewed it all yet, but one question: have you tested the failing of all these?
- build of conny
- unsigned image not found ...
- signed image not signed or not found
And did you also test for the failing of unexpected error messages?
Just asking this to be sure because I see that the integration test will lead us to auto-building and pushing conny and in that case any breaking would be bad if handled falsely...

.github/workflows/cicd.yaml
Starkteetje: Correct, will remove

Starkteetje:
> Really neat work! Have not reviewed it all yet, but one question: have you tested the failing of all these? [...]

 Quoting from the description:
> Tests building of docker images, installing Connaisseur, rejecting unsigned image, accepting signed image and uninstallation.

> And did you also test for the failing of unexpected error messages?

Test script has -e flag set, same for the general Github Actions Pipeline

.github/workflows/cicd.yaml
xopham: probably should also delete the creds from github...for now...

xopham: or keep if you assume the next step is up soon...

Starkteetje: Would keep since I guess I'm doing release soonish

connaisseur/tests/integration/integration-test.sh
```diff
@@ -0,0 +1,75 @@
+#! /bin/bash
+set -euo pipefail
+
+envsubst <connaisseur/tests/integration/values.yaml >helm/values.yaml
+
+make install || { echo "Failed to install Connaisseur"; exit 1; }
```
xopham, line 6: I noticed that on my machine `make install` can get stuck and I am not sure whether it times out. That would potentially make the ci freeze here, right?

xopham (commented):
lgtm!

connaisseur/tests/integration/integration-test.sh
Starkteetje: If it times out, it would "freeze" for the 10min timeout then die
Not sure what it would be if not a timeout. Why are you not sure whether it's a timeout on your end? But yes, if `make install` produced an endless loop, then this would be an endless loop and "freeze" CI

xopham (dismissed):
lgtm

xopham (approved):

phbelitz (commented):
https://github.com/sse-secure-systems/connaisseur/blob/0f016d2bf3e1387eb094681aedbd5198201beb8b/.github/workflows/cicd.yaml#L11-L13
This would be the third occurrence, where we'd have to change the image version manually. Maybe let's agree on a single point and all other parse the version from there. I would say the `values.yaml` is the origin.

Starkteetje:
> https://github.com/sse-secure-systems/connaisseur/blob/0f016d2bf3e1387eb094681aedbd5198201beb8b/.github/workflows/cicd.yaml#L11-L13
> 
> This would be the third occurrence, where we'd have to change the image version manually. Maybe let's agree on a single point and all other parse the version from there. I would say the `values.yaml` is the origin.

Agreed. However, the "version" here is completely irrelevant, since it is never pushed but only used locally. Wanted to change the naming only in the release PR, since it is orthogonal to the integration test

connaisseur/tests/integration/integration-test.sh
```diff
@@ -0,0 +1,75 @@
...
+echo "Successfully installed Connaisseur"
+
+kubectl create -f - << EOF >output.log 2>&1 || true
```
phbelitz, line 9: maybe use `kubectl run pod --image=connytest/testimage:unsigned`. it's a lot smaller and should suffice for the test. same for the signed one. note that the output messages will change.

connaisseur/tests/integration/values.yaml
```diff
@@ -0,0 +1,72 @@
+# configure connaisser deployment
```
annekebr, line 1: Do we need the duplicate here or could we alter the "real" `values.yaml` file directly to prevent that changes to the `values.yaml` are forgotten to be updated here?

connaisseur/tests/integration/integration-test.sh
```diff
@@ -0,0 +1,48 @@
...
+
+echo "Installing Connaisseur..."
+make install || { echo "Failed to install Connaisseur"; exit 1; }
+echo "Successfully installed Connaisseur"
+
+echo "Testing unsigned image..."
+kubectl run pod --image=connytest/testimage:unsigned >output.log 2>&1 || true
+
+if [[ "$(cat output.log)" != 'Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: could not find signed digest for image "docker.io/connytest/testimage:unsigned" in trust data.' ]]; then
+  echo "Failed to deny unsigned deployment. Output of deployment:"
```
annekebr, line 14: Nitpicky, because it is contained in the output, but I think that the echoed message is not necessarily true - if I am not mistaken the image could also be denied with an unknown error, in which case we'd also want to exit with 1, but we could adapt the echo message to make it clear and prevent contradiction of output and echoed message. If my assumption is correct, I would suggest
```suggestion
  echo "Failed to deny unsigned deployment or denied with an unkown error. Output of deployment:"
```

annekebr: Might also refer to @xopham s comment here https://github.com/sse-secure-systems/connaisseur/pull/42#pullrequestreview-521579294

connaisseur/tests/integration/values.yaml
Starkteetje: Thought it would be easier this way in the sense that moving this prepared values is a single `envsubst` step, while altering the `values.yaml` will be multiple steps (which may become more over time) that could fail but have nothing to do with the deployment.
In both cases though, one would immediately get the feedback on the PR's first pipefile, so neither should be a problem.

So choice was deliberate, but I'm not strongly opinionated. If, given the above consideration,you still feel altering > duplication, I'll change it

annekebr: Hmm, you're right, that would increase the lines of code in the ci/cd file :thinking: 
Due to some bad experience with duplicated code I would still tend to altering, but also get your point.

CONTRIBUTING.md
```diff
@@ -39,7 +39,7 @@ In order to review the effects of your changes, you should create your own Kuber
 In case you make changes to the Connaisseur Docker image itself or code for that matter, you need to re-build the image and install it locally for testing. This requires a few steps:
 
 1. In `helm/values.yaml`, set `imagePullPolicy` to `IfNotPresent`.
-2. Configure your local environment to use the Kubernetes docker daemen. In Minikube, this can be done via `eval (minikube docker-env)`.
+2. Configure your local environment to use the Kubernetes docker daemon. In Minikube, this can be done via `eval (minikube docker-env)`.
```
annekebr, line 42: The integration test makes the following lines obsolete doesn't it? https://github.com/sse-secure-systems/connaisseur/pull/42/files#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055R71

We could mention in the Contributing Guidelines that we encourage to add integration testing for new features maybe

Starkteetje:
Pushed typo/capitalization to another PR so as not to clutter this one: #48

CONTRIBUTING.md
Starkteetje: Link doesnt work for me. Found some lines you might've meant though. Can you check whether new changes work for you?

annekebr: Exactly! Thanks and sorry for the broken link!

annekebr (approved):

PR #44, opened 2020-11-02
 ci/docs: Release concept for Connaisseur image and release check on new tagged commit on master branch

Starkteetje:
Previously we had no automatic releases. Now a delegation key is used to sign the images, which will subsequently be pushed to the repository with a technical user. Images will be released for each new tag on the master branch. It is ensured that the tag matches the versions defined in both the Makefile and the helm deployment.

TODO:
- [x] Add real integration test to release pipeline
- [ ] Agree on way forward
- [ ] Set delegation key
- [ ] Complete test

Discussion:
Idea is to have a PR to master that increments version number. After merge, one needs to `git tag` the new master commit and push the tag. Then the release pipeline would run
In my mind the PR would be dedicated and only increment the version and add to a Changelog.md

Are you fine with releasing this way? Do we want to do something different? Do you see potential problems / areas to improve?

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -1,0 +1,53 @@
+name: release
+
+on:
+  push:
+    tags: "*"
+    branches: master
+
+jobs:
+  version-match:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install yq
+        run: sudo snap install yq
+      - name: Ensure version equality
+        run: |
+          if [[ "$(echo $GITHUB_REF | sed s+refs/tags/++)" == "" ]] \
+          || [[ "$(echo $GITHUB_REF | sed s+refs/tags/++)" != "$(yq r helm/values.yaml 'deployment.image' | cut -d ':' -f2)" ]]
+            then exit 1
+          fi
+
+  integration-test:
+    runs-on: ubuntu-latest
+    needs: [version-match]
+    steps:
+      - uses: actions/checkout@v2
+
+      - name: Install yq and bash
+        run: |
+          sudo snap install yq
+          sudo apt update
+          sudo apt install bash -y
+
+      - name: Pull Connaisseur images
+        run: |
+          DOCKER_CONTENT_TRUST=1 docker pull $(yq r helm/values.yaml 'deployment.image')
+          DOCKER_CONTENT_TRUST=1 docker pull $(yq r helm/values.yaml 'deployment.image' | cut -d ':' -f1):helm-hook
+
+      - name: Create KinD cluster
+        run: |
+          GO111MODULE="on" go get sigs.k8s.io/kind
+          kind create cluster --wait 120s
+
+      - name: Check KinD cluster
+        run: kubectl get nodes
+
+      - name: Add images to KinD
+        run: |
+          kind load docker-image $(yq r helm/values.yaml 'deployment.image')
+          kind load docker-image $(yq r helm/values.yaml 'deployment.image' | cut -d ':' -f1):helm-hook
+
+      - name: Run actual integration test
+        run: bash connaisseur/tests/integration/integration-test.sh
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,6 @@
 NAMESPACE = connaisseur
-IMAGE = $(IMAGE_NAME):$(TAG)
-IMAGE_NAME = securesystemsengineering/connaisseur
-TAG = v1.4.0
+IMAGE := $(shell yq r helm/values.yaml 'deployment.image')
+IMAGE_NAME := $(shell echo $(IMAGE) | cut -d ':' -f1)
 
 .PHONY: all docker certs install unistall upgrade annihilate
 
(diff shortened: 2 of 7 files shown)


Makefile
```diff
@@ -1,9 +1,8 @@
 NAMESPACE = connaisseur
-IMAGE = $(IMAGE_NAME):$(TAG)
-IMAGE_NAME = securesystemsengineering/connaisseur
-TAG = v1.3.0
+IMAGE := $(shell yq r helm/values.yaml 'deployment.image')
+IMAGE_NAME := $(shell echo $(IMAGE) | cut -d ':' -f1)
 
-.PHONY: all docker certs install unistall upgrade annihilate
+.PHONY: all docker certs install unistall upgrade annihilate ayy
```
phbelitz: the `ayy` was in there for testing purposes, pls delete

phbelitz (changes requested):

Makefile
phbelitz: arrrr too fast

setup/README.md
```diff
@@ -38,6 +38,8 @@ git clone https://github.com/sse-secure-systems/connaisseur.git
 cd connaisseur
 ```
 
+> The default of what you will install when following this guide will always be the latest release of Connaisseur. If you want to run Connaisseur with the most recent changes that haven't yet made it into a release, you will have to build an image yourself with `make docker`.
```
xopham, line 41: would refer to the contributing guide here, as it sketches how to setup your own env: https://github.com/sse-secure-systems/connaisseur/blob/master/CONTRIBUTING.md#setup-the-environment

xopham:
> Previously we had no automatic releases. Now a delegation key is used to sign the images, which will subsequently be pushed to the repository with a technical user. Images will be released for each new tag on the master branch. It is ensured that the tag matches the versions defined in both the Makefile and the helm deployment.
> 
> TODO:
> 
>     * [x]  Add real integration test to release pipeline
> 
>     * [ ]  Agree on way forward
> 
>     * [ ]  Set delegation key
> 
>     * [ ]  Complete test
> 
> 
> Discussion:
> Idea is to have a PR to master that increments version number. After merge, one needs to `git tag` the new master commit and push the tag. Then the release pipeline would run
> In my mind the PR would be dedicated and only increment the version and add to a Changelog.md
> 
> Are you fine with releasing this way? Do we want to do something different? Do you see potential problems / areas to improve?

I have a few questions:
- do we automatically create the changelog.md form the commit messages? The general requirement was to make explicit and clear commit messages. Otherwise, I could see that we might push releases in case nobody wants to go through all of the past commits. Adding a second point at which we have to make notes to a changelog for each commit seems a bit excessive.
- tag is envisioned as version number, e.g. `1.3.1`?
- do we go with annotated tags?
- how do we decide on new tags? Every week, every time we 'feel' it is about time...
- think we need to have a bit more of an idea of the process. Like which detailed steps happen and where are we required to do manual action, collect information, ... which often is failure prone
- would we release tags?

.github/workflows/release.yaml
```diff
@@ -0,0 +1,43 @@
...
+      - uses: actions/checkout@v2
+      - name: Login to DockerHub
+        uses: docker/login-action@v1
+        with:
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_PASSWORD }}
+      - name: Prepare docker content trust
+        run: mkdir -p ~/.docker/trust/private && echo "${{ secrets.DOCKER_KEY }}" > ~/.docker/trust/private/${{ secrets.DOCKER_KEY_NAME }}
+      - name: Build image
+        run: docker build -t $(yq r helm/values.yaml 'deployment.image') -f docker/Dockerfile .
```
xopham, line 65: would opt for building with DCT and force us to use a signed base image. What do you think?

xopham (commented):
Like the general approach so far, but still feel like I need to understand the whole concept.

I think we should have periodic scans on the latest release as well.

Great progress!

.github/workflows/release.yaml
```diff
@@ -0,0 +1,43 @@
...
+
+jobs:
+  version-match:
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install yq
+        run: sudo snap install yq
+      - name: Ensure version equality
+        run: if [[ "$(echo $GITHUB_REF | sed s+refs/tags/++)" == "" ]] || [[ "$(echo $GITHUB_REF | sed s+refs/tags/++)" != "$(yq r helm/values.yaml 'deployment.image' | cut -d ':' -f2)" ]]; then exit 1; fi
```
xopham, line 16: line too long, should be linebroken. Same below.

xopham (commented):
couldn't we set the Versions via the tag value instead of comparing?

.github/workflows/release.yaml
Starkteetje: Would say that's definetly out of scope for this PR.
Afaik DCT isnt spread far enough that we could just do that, except if we signed our own copies of some base image. That would make the build less trustworthy from an outsider's point though...

xopham (dismissed):
Add versioning for helm-hook

Starkteetje:
> Add versioning for helm-hook

Tracked in #67

phbelitz (approved):

Issue #41, opened 2020-11-01
Move connytest images into same docker repo as connaisseur


xopham:
**Describe the feature**
Currently, the test images are hosted on docker.io/connytest/testimage. These should be moved into docker.io/securesystemsengineering/connaisseur or docker.io/securesystemsengineering/connytest

**Optional: Is your feature request related to a problem? Please describe.**
Would be more consistent than having a separate repo. However, makes the path much longer. 

**Optional: Additional context**
Would require quite some changes and techincally also an update of the demo gif.

Starkteetje:
Note that the integration test also references these images

Starkteetje:
Should also be done for the references in unittests, i.e. https://github.com/sse-secure-systems/connaisseur/blob/master/connaisseur/tests/test_mutate.py#L17 even though they may not actually be pulled at this point in time

PR #49, opened 2020-11-04
fix: validate images of initContainers

annekebr:
Currently, Connaisseur only validates the images referenced 'containers'. This adds validation for images referenced in 'initContainers'.

### DIFF
--- a/connaisseur/tests/test_mutate.py
+++ b/connaisseur/tests/test_mutate.py
@@ -18,6 +18,22 @@ request_obj_pod = {
         ]
     },
 }
+request_obj_pod_with_init_container = {
+    "kind": "Pod",
+    "apiVersion": "v1",
+    "metadata": {},
+    "spec": {
+        "containers": [
+            {"name": "test-connaisseur", "image": "phbelitz/charlie-image:test"}
+        ],
+        "initContainers": [
+            {
+                "name": "init-container",
+                "image": "docker.io/abreust/testing_conny:unsigned",
+            }
+        ],
+    },
+}
 request_obj_cronjob = {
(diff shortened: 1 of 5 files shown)


connaisseur/tests/test_mutate.py
```diff
@@ -18,6 +18,22 @@
         ]
     },
 }
+request_obj_pod_with_init_container = {
+    "kind": "Pod",
+    "apiVersion": "v1",
+    "metadata": {},
+    "spec": {
+        "containers": [
+            {"name": "test-connaisseur", "image": "phbelitz/charlie-image:test"}
```
Starkteetje, line 27: I'd say similarly to moving the demo images to sse, we should also move these test images to sse repo, but might be out of scope here

annekebr: Good point! However, at least for the tests I added, it doesn't fully apply as this just parses the values, so I could use any name I want and the repo would not even need to exist. We don't have test images in the sse repo yet, do we?

Starkteetje: Nope, added comment to #41 and will resolve here

phbelitz (dismissed):
looks good. Thought I'd be more work, but I guess it was easy after all :muscle:

phbelitz:
for the integration test, maybe only check if the `kubectl apply` commands are successful (so check if `$?` == 0) and if the string "admission webhook "connaisseur-svc.connaisseur.svc" denied the request" is contained in the error message (string='My long string';if [[ $string == *"My long"* ]]; then), instead of requiring a specific message.

phbelitz (approved):

PR #50, opened 2020-11-04
ci: Add check for formatting

Starkteetje:
Previously, we claimed to want to format code with black prior to pushing. However, we regularly introduced code that was not formatted properly. This introduces a CI check to ensure pipeline failure if autoformatting would change the code.


Starkteetje:
See https://github.com/sse-secure-systems/connaisseur/actions/runs/345277001
vs https://github.com/sse-secure-systems/connaisseur/runs/1352115335

annekebr (approved):
Nice! I like you took initiative here :+1:

PR #52, opened 2020-11-05
docs: Shell compatibility

### LINKED ISSUE #36: setup guide only works with bash and is not explicit enough otherwise
**Describe the bug**
The steps to set the values in helm/values.yml in the setup guide (https://github.com/sse-secure-systems/connaisseur/tree/master/setup#3-configure-connaisseur) only work for bash and are not widely compatible.

**Expected behavior**
A more generic explanation should be added to make sure the setup is independent of the shell or OS.

**Optional: To reproduce**
Use e.g. Zsh to run the steps under: https://github.com/sse-secure-systems/connaisseur/tree/master/setup#3-configure-connaisseur

 na

na

Starkteetje:
Removes bash-specifics and removes jq dependency for ACR

Fix #36

### DIFF
--- a/setup/README.md
+++ b/setup/README.md
@@ -100,7 +100,7 @@ Lastly, we need to configure the `notary.rootPubKey` that serves as a trust anch
     -----END PUBLIC KEY-----
 ```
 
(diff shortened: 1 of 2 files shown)


xopham (approved):
real nice! like it

PR #56, opened 2020-11-06
ci: add build step

### LINKED ISSUE #54: Add build step in CI
**Describe the bug**
Currently, the connaisseur image is built multiple times in the CI for different purposes (e.g. trivy scan, integration test). This has several disadvantages:
- time cost for building the image each time
- code duplication / higher maintanence due different solutions to build image
- if build fails, the specific scan or test fails which happens late in the ci, is non-intuitive and costs time

**Expected behavior**
Specific build step before or after tests to make sure image can be built, save time for later steps and reduce maintenance and code duplication.

**Optional: To reproduce**
See: https://github.com/sse-secure-systems/connaisseur/actions/runs/349660519

xopham:
Originally, the Connaisseur docker image is built in multiple later steps of the pipeline. This causes code duplication, maintenance increase, unintuitive failing and unnecessary ci run time.

Fixes #54

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -1,0 +1,6 @@
+# ignore Markdown files
+*.md
+# ignore tests
+connaisseur/tests/
+# ignore img
+img/
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -27,6 +27,7 @@ jobs:
         with:
           name: images
           path: images
+          retention-days: 1
 
   black:
     runs-on: ubuntu-latest
@@ -56,7 +57,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install packages
-        run: pip3 install -r requirements.txt
+        run: pip3 install -r requirements_dev.txt
       - name: Lint
         run: cd connaisseur && pylint --ignore-patterns=tests,coverage *.*
 
@@ -68,7 +69,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install packages
-        run: pip3 install -r requirements.txt && pip3 install .
+        run: pip3 install -r requirements_dev.txt && pip3 install .
       - name: Test
         run: cd connaisseur && pytest --cov=connaisseur --cov-report=xml tests/
       - name: Upload code coverage
@@ -104,7 +105,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install packages
-        run: pip3 install -r requirements.txt
+        run: pip3 install -r requirements_dev.txt
       - name: Freeze packages
         run: pip3 freeze > actual_package_versions.txt
       - name: Install safety
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -10,6 +10,24 @@ on:
       - develop
 
 jobs:
+  build:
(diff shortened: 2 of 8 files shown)


.github/workflows/cicd.yaml
```diff
@@ -113,58 +128,66 @@ jobs:
...
+          docker load -i ${GITHUB_SHA}_image.tar
+          docker load -i ${GITHUB_SHA}_hook.tar
       - name: Install yq and bash
         run: |
           sudo snap install yq
           sudo apt update
           sudo apt install bash -y
-
-      - name: Build Connaisseur image
+      - name: Set version
```
Starkteetje, line 176: Since the vars set in this are only used during `make docker` this section can be dropped completely

Starkteetje (dismissed):

.github/workflows/cicd.yaml
```diff
@@ -13,6 +13,25 @@ env:
...
+      - name: Save images
+        run: |
+          mkdir images
+          docker save ${RELEASE_IMAGE}:${RELEASE_VERSION} -o images/${GITHUB_SHA}_image.tar
+          docker save ${RELEASE_IMAGE}:helm-hook -o images/${GITHUB_SHA}_hook.tar
+      - uses: actions/upload-artifact@v2
+        with:
+          name: images
+          path: images
+
```
Starkteetje, line 34: Would add retention-days: 1

.dockerignore
```diff
@@ -0,0 +1,4 @@
+# ignore Markdown files
+*.md
+# ignore tests
+connaisseur/tests/
```
Starkteetje, line 4: If you want marginally faster builds, you can also add `img/` since that accounts for ~2/3 of the repo size

.github/workflows/cicd.yaml
```diff
@@ -13,6 +13,25 @@ env:
...
+      - name: Build Connaisseur image
+        run: |
+          sed -i "s+TAG =.*+TAG = ${RELEASE_VERSION}+" Makefile
+          sed -i "s+IMAGE_NAME =.*+IMAGE_NAME = ${RELEASE_IMAGE}+" Makefile
+          make docker
+      - name: Save images
+        run: |
+          mkdir images
+          docker save ${RELEASE_IMAGE}:${RELEASE_VERSION} -o images/${GITHUB_SHA}_image.tar
+          docker save ${RELEASE_IMAGE}:helm-hook -o images/${GITHUB_SHA}_hook.tar
```
Starkteetje, line 29: Maybe out of scope for this PR, but we are not versioning the helm-hook here, so we technically can't change the hook script without changing naming here (or we break older versions on redeployment), right?

.dockerignore
xopham: think that is excluded anyways. Checking...

Starkteetje: For build size, it is, since its not used. But for speed during building, the whole context (the . for docker build . ) is copied to the docker daemon. Is local copy of few MB, so :man_shrugging: , but if we put a 1GB file in that folder you'll notice

xopham: will however only increase the build speed, but good point!

.github/workflows/cicd.yaml
xopham: added to release discussion

phbelitz (approved):

PR #64, opened 2020-11-17
fix: Add output during installation and increase sentinel timeout

annekebr:
Currently, 'helm install' does not only not reveal much information about the progress but also does not  proceed anymore as soon as the sentinel has completed. Thus, in case the sentinel times out without connaisseur being ready, the last 4 minutes waiting for the helm timeout are wasted.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,14 @@ certs:
 install: certs
 	kubectl create ns $(NAMESPACE) || true
 	kubectl config set-context --current --namespace $(NAMESPACE)
-	helm install connaisseur helm --wait
+	#
+	#=============================================
+	#
+	# The installation may last up to 5 minutes.
+	#
+	#=============================================
+	#
+	helm install connaisseur helm --atomic
 
 uninstall:
 	kubectl config set-context --current --namespace $(NAMESPACE)
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -1,7 +1,7 @@
 # configure connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v1.4.3
+  image: securesystemsengineering/connaisseur:v1.4.4
   imagePullPolicy: Always
   resources: {}
   # limits:
@@ -13,9 +13,6 @@ deployment:
   nodeSelector: {}
   tolerations: []
   affinity: {}
-  # bootstrap sentinel timeout in seconds
-  # deployment will fail if Connaisseur pods do not report ready within this timeframe
-  sentinelTimeout: 60
 
 # configure connaisseur service
 service:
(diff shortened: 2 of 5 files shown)


Makefile
```diff
@@ -17,7 +17,14 @@ certs:
...
 	kubectl config set-context --current --namespace $(NAMESPACE)
-	helm install connaisseur helm --wait
+	#
+	#=============================================
+	#
+	# The installation may last up to 5 minutes.
+	#
+	#=============================================
+	#
+	helm install connaisseur helm --atomic
```
Starkteetje, line 27: I like the idea of the `--atomic`, but since there's resources like the webhook, which are not directly deployed, would it be advantageous to have something along the lines of `helm instal || make uninstall` instead (or additionally), which would also take care of indirectly created resources?

Makefile
```diff
@@ -17,7 +17,14 @@ certs:
...
 	kubectl create ns $(NAMESPACE) || true
 	kubectl config set-context --current --namespace $(NAMESPACE)
-	helm install connaisseur helm --wait
+	#
+	#=============================================
+	#
+	# The installation may last up to 5 minutes.
+	#
+	#=============================================
+	#
```
peterthomassen, line 26: Why the large comment? Is it displayed during install, or only visible in the file source code? (not sure how `make` behaves)

annekebr: The former, it is displayed as you see it here- I just wanted to make sure that the user reads it :grinning:

helm/values.yaml
```diff
@@ -1,7 +1,7 @@
 # configure connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v1.4.0
+  image: securesystemsengineering/connaisseur:v1.3.1
```
peterthomassen: that's a downgrade. Is that intended?

annekebr: it's changed in the latest commit :+1:

Makefile
annekebr: I don't have a simple solution for that in mind and I think that adding `--atomic` already improves the installation process in some cases as you can run `make install` even if a previous `make install` failed.

helm/values.yaml
peterthomassen: I don't see why any intermediate commit would lower the version number...?

annekebr:
@Starkteetje I reverted your changes because I thought the configuration option does not really make sense anymore if the timeout is set to 5 min :cry: Feel free to block this PR if you think otherwise!

helm/values.yaml
annekebr: I reverted one of Teetjes commit and reordered commits - the revert commit changed the version number.

phbelitz (approved):

PR #73, opened 2020-12-14
build: Added versioning to helm-hook-image

### LINKED ISSUE #67: Helm-hook image is not versioned
**Describe the bug**
The helm-hook image has no version information, so it cannot be updated without endangering past installations of Connaisseur

E.g. can't rebuild image due to base image vulnerabilites without oerwriting prior image

phbelitz:
Minor changes to the helm-hook-image could break prvious versions of connaisseur, which is why versioning is introduced to circumvent this.

fixes #67

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 NAMESPACE = connaisseur
 IMAGE := $(shell yq r helm/values.yaml 'deployment.image')
-IMAGE_NAME := $(shell echo $(IMAGE) | cut -d ':' -f1)
+HELM_HOOK_IMAGE := $(shell yq r helm/values.yaml 'deployment.helmHookImage')
 
 .PHONY: all docker certs install unistall upgrade annihilate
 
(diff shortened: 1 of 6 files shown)


peterthomassen (dismissed):
lgtm!

Makefile
```diff
@@ -1,14 +1,14 @@
 NAMESPACE = connaisseur
 IMAGE := $(shell yq r helm/values.yaml 'deployment.image')
-IMAGE_NAME := $(shell echo $(IMAGE) | cut -d ':' -f1)
+HELM_HOOK_IMAGE := $(shell yq r helm/values.yaml 'deployment.hook-image')
```
Starkteetje, line 3: Why is this HELM_HOOK_IMAGE and in release.yml it's hook-image? Guess we should have one name for it, either with or without helm

phbelitz: Sure. Changed to `HOOK_IMAGE`.

Starkteetje (dismissed):

Makefile
phbelitz: Eh. What about the actual tag of the hook image? it's `helm-hook-v1.0` curently. change that too? thoughts?

Starkteetje: You're right. Would go for consistency. But maybe then since the helm-hook name is more prominent due to the image, I'd rename the vars to include helm

phbelitz: done.

Starkteetje (approved):

PR #75, opened 2020-12-16
fix: Fixed JSON validation schema for trust data

### LINKED ISSUE #74: Regex in schema json prevent connaisseur to validate signers with dash
**Describe the bug**

We have a container registry in azure where the signer's name is "mysigner-dev".

When we try to deploy an application from this registry the **snapshot.json** returns the list of the targets it includes the **target/mysigner-dev** the image is rejected with this error message:

```
ERROR: {'message': 'trust data has invalid format.', 'context': {'trust_data_type': 'snapshot', 'user': 'system:serviceaccount:ns:sa', 'operation': 'CREATE', 'kind': 'Deployment', 'namespace':
'mywebapp', 'name': 'mywebapp', 'image': 'myregistry.azurecr.io/webapp:0.1.20.0'}, 'detection_mode': True}
```

After some research, it appears that the code uses a regex which does not handle anything but words.

**Expected behavior**

Images are not rejected.

**Optional: To reproduce**

Add a signer with a "-" in its name or any char which is not a word as per the regex provided in the snapshot schema:

```
                "meta": {
                    "type": "object",
                    "patternProperties": {
                        "^(root|targets(\/\\w*)?)$": {
```

This can be easily reproduced through your unit tests.

 - Kubernetes Cluster: k3s 1.19.4
 - Notary Server: azure
 - Container registry: azure
 - Connaisseur: latest (1.4.0)

------

Ps: Your tool is really great and well made, thanks for the work you put into it! 💯

phbelitz:
The JSON validation schema for the trust data doesn't allow special characters like '-' in delegation names, thus failing the validation even though it should be valid. This has been fixed.

fixes #74


connaisseur/res/snapshot_schema.json
```diff
@@ -20,7 +20,7 @@
                 "meta": {
                     "type": "object",
                     "patternProperties": {
-                        "^(root|targets(\/\\w*)?)$": {
+                        "^(root|targets(\/[\\w_:\\.-]*)?)$": {
```
xopham, line 23: is this the spec?

connaisseur/res/targets_schema.json
```diff
@@ -75,7 +75,7 @@
                                     },
                                     "name": {
                                         "type": "string",
-                                        "pattern": "^targets\/\\w*$"
+                                        "pattern": "^targets\/[\\w_:\\.-]*$"
```
xopham, line 78: see above

xopham (commented):
lgtm

we might want to push this to develop as it fails for bandit due to known issues fixed on develop. Could do a standard release tomorrow afternoon

connaisseur/res/snapshot_schema.json
phbelitz: According to the spec, it can be any string. https://github.com/theupdateframework/tuf/blob/f7695dace85444041489b83d5a66cd39c761bbd6/tuf/formats.py#L113

So i guess i'll change it to `"^(root|targets(\/.*)?)$"` ?

peterthomassen: We definitely should exempt some characters, like `/` (i.e. `[^/]`) or null bytes. I clearly vote for listing allowed characters explicitly and not use `.` though, otherwise you must be very certain to have thought of all characters to exempt.

peterthomassen (changes requested):
see my code comment

connaisseur/res/snapshot_schema.json
phbelitz: ok so back to `\w_:\.-` ? these are all that came to my mind

peterthomassen: I would say so. We could add `+`, `;`, `~` and the like, but I wouldn't assume anyone does that. What do others think? @annekebr @Starkteetje 

Also, `\w` includes `_`, and `.` does not need to be escaped when inside `[]`. (Inside a character class, each character stands for itself, except `\w` and the like, and `-` which denotes ranges unless it appears at the beginning or end of the character class.)

Starkteetje: AnyString allows whitespace, so we certainly dont want that

Doing `docker trust key generate something` I get

```
key name "hallo du" must start with lowercase alphanumeric characters and can include "-" or "_" after the first character
key name "hallo++du" must start with lowercase alphanumeric characters and can include "-" or "_" after the first character
key name "halloDu" must start with lowercase alphanumeric characters and can include "-" or "_" after the first character
```
so I guess, dash and underscore are sufficient in addition to lowercase alphanum

peterthomassen (approved):

peterthomassen (approved):
lgtm!

PR #78, opened 2020-12-22
ci: Improve tag detection during release

Starkteetje:
Previously we used the GITHUB_REF variable set during the execution of of Github Actions to get to the current tag, which failed if tagging was done on prior to the push/on a different branch. Now we use the information contained in the git repository that is checked out anyway. Additionally the release pipeline previously also ran for any new commit on master (without a new tag being present) and now runs only for tags in potentially any branch (which is fine, as the release pipeline does no privileged action).

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -2,8 +2,7 @@ name: release
 
 on:
   push:
-    tags: "*"
-    branches: master
+    tags: "v*"
 
 jobs:
   version-match:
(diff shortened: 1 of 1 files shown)


phbelitz:
Any example pipeline where this was successfully tested?

Starkteetje:
> Any example pipeline where this was successfully tested?

https://github.com/sse-secure-systems/connaisseur/runs/1594978442?check_suite_focus=true

phbelitz (approved):

PR #80, opened 2021-01-15
fix: ensure compatibility with new yq v4

annekebr:
The new version of yq introduced a breaking change when reading from a yaml file. This ensures compatibility with the latest version which we should use as v3 is marked as deprecated and will run out of support in August.


connaisseur/tests/integration/integration-test.sh
```diff
@@ -4,9 +4,9 @@ set -euo pipefail
 # This script is expected to be called from the root folder of Connaisseur
 
 echo 'Preparing Connaisseur config...'
-export RELEASE_IMAGE=$(yq r helm/values.yaml 'deployment.image')
+export RELEASE_IMAGE=$(yq e '.deployment.image' helm/values.yaml)
 envsubst < connaisseur/tests/integration/update-config.yaml > update
-yq write --inplace --script update helm/values.yaml # Coincidentally, this also ensure all used env variables are set
```
annekebr, line 9: @Starkteetje This breaks the coincidental checking for unset env vars in `update-config.yaml`. Do you have any idea how to reintroduce it here?

README.md
```diff
@@ -83,7 +83,8 @@ Connaisseur can be integrated into a multitude of different systems, with a few
 1. **Requirements**: A few tools are needed in advance. You'll need `git` for getting the source code and `make` for a convenient setup. Since Connaisseur will be deployed as a container to a Kubernetes cluster, [docker](https://docs.docker.com/v17.09/engine/installation/linux/docker-ce/ubuntu/), [helm](https://helm.sh/docs/intro/install/) and [kubectl](https://kubernetes.io/docs/tasks/tools/install-kubectl/) are required for building the image, generating a Deployment file and deploying it to the cluster. During the deployment, some certificates need to be created, which requires `openssl`. To install everything except docker, helm and kubectl, use
 
 ```bash
-sudo apt install git make openssl yq
+sudo apt install git make openssl
```
Starkteetje, line 86: Do we want to put -y options now that this is multiline?

setup/README.md
```diff
@@ -125,7 +125,7 @@ openssl ec -in root-priv.key -pubout -out root.pub
 - After entering your password, copy the public key to the `helm/values.yaml`: 
 
 ```bash
-yq write --inplace ${OLDPWD}/helm/values.yaml -- notary.rootPubKey "$(cat root.pub)"
+yq e '.notary.rootPubKey=$(cat root.pub)' -i ${OLDPWD}/helm/values.yaml
```
Starkteetje, line 128: Was also present before, but if OLD_PW contains whitespace, does this break?

connaisseur/tests/integration/integration-test.sh
Starkteetje:
```yq eval-all --null-input ".a=env(myenv)"
Error: Value for env variable 'myenv' not provided in env()```
seems like it will reject them if defined with `env`.Might work with files as well 🤷

File: README.md (hunk 2 above)
Comment (annekebr) [reply README.md:86]: Hm, that doesn't help because you need to enter your passphrase, does it? Are you able to install `yq`via apt? It would be better to have it in one line, but apt wasn't able to locate `yq`...

File: setup/README.md (hunk 3 above)
Comment (annekebr) [reply setup/README.md:128]: just tried it. You're right, it fails! Can we address this case by code or should we add a hint in the paragraph above this snippet?

File: connaisseur/tests/integration/integration-test.sh (hunk 1 above)
Comment (annekebr) [reply connaisseur/tests/integration/integration-test.sh:9]: no, I tested. However, what are we trying to achieve here actually? Am I reading it correctly, that we cut the deployment image from the `helm/values.yml` to update the `helm/values.yml` with its value? What am I missing?

File: setup/README.md (hunk 3 above)
Comment (peterthomassen) [reply setup/README.md:128]: You can emded the path string in quotes

**Comment by peterthomassen** [comment]
I don't feel comfortable reviewing/approving this because I don't know enough about helm. @Starkteetje Would you mind taking this on?

File: README.md (hunk 2 above)
Comment (Starkteetje) [reply README.md:86]: Would probably still put the `-y` so it works if run from elevated context

File: connaisseur/tests/integration/integration-test.sh (hunk 1 above)
Comment (Starkteetje) [reply connaisseur/tests/integration/integration-test.sh:9]: We also update the root key and ensure some other settings as described by https://github.com/sse-secure-systems/connaisseur/blob/master/connaisseur/tests/integration/update-config.yaml
But yes, the env variable depended stuff has apparently become obsolete

File: helm/values.yaml [hunk 4]
```diff
@@ -1,7 +1,7 @@
 # configure connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v1.4.5
+  image: securesystemsengineering/connaisseur:v1.4.6
```
Starkteetje, line 4: Since we did not release 1.4.5, there shouldn't be an increment yet, right?

Starkteetje (approved):

PR #81, opened 2021-01-15
fix: missing delegation files

phbelitz:
in the special case, that delegations haven been added to an imgae repository, but no signatures were created using said delegations, the delegations appear in the targets JSON, but the actual delegation JSON files don't exist yet. Previously connaisseur assumed them to be present if they are in the targets JSON. This was changed to use the targets in the targets JSON file, instead of the missing delegation JSON files.

### DIFF
--- a/connaisseur/notary_api.py
+++ b/connaisseur/notary_api.py
@@ -97,6 +97,18 @@ def get_trust_data(host: str, image: Image, role: TUFRole, token: str = None):
     return TrustData(data, role.role)
 
 
+def get_delegation_trust_data(
+    host: str, image: Image, role: TUFRole, token: str = None
+):
+    if os.environ.get("LOG_LEVEL", "INFO") == "DEBUG":
+        return get_trust_data(host, image, role, token)
+
+    try:
+        return get_trust_data(host, image, role, token)
+    except Exception:
+        return None
+
+
 def parse_auth(auth_header: str):
     """
     Generates an URL from the 'Www-authenticate' header, where a token can be
--- a/connaisseur/validate.py
+++ b/connaisseur/validate.py
@@ -2,7 +2,7 @@ import base64
 from connaisseur.image import Image
 from connaisseur.key_store import KeyStore
 from connaisseur.util import normalize_delegation
-from connaisseur.notary_api import get_trust_data
+from connaisseur.notary_api import get_trust_data, get_delegation_trust_data
 from connaisseur.tuf_role import TUFRole
 from connaisseur.exceptions import (
     NotFoundException,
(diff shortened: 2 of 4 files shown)


connaisseur/notary_api.py
```diff
@@ -97,6 +97,15 @@ def get_trust_data(host: str, image: Image, role: TUFRole, token: str = None):
     return TrustData(data, role.role)
 
 
+def get_delegation_trust_data(
+    host: str, image: Image, role: TUFRole, token: str = None
+):
+    try:
+        return get_trust_data(host, image, role, token)
+    except Exception:
```
peterthomassen, line 105: You probably want only `connaisseur.exceptions.NotFoundException`?

phbelitz: welp. the problem is that we are actually getting a `500 internal server error` back, even though the problem should be, that the requested files do not exist, meaning theoretically it should be `404 not found`. so the `NotFoundException` is not raised here. instead of grabbing `500` now in the `get_trust_data` function, we decided to grab all exceptions instead, as we are not sure what other kind of other weird behavior could come up further down the line (since returning 500 in case is already weird).

connaisseur/validate.py
```diff
@@ -124,14 +128,29 @@ def process_chain_of_trust(host: str, image: Image, req_delegations: list):
...
     if req_delegations:
+        for target_role in req_delegations:
+            if not trust_data[target_role]:
+                raise NotFoundException(
+                    'no trust data for delegation role "{}" for image "{}"'.format(
+                        target_role, str(image)
+                    ),
+                    {"tuf_role": target_role},
+                )
+
```
peterthomassen, line 143: You can make this a bit more neat as

```python
if not all(trust_data[target_role] for target_role in req_delegations):
    tuf_roles = [target_role in req_delegations if not trust_data[target_role]]
    msg = f"no trust data for delegation roles {tuf_roles} for image {image}"
    raise NotFoundException(msg, {"tuf_roles": tuf_roles})
```

or, weirder,

```python
if tuf_roles := [target_role in req_delegations if not trust_data[target_role]]:
    msg = f"no trust data for delegation roles {tuf_roles} for image {image}"
    raise NotFoundException(msg, {"tuf_roles": tuf_roles})
```

I think nobody would understand the second, and I really dislike inline assignments like that.

connaisseur/notary_api.py
peterthomassen: hm. Maybe catch all in production, but only specific stuff in debug mode? That way you can learn more about exceptions in debug mode.

phbelitz: done.

annekebr (approved):

connaisseur/notary_api.py
```diff
@@ -97,6 +97,18 @@ def get_trust_data(host: str, image: Image, role: TUFRole, token: str = None):
...
+def get_delegation_trust_data(
+    host: str, image: Image, role: TUFRole, token: str = None
+):
+    if os.environ.get("LOG_LEVEL", "INFO") == "DEBUG":
+        return get_trust_data(host, image, role, token)
+
+    try:
+        return get_trust_data(host, image, role, token)
+    except Exception:
+        return None
```
peterthomassen, line 109:
```python
try:
    return get_trust_data(host, image, role, token)
except:
    if os.environ.get("LOG_LEVEL", "INFO") == "DEBUG":
        raise
    return None
```

phbelitz: will be done in #80 ... hehe ...

PR #83, opened 2021-01-21
Feat/multi root key

phbelitz:
In order to verify signatures that were created by yourself and signatures from others, multiple public root keys must be provisioned. 

Previous behavior: A single root key was provisioned and added to the internal `key_store` as `root`. When verifying any trust data file of type root, this `root` key was used, regardless of the key ID declared in the trust data.

New behavior: One or more root keys are provisioned, each given with a key ID and stored in the internal `key_store` under their key ID. When verifying any trust data file of type root, the declared key ID used for signing the file is looked up in the `key_store`. If there is no key found an error will be raised, otherwise validation will go forth.

### DIFF
--- a/connaisseur/trust_data.py
+++ b/connaisseur/trust_data.py
@@ -97,9 +97,15 @@ class TrustData:
         """
         msg = json.dumps(self.signed, separators=(",", ":"))
         for signature in self.signatures:
-            key_id = "root" if self.kind == "root" else signature["keyid"]
+            key_id = signature["keyid"]
             pub_key = keystore.get_key(key_id)
             sig = signature["sig"]
+
+            if not pub_key:
+                raise NotFoundException(
+                    "couldn't find right public for trust data.",
+                    {"key_id": key_id, "trust_data_type": self.signed.get("_type")},
+                )
 
             try:
                 verify_signature(pub_key, sig, msg)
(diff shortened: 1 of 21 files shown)


connaisseur/trust_data.py
```diff
@@ -97,10 +97,16 @@ def _validate_signature(self, keystore: KeyStore):
...
         msg = json.dumps(self.signed, separators=(",", ":"))
         for signature in self.signatures:
-            key_id = "root" if self.kind == "root" else signature["keyid"]
+            key_id = signature["keyid"]
             pub_key = keystore.get_key(key_id)
             sig = signature["sig"]
 
+            if not pub_key:
+                raise NotFoundException(
+                    "couldn't find right public for trust data.",
```
Starkteetje, line 106: there's something missing in this sentence

connaisseur/validate.py
```diff
@@ -66,7 +66,9 @@ def get_trusted_digest(host: str, image: Image, policy_rule: dict):
     return digests.pop()
 
 
-def process_chain_of_trust(host: str, image: Image, req_delegations: list):
+def process_chain_of_trust(
+    host: str, image: Image, req_delegations: list
+):  # pylint: disable=too-many-branches
```
Starkteetje, line 71: Maybe it would be better if we refactored the function into more digestible (haha) blocks, instead of turning of linting ;)

phbelitz:
closing this because of PR #90 including these changes

PR #85, opened 2021-01-28
fixup! fix: missing delegation files

phbelitz:
there was an error with the last PR, where we tried to call `validate()` on a `None` object.

### DIFF
--- a/connaisseur/notary_api.py
+++ b/connaisseur/notary_api.py
@@ -100,12 +100,11 @@ def get_trust_data(host: str, image: Image, role: TUFRole, token: str = None):
 def get_delegation_trust_data(
     host: str, image: Image, role: TUFRole, token: str = None
 ):
-    if os.environ.get("LOG_LEVEL", "INFO") == "DEBUG":
-        return get_trust_data(host, image, role, token)
-
     try:
         return get_trust_data(host, image, role, token)
-    except Exception:
+    except Exception as ex:
+        if os.environ.get("LOG_LEVEL", "INFO") == "DEBUG":
+            raise ex
         return None
 
 
--- a/connaisseur/notary_api.py
+++ b/connaisseur/notary_api.py
@@ -97,6 +97,18 @@ def get_trust_data(host: str, image: Image, role: TUFRole, token: str = None):
     return TrustData(data, role.role)
 
 
+def get_delegation_trust_data(
+    host: str, image: Image, role: TUFRole, token: str = None
+):
+    if os.environ.get("LOG_LEVEL", "INFO") == "DEBUG":
+        return get_trust_data(host, image, role, token)
+
+    try:
+        return get_trust_data(host, image, role, token)
+    except Exception:
+        return None
+
+
 def parse_auth(auth_header: str):
     """
     Generates an URL from the 'Www-authenticate' header, where a token can be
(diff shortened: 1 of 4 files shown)


connaisseur/notary_api.py
```diff
@@ -100,12 +100,11 @@ def get_trust_data(host: str, image: Image, role: TUFRole, token: str = None):
 def get_delegation_trust_data(
     host: str, image: Image, role: TUFRole, token: str = None
 ):
-    if os.environ.get("LOG_LEVEL", "INFO") == "DEBUG":
-        return get_trust_data(host, image, role, token)
-
     try:
         return get_trust_data(host, image, role, token)
-    except Exception:
+    except Exception as ex:
```
peterthomassen, line 105: What does this have to do with the bug you're fixing in the other file? -- I've seen this fix in three different PRs now (one by @annekebr). I'm not sure what's going on, but as long as you know, it's fine :-)

connaisseur/validate.py
```diff
@@ -98,9 +98,13 @@ def process_chain_of_trust(
...
             )
 
-    # validate all trust data's signatures, expiry dates and hashes
+    # validate all trust data's signatures, expiry dates and hashes.
+    # when delegation are added to the repository, but weren't yet used for signing, the
+    # delegation files don't exist yet and are `None`. in this case they can't be
+    # validated and must be skipped.
     for role in trust_data:
-        trust_data[role].validate(key_store)
+        if trust_data[role]:
```
peterthomassen, line 106: `if trust_data[role] is not None:`

Otherwise, it depends on the truthiness of the value itself. It may not be a problem here, but it's good practice to be explicit about `None` (or the absence of things in general).

peterthomassen (changes requested):

connaisseur/notary_api.py
phbelitz: ya well i figured i put it in here, as it was supposed to be done in #81 and since this is a fixup for that, well...

peterthomassen (approved):

connaisseur/validate.py
```diff
@@ -92,11 +94,17 @@ def process_chain_of_trust(host: str, image: Image, req_delegations: list):
...
+            )
 
-    # validate all trust data's signatures, expiry dates and hashes
+    # validate all trust data's signatures, expiry dates and hashes.
+    # when delegation are added to the repository, but weren't yet used for signing, the
+    # delegation files don't exist yet and are `None`. in this case they can't be
+    # validated and must be skipped.
     for role in trust_data:
-        trust_data[role].validate(key_store)
+        if trust_data[role] is not None:
```
annekebr, line 106: What happens if an attacker controlling the notary server deletes e.g. the `timestamp.json`? Is the verification of that file skipped in that case?

phbelitz: All the mayor trust data (root, targets, timestamp, snapshot) will be retrieved using the `get_trust_data()` function, which will raise an error if the corresponding trust data can't be found. Meaning they can't be `None` there.

Even if the attacker just deletes the content of the file, while still keeping it, a `TrustData` object is still created, just with empty data. So even in this case it shouldn't be `None`. Only the delegation trust data can be `None`, but those are further checked, should specific delegations be needed, or ignored if not needed.

annekebr: thanks!

peterthomassen: If you find yourself bored, you could create testcases for the behavior that occurs when trust data files are partially missing etc.

peterthomassen: The point being that any later refactoring/change would be prevented from introducing the kind of bug @annekebr is worried about.

phbelitz:
the fixup commit has been integrated into the dev branch

PR #86, opened 2021-01-29
fix: ensure compatibility with new yq v4

annekebr:
The new version of yq introduced a breaking change when reading from a yaml file. This ensures compatibility with the latest version which we should use as v3 is marked as deprecated and will run out of support in August.


annekebr:
Duplicate of #80

PR #88, opened 2021-02-02
docs/test: Documentation for image validation, improve validation test coverage

Starkteetje:
This PR documents the existing validation procedure employed by Connaisseur. In addition, it adds some test cases for all supported and one unsupported Kubernetes resource, tests more invalid trust data and checks the combination of valid and invalid delegations alongside each other. Also typos

### DIFF
--- a/docs/validation.md
+++ b/docs/validation.md
@@ -1,0 +1,28 @@
+# Validation
+
+This document describes the image validation process as employed by Connaisseur.
+
+Since Connaisseur is a Kubernetes AdmissionController it operates on requests and has these requests, the Kubernetes API, the registry and the notary to work with. First off, Connaisseur does some light checks on the semantics of the request it receives, e.g. whether it is capable of using the API version of the admission request.
+
+Then it parses images to deploy from the request. Connaisseur does this for Pods, Jobs, CronJobs, Deployments, DaemonSets, StatefulSets, ReplicaSets and ReplicationControllers. These images are then validated as follows:
+
+1. If the image is deployed as part of an already deployed object (i.e. a Pod gets deployed as a child of a Deployment and the Deployment was already validated), it will be admitted. This is done as the parent (and thus the child) might have been mutated, which could lead to duplicate validation or rule mismatch. For example, given a Deployment which contains Pods with `image:tag` that gets mutated to contain Pods with `image@sha256:digest`. Then a) the Pod would not need another validation as the image was validated during the admittance of the Deployment and b) if there exists a specific rule for `image:tag` and another for `image:*`, then after mutating the Deployment, the Pod would be falsely validated against `image:*` instead of `image:tag`. To ensure the child resource is legit in this case, the parent resource is requested via the Kubernetes API and only those images it lists are accepted.
+2. The best matching rule from the image policy is looked up and if `verify: false`, the image is admitted.
+3. Trust data of the image is consulted as described in the next section.
+
+## Validation of trust data
+
+Validation of trust data on a high level boils down to two steps:
+
+1. Get all trusted (i.e. signed) image digests related to the tag or digest of the image.
+2. If there is exactly one, admit the image.
+
+Regarding the latter step, rejection upon no matching trusted digests is obvious. However, Connaisseur also needs to reject the image if there is more than one trusted digest, since at this point in time Connaisseur doesn't have the ability to distinguish between the right and wrong trusted digest. This only occurs in some edge cases, but nonetheless has to be addressed.
+
+Let's now focus on the integral part of Connaisseur that is how to get all trusted digests for an `image:tag` or `image:digest` combination:
+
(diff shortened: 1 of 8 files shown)


docs/validation.md
```diff
@@ -0,0 +1,28 @@
+# Validation
+
+This document describes the image validation process as employed by Connaisseur.
+
+Since Connaisseur is a Kubernetes AdmissionController it works on requests and has these requests, the Kubernetes API, the registry and and the notary to work with. First off, Connaisseur does some light checks on the semantics of the request it receives, e.g. whether it is capable of using the API version of the admission request.
```
peterthomassen, line 5: `and and`

Also, I find "works on" and "work with" in the same sentence a bit confusing.

peterthomassen, line 5: `and and`

Also, I find "works on" and "work with" in the same sentence a bit confusing.

docs/validation.md
```diff
@@ -0,0 +1,28 @@
...
+
+Let's now focus on the integral part of Connaisseur that is how to get all trusted digests for an `image:tag` or `image:digest` combination:
+
+Connaisseur looks up trust data of the image in the `root`, `snapshot`, `timestamp` and `targets` files by querying the API of the notary server. Trust data syntax is validated against [their known schemas](https://github.com/sse-secure-systems/connaisseur/tree/master/connaisseur/res) (for an introduction to the trust data formats see the [presentation of our colleague Philipp Belitz](https://berlin-crypto.github.io/assets/Berlin%20Crypto%20Connaisseur.pdf)). Then, the files' signatures are validated against the pinned root key for the `root` file and against the respective keys validated in previous steps for later files. Connaisseur further gathers trust data of potential delegations linked in the `targets` file.
+
+At this point, Connaisseur is left with a set of potentially disjoint or overlapping sets of trust data. Connaisseur filters the trust data for digests that actually relate to the image in validation.
```
peterthomassen, line 26: `in validation` --> `under validation`

docs/validation.md
```diff
@@ -0,0 +1,28 @@
...
+
+If the image policy rule that governs the image to be validated does contain a `delegations` field, Connaisseur checks whether one of the delegations' sets of trust data does not contain an entry for the image that is being validated, in which case the request is rejected. Otherwise, Connaisseur builds a set over the union of the digests and proceeds with step 2., i.e. accpeting if the size of the resulting set equals 1.
```
peterthomassen, line 28: `accpeting`

Also, I'd rephrase the first sentence to avoid the negation: "field, Connaisseur makes sure that all delegations' sets of trust data do contain an entry for the image" (don't need to repeat "that is being validated"); "if that is not the case, the request is rejected".

peterthomassen (changes requested):
very concise description of the algorithm! I like the sentence that ends with "validated in previous steps for later files" :)

Personally, I'd say that unrelated things (missing tests, and docs of something else) should go into different commits. I like having the upside of being able to revert a commit (e.g. when you don't want the tests anymore) without loosing something else unintentionally. (I'll approve regardless.)

docs/validation.md
```diff
@@ -0,0 +1,28 @@
...
+3. Trust data of the image is consulted as described in the next section.
+
+## Validation of trust data
+
+Validation of trust data on a high level boils down to two steps:
+
+1. Get all trusted (i.e. signed) image digests related to the tag or digest of the image.
+2. If there is exactly one, admit the image.
+
+Regarding the latter step, rejection upon no matching trusted digests is obvious. However, Connaisseur also needs to reject the image if there is more than one trusted digest since Connaisseur has no context to choose which of the digests the initial tag should be mutated to. Doing so would potentially constitute changing the code you see deployed in your Kubernetes cluster. This means you cannot tag two different images with the same tag (or you need to scrub your notary server of the previous trust data).
```
phbelitz, line 20:
```
This means you cannot tag two different images with the same tag (or you need to scrub your notary server of the previous trust data).
```

To my knowledge, there is only one scenario where 2 different digests could appear for the same image tag and connaisseur can't decide which to take. 

That is when 2 different delegations sign different images under the same image + tag. So let's say delegation 'del1' signs `image:tag` resulting in digest `123` and then 'del2' takes the same image, does some modifications and resigns, resulting in digest `456`. Now in the notary there will be 2 files: *del1.json* that says `image:tag` has digest `123` and *del2.json* that says `image:tag` has digest `456`.
This setup in itself is no problem, as there is an additional file called *releases.json* that has the most up-to-date digest for `image:tag`, which would be `456`. The problem occurs when both 'del1' and 'del2' are required to be signers according to the image policy. The we expect them to have signed the exact same image and look whether the digests in *del1.json* and *del2.json* are the same (which they are not). In this case we would have two different digest and don't know whom to trust more.

**ANYWAYS**, what I'm trying to say is, it's not really an issue that two different images have the same tag. There is just this one edge case, that makes things a bit complicated.

Starkteetje: Ok, so the part of "Conny doesnt have the context" is not quite accurate and should be "at this point we havent given Conny the ability to distinguish (and in some edge case it will fail)"?

phbelitz:
```suggestion
Regarding the latter step, rejection upon no matching trusted digests is obvious. However, Connaisseur also needs to reject the image if there is more than one trusted digest, since at this point in time Connaisseur doesn't have the ability to distinguish between the right and wrong trusted digest. This only occurs in some edge cases, but nonetheless has to be addressed.
```

Something like this? yo?

phbelitz (approved):

peterthomassen (approved):
lgtm!

PR #89, opened 2021-02-02
Fix/refactor: Refactor validate method to not use non-validated data

Starkteetje:
Previously, Connaisseur accepted all trust data files at first and then validated them. This was not an immediate security issue, since the root key could not be overwritten and since the KeyStore is write-once, so keys will only be used after they have been validated. However, Connaisseur would have pulled all delegations in a malicious targets.json without prior validation, which would have allowed an attacker to specify many non-existant delegations, potentially causing a denial of service. This PR fixes the issue by first validating and then processing the trust data files. In addition, the way Connaisseur previously validated trust data files would have allowed an attacker that compromised the long-term snapshot key to mount freeze attacks (i.e. ignoring the validation via timestamp key) by mounting a targeted collision attack instead of a 2nd-preimage attack against the DCT hash function.

### DIFF
--- a/connaisseur/validate.py
+++ b/connaisseur/validate.py
@@ -85,13 +85,12 @@ def process_chain_of_trust(
 
     tuf_roles = ["root", "snapshot", "timestamp", "targets"]
 
-    # Load all trust data
+    # load all trust data
     for role in tuf_roles:
-        role_trust_data = get_trust_data(host, image, TUFRole(role))
-        trust_data[role] = role_trust_data
+        trust_data[role] = get_trust_data(host, image, TUFRole(role))
 
-    # Validate signature and expiry data of and load root file
-    # This does NOT conclude the validation of the root file. To prevent roleback/freeze attacks,
+    # validate signature and expiry data of and load root file
+    # this does NOT conclude the validation of the root file. To prevent roleback/freeze attacks,
     # the hash still needs to be validated against the snapshot file
     root_trust_data = get_trust_data(host, image, TUFRole("root"))
     root_trust_data.validate_signature(key_store)
@@ -99,14 +98,14 @@ def process_chain_of_trust(
     trust_data["root"] = root_trust_data
     key_store.update(root_trust_data)
 
-    # Validate timestamp file to prevent freeze attacks
+    # validate timestamp file to prevent freeze attacks
     # validates signature and expiry data
     # there is no hash to verify it against since it is short lived
     # TODO should we ensure short expiry duration here?
     timestamp_trust_data = trust_data["timestamp"]
     timestamp_trust_data.validate(key_store)
 
-    # Validate snapshot file signature against the key defined in the root file
+    # validate snapshot file signature against the key defined in the root file
     # and its hash against the one from the timestamp file
     # and validate expiry
     snapshot_trust_data = trust_data["snapshot"]
@@ -118,14 +117,14 @@ def process_chain_of_trust(
 
     snapshot_trust_data.validate_expiry()
 
-    # Now snapshot and timestamp files are validated, we can be safe against
+    # now snapshot and timestamp files are validated, we can be safe against
     # roleback and freeze attacks if the root file matches the hash of the snapshot file
     # (or the root key has been compromised, which Connaisseur cannot defend against)
     snapshot_key_store = KeyStore()
     snapshot_key_store.update(snapshot_trust_data)
     root_trust_data.validate_hash(snapshot_key_store)
 
-    # If we are safe at this point, we can add the snapshot data to the main KeyStore
+    # if we are safe at this point, we can add the snapshot data to the main KeyStore
     # and proceed with validating the targets file and (potentially) delegation files
     key_store.update(snapshot_trust_data)
     targets_trust_data = trust_data["targets"]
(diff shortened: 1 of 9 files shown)


connaisseur/validate.py
```diff
@@ -136,11 +137,11 @@ def process_chain_of_trust(
     # repository, but no signatures were created using the delegations. in this special
     # case, the releases JSON doesn't exist yet and the targets JSON must be used instead
     if req_delegations:
-        if not all(trust_data[target_role] for target_role in req_delegations):
+        if not all(trust_data.get(target_role) for target_role in req_delegations):
             tuf_roles = [
                 target_role
                 for target_role in req_delegations
-                if not trust_data[target_role]
+                if not trust_data.get(target_role)
```
peterthomassen, line 144: Can it happen that `target_role` is not in `trust_data`? If so, `get()` returns `None`, and the `if` conditions all evaluate to true. Is that intended?

If it can't happen, it will be safer to use `[]` subscripts, and if it really doesn't happen in practice, it will also never throw a `KeyError`. :)

Starkteetje: The way I re-wrote it calls `trust_data[delegation] = delegation_trust_data` only `if delegation_trust_data is not None:` (as part of setting `trust_data` entries only after validation [and if delegation data is None, we dont validate]), thus it can actually not exist and then correctly `get`s None
But maybe that's somewhat unexpected and I should set `None` inside the `trust_data` and stick to `[]`? Works in this context since validation is only skipped for `None` and that's a "safe" value in this context. Not sure :man_shrugging:

peterthomassen: ah, I see now (did not all of that in my working memory at the same time :-) ). I slightly lean towards unindenting `trust_data[delegation] = delegation_trust_data` by one level and sticking to `[]`, but it's probably not worth the effort.

connaisseur/validate.py
```diff
@@ -225,3 +201,38 @@ def search_image_targets_for_tag(trust_data: dict, image: Image):
 
     base64_digest = trust_data[image_tag]["hashes"]["sha256"]
     return base64.b64decode(base64_digest).hex()
+
+
+def update_with_delegation_trust_data(trust_data, delegations, key_store, host, image):
```
peterthomassen, line 206: If this function is only used in this file, you may want to make it private (prepend with `_`)

connaisseur/validate.py
```diff
@@ -225,3 +201,38 @@ def search_image_targets_for_tag(trust_data: dict, image: Image):
...
+            host, image, TUFRole(delegation)
+        )
+        # when delegations are added to the repository, but weren't yet used for signing, the
+        # delegation files don't exist yet and are `None`. in this case validation must be skipped
+        if delegation_trust_data is not None:
+            delegation_trust_data.validate(key_store)
+        trust_data[delegation] = delegation_trust_data
+
+
+def validate_all_required_delegations_present(
```
peterthomassen, line 218: If this function is only used in this file, you may want to make it private (prepend with `_`)

connaisseur/validate.py
```diff
@@ -72,41 +72,80 @@ def process_chain_of_trust(
...
-    # get all trust data and collect keys (from root and targets), as well as
-    # hashes (from snapshot and timestamp)
+    tuf_roles = ["root", "snapshot", "timestamp", "targets"]
+
+    # Load all trust data
     for role in tuf_roles:
-        trust_data[role] = get_trust_data(host, image, TUFRole(role))
-        key_store.update(trust_data[role])
+        role_trust_data = get_trust_data(host, image, TUFRole(role))
+        trust_data[role] = role_trust_data
```
peterthomassen, line 91: unnecessary temporary variable?

connaisseur/validate.py
```diff
@@ -72,41 +72,80 @@ def process_chain_of_trust(
...
+
+    # Validate signature and expiry data of and load root file
+    # This does NOT conclude the validation of the root file. To prevent roleback/freeze attacks,
+    # the hash still needs to be validated against the snapshot file
```
peterthomassen: I am not following why the validation now has been expanded. That was not the case in the previous iteration of the commit, and it also does not seem to be the objective of the commit (which only is to change order). Am I missing something obvious?

peterthomassen (changes requested):

connaisseur/validate.py
Starkteetje: It hasn't really been expanded but the order has changed. Previously `validate` was called, which in turn called `validate_signature`, `validate_expiry` and `validate_hash`. But it did so by loading and "trusting" all DCT files and then using this data to validate said files. This commit now does validation in an order such that there's never (potentially) untrusted data is used to validate other files. But that necessitates that the order be changed, since e.g. the hash of th root.json can only be compared to snapshot state once the signature on snapshot has been validated, yet that requires first validating previous files (as good as it gets).

Starkteetje: However, yes it changes the original intent of the commit. It fixes an additional vector of attack where the security was based on targeted collision resistance instead of 2nd preimage resistance of the DCT hash function (still theoretical only)

peterthomassen (dismissed):
lgtm! :)

phbelitz (approved):

PR #90, opened 2021-02-03
Feat/multi notary

phbelitz:
multi notary support


connaisseur/flask_server.py
```diff
@@ -5,7 +5,8 @@
 from requests.exceptions import HTTPError
 from connaisseur.exceptions import BaseConnaisseurException, UnknownVersionError
 from connaisseur.mutate import admit, validate
-from connaisseur.notary_api import health_check
+
+# from connaisseur.notary_api import health_check
```
Starkteetje, line 9: Remove unused import :)

connaisseur/flask_server.py
```diff
@@ -111,10 +112,4 @@ def readyz():
...
         webhook_response = None
 
-    notary_health = health_check(os.environ.get("NOTARY_SERVER"))
-
-    return (
-        ("", 200)
-        if ((webhook_response or sentinel_running) and notary_health)
-        else ("", 500)
-    )
+    return ("", 200) if (webhook_response or sentinel_running) else ("", 500)
```
Starkteetje, line 115: Since this changes behaviour from ADR-1, we might want to add an update as ADR-3. Would probably be a good idea anyways, covering the config layout we discussed at length

Starkteetje (changes requested):
Not yet done with review, but guess I'll comment already in case you need stuff to do :P

Makefile
```diff
@@ -28,11 +28,10 @@ install: certs
 uninstall:
 	kubectl config set-context --current --namespace $(NAMESPACE)
 	helm uninstall connaisseur
-	kubectl delete ns $(NAMESPACE) --timeout=120s
```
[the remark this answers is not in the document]
phbelitz: I found it to be impratical when using predefined secrets, like image pull secrets or general authentication secrets. They would be deleted as well, which might not be the wished for behavior.

peterthomassen:
I'll let @Starkteetje finish his review and wait for the subsequent changes. Happy to do a review then -- just hit the "request review" button :)

Makefile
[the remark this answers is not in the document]
Starkteetje: Good point. Would add that part to https://github.com/sse-secure-systems/connaisseur/blob/master/setup/README.md#cleanup then. "We do not delete the namespace, since you might have deployed things in there, which we would inadvertently delete. If you want to get rid of the namespace, run `kubectl delete ns WHATEVEROURDEFAULTIS`"

helm/values.yaml
```diff
@@ -1,7 +1,7 @@
 # configure connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v1.4.5
+  image: securesystemsengineering/connaisseur:v1.5.0
```
Starkteetje, line 4: From the single entry to list change in the config, I would assume this breaks existing configs and would assume 2.0.0 or does `self.notaries = [Notary(notary) for notary in config_content]` parse both a single notary and a notary that is a list?

connaisseur/config.py
```diff
@@ -0,0 +1,155 @@
...
+                )
+            return next(iter(self.notaries))
+        except StopIteration:
+            raise NotFoundException(
+                "the given notary configuration could not be found.",
+                {"notary_name": notary_name},
+            )
+
+
+class Notary:  # pylint: disable=too-many-instance-attributes
```
Starkteetje, line 65: Would this class be better suited for a separate file?

setup/README.md
```diff
@@ -56,81 +56,76 @@ Before you can finally set up Connaisseur, you will need to configure its connec
...
+In `helm/values.yaml`, we first need to specify the location of the to be used notary instance. Connaisseur supports the configuration of multiple notary instances, should your image signatures be at different locations. Per default the official DockerHub notary is configured, with the public key for all official images, but for now let's assume you want to add a new notary instance. Simply add a new entry in the `notary` field, specifying a `name`, `host` and a list of public root keys (`rootKeys`). The `name` should be unique among all notary configurations, the `host` points to the host address of the notary instance and the list of public root keys will be used to verify the image signatures residing in this notary instance. Why a list, you may ask? Just take a look at the the official DockerHub notary, where the official images are signed with a different key as your own images you can push to DockerHub. If you want to verify both, official images and your own (all residing on DockerHub), you'll need multiple keys. How to obtain these will be described further below ([here](#retrieving-the-public-root-key)).
```
Starkteetje, line 59:
```suggestion
In `helm/values.yaml`, we first need to specify the location of the notary instance to be used. Connaisseur supports the configuration of multiple notary instances, should your image signatures reside at different locations. Per default, the official DockerHub notary is configured, with the public key for all official images, but for now let's assume you want to add a new notary instance. Simply add a new entry in the `notary` field, specifying a `name`, `host` and a list of public root keys (`rootKeys`). The `name` should be unique among all notary configurations, the `host` points to the host address of the notary instance and the list of public root keys will be used to verify the image signatures residing in this notary instance. Why a list, you may ask? Just take a look at the the official DockerHub notary, where the official images are signed with a different key as your own images you can push to DockerHub. If you want to verify both, official images and your own (all residing on DockerHub), you'll need multiple keys. How to obtain these will be described further below ([here](#retrieving-the-public-root-key)).
```

setup/README.md
```diff
@@ -56,81 +56,76 @@ Before you can finally set up Connaisseur, you will need to configure its connec
...
+In addition to these three mandatory settings, a self-signed certificate and authentication can be added, as well as marking a notary belonging to a azure container registry (acr). The self-signed certificate can be added with the `selfsignedCert` field, where it will be pasted in a PEM format. The authentication detail can be defined in the `auth` field, where either the credentials are given directly with `auth.user` and `auth.password` or a predefined [Kubernetes secret](https://kubernetes.io/docs/concepts/configuration/secret/) is referenced in the `auth.secretName` field. The data of the predefined secret needs to be a `cred.yaml` with the fields `USER` and `PASS`. Lastly the `isAcr` can be used to mark the notary as being part of an azure container registry, as these types of notaries present a slightly different kind of behavior. All in all a sample notary configuration might look like this:
```
Starkteetje, line 61:
```suggestion
In addition to these three mandatory settings, a self-signed certificate and authentication can be added, as well as marking a notary belonging to an Azure Container Registry (ACR). The self-signed certificate can be added with the `selfsignedCert` field, where it will be pasted in PEM format. The authentication detail can be defined in the `auth` field, where either the credentials are given directly with `auth.user` and `auth.password` or a predefined [Kubernetes secret](https://kubernetes.io/docs/concepts/configuration/secret/) is referenced in the `auth.secretName` field. The data of the predefined secret needs to be a `cred.yaml` with the fields `USER` and `PASS`. Lastly the `isAcr` can be used to mark the notary as being part of an Azure Container Registry, as these types of notaries present a slightly different kind of behavior. All in all a sample notary configuration might look like this:
```

setup/README.md
```diff
@@ -56,81 +56,76 @@ Before you can finally set up Connaisseur, you will need to configure its connec
...
-    -----BEGIN PUBLIC KEY-----
-    MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAETBDLAICCabJQXB01DOy315nDm0aD
-    BREZ4aWG+uphuFrZWw0uAVLW9B/AIcJkHa7xQ/NLtrDi3Ou5dENzDy+Lkg==
-    -----END PUBLIC KEY-----
-```
-
-2. In case you already had DCT setup or cannot find the file, we need to retrieve and convert the key manually from your previously created root key. For most shell users, there is a short script below. Otherwise, just follow the manual steps:
-
-##### manual
+For now let's assume you only want to verify image signatures you created/will create, using the key generated in step [two](#set-up-docker-content-trust). To get the public part of this you have to convert the key manually.
```
Starkteetje, line 96:
```suggestion
For now let's assume you only want to verify image signatures you created/will create, using the key generated in [step two](#set-up-docker-content-trust). To get the public part of this you have to convert the key manually.
```

setup/README.md
```diff
@@ -56,81 +56,76 @@ Before you can finally set up Connaisseur, you will need to configure its connec
...
 
-- After entering your password, copy the public key to the `helm/values.yaml`: 
+#### Configure the image policy
 
-```bash
-yq e '.notary.rootPubKey=$(cat root.pub)' -i "${OLDPWD}/helm/values.yaml"
-rm root-priv.key root.pub
-cd -
-```
+Since we have potentially multiple notary instances and multiple keys for each instance, Connaisseur somehow needs to know for which images it should use what kind of combination of instance and key. This is configured in the `policy` part of the `helm/values.yaml`, where different rules are defined, that need to be matched against a given image. Only one rule will be matched at the same time and this rule can specify which notary should be used, with the `notary` field. The value of this field has to match one of the notary `name` fields. Additionally a `key` field can be given in the rule, which will then specify which key within the chosen notary configuration should be used. This again has to match a value of a `name` field of one of the `rootKeys` entries. If no `notary` or `key` fields are given inside a rule, then the top most entry will be chosen respectively.
```
Starkteetje, line 114:
```suggestion
Since we have potentially multiple notary instances and multiple keys for each instance, Connaisseur somehow needs to know for which images it should use what kind of combination of instance and key (otherwise, any key compromised in your supply chain would constitute a (potential) compromise of all your images). This is configured in the `policy` part of the `helm/values.yaml`, where different rules are defined, that need to be matched against a given image. Only one rule will be matched at the same time and this rule can specify which notary should be used, with the `notary` field. The value of this field has to match one of the notary `name` fields. Additionally a `key` field can be given in the rule, which will then specify which key within the chosen notary configuration should be used. This again has to match a value of a `name` field of one of the `rootKeys` entries. If no `notary` or `key` fields are given inside a rule, then the top most entry will be chosen respectively.
```

Starkteetje, line 114: If we choose differently in the ADR, the last sentence will need to be changed

setup/README.md
```diff
@@ -56,81 +56,76 @@ Before you can finally set up Connaisseur, you will need to configure its connec
...
+
+In the below example, all images coming from a `core.harbor.domain` registry will be verified using the `harbor` notary from the example above and use it's `app1` key. All other images take the predefined `dockerhub` notary and it's `library` key.
```
Starkteetje, line 116:
```suggestion
In the below example, all images coming from a `core.harbor.domain` registry will be verified using the `harbor` notary from the example above and use its `app1` key. All other images take the predefined `dockerhub` notary and its `library` key.
```

setup/local/README.md
```diff
@@ -264,34 +264,53 @@ git clone https://github.com/sse-secure-systems/connaisseur.git
 cd connaisseur
 ```
 
-In there, you will find a `helm` directory, which holds a `values.yaml` file, which needs to be configured. Change the `notary` section to reflect the appropriate values. In `notary.selfsignedCert` you have to put the certificate from `harbor-ca.crt`.
+In there, you will find a `helm` directory, which holds a `values.yaml` file, which needs to be configured. Add a new notary configuration with all appropriate values. The `selfsignedCert` filed should be filled with the contents of the `harbor-ca.crt` from earlier.
```
Starkteetje, line 267:
```suggestion
In there, you will find a `helm` directory, which holds a `values.yaml` file, which needs to be configured. Add a new notary configuration with all appropriate values. The `selfsignedCert` field should be filled with the contents of the `harbor-ca.crt` from earlier.
```

Starkteetje (changes requested):

Makefile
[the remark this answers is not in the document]
phbelitz: resolved in 3b7d64cd6e6c5c09c2f7370d57ddb8a3b6c8869c

setup/local/README.md
phbelitz: resolved in 3b7d64cd6e6c5c09c2f7370d57ddb8a3b6c8869c

setup/README.md
phbelitz: resolved in 3b7d64cd6e6c5c09c2f7370d57ddb8a3b6c8869c

setup/README.md
phbelitz: resolved in 3b7d64cd6e6c5c09c2f7370d57ddb8a3b6c8869c

setup/README.md
phbelitz: resolved in 3b7d64cd6e6c5c09c2f7370d57ddb8a3b6c8869c

adr/ADR-3_multi_notary_config.md
```diff
@@ -0,0 +1,103 @@
...
+
+## Considered options
+
+### Choice 1
+
+The overall notary configuration setup in `helm/values.yaml`.
+
+#### Option 1 (Per Notary)
+
+The `notary` field becomes a list. Per to be used physical notary instance, there will be one entry in this list.
```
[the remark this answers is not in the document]
phbelitz: resolved in 1c4d113cc7ef7420ff8fa8cb574619cabc49c86f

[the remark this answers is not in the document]
phbelitz: resolved in 212f8aacbd3c58392374aa2dbb6eaef5d6dd4bcf

adr/ADR-3_multi_notary_config.md
```diff
@@ -0,0 +1,103 @@
...
+
+The entry will have the following data fields (**bold** are mandatory):
+
+- **`name`** -- An unique identifier for the notary configuration, which will be used in the image policy.
```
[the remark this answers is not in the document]
phbelitz: 212f8aacbd3c58392374aa2dbb6eaef5d6dd4bcf

adr/ADR-3_multi_notary_config.md
```diff
@@ -0,0 +1,103 @@
...
+- **`host`** -- The host address of the notary instance, where to reach it.
```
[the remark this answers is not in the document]
phbelitz: 212f8aacbd3c58392374aa2dbb6eaef5d6dd4bcf

adr/ADR-3_multi_notary_config.md
```diff
@@ -0,0 +1,103 @@
...
+    - `password` -- Password to authenticate with.
+    - `secretName` -- Kubernetes secret reference to use *INSTEAD* of user/password combination.
+- `isAcr` -- Marks the notary as being part of an Azure Container Registry.
+
+The image policy will have two additional fields per rule entry (in "quotes" are already present fields):
+
+- "`pattern`" -- Image pattern to match against, for rule to apply.
+- "`verify`" -- Whether the images should be verified or not.
+- "`delegations`" -- List of required delegation roles.
+- `notary` -- Which notary to use, for any matching image. This has to correspond to one of notary configuration `name` fields. If none is given, the top most notary configuration is taken.
```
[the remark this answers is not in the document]
phbelitz: 212f8aacbd3c58392374aa2dbb6eaef5d6dd4bcf

connaisseur/key_store.py
```diff
@@ -10,23 +10,11 @@ class KeyStore:
...
     hashes: dict
 
-    def __init__(self):
+    def __init__(self, root_pub_key: str):
         # will always be loaded there as k8s secret
-        root_path = "/etc/certs/root-pub.pem"
-        root_key = KeyStore.load_root_pub_key(root_path)
-
-        self.keys = {"root": root_key}
+        self.keys = {"root": root_pub_key}
```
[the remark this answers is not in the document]
phbelitz: Yea you're right. We should do some validation here, and everytime a key is loaded into the key_store. I'll create an issue.

connaisseur/config.py
```diff
@@ -0,0 +1,155 @@
...
+        self.root_keys = notary_config.get("rootKeys")
+        self.is_acr = notary_config.get("isAcr", False)
+
+        if not (self.name and self.host and self.root_keys):
+            raise InvalidFormatException(
+                "error parsing the the connaisseur configuration file."
+            )
+
+        self.name = self.name.replace("..", "")  # no path traversal
+        self.is_selfsigned = os.path.exists(self.SELFSIGNED_PATH.format(self.name))
```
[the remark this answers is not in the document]
phbelitz: https://github.com/sse-secure-systems/connaisseur/blob/9687028839e5fd68ad9dbc352365661a80aa93bf/connaisseur/config.py#L94-L97

This makes sure `self.name` can't be `None`, thus we won't be looking for `/etc/certs/None.crt`.

helm/values.yaml
```diff
@@ -22,46 +22,49 @@ service:
...
+#     -----BEGIN CERTIFICATE-----
+#     ...
+#     -----END CERTIFICATE-----
+#   auth:  # authentication details
+      # either
+#     user: notaryuser  # username
+#     password: notarypassword  # password
+      # or
+#     secretName: notary-secret  # external kubernetes secret
+#   isAcr: true  # mark notary as part of an ACR
```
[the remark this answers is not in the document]
phbelitz: https://github.com/sse-secure-systems/connaisseur/blob/9687028839e5fd68ad9dbc352365661a80aa93bf/connaisseur/notary_api.py#L157-L160

Welp, in a ACR the access token is returned differently, but I guess we could look for both `"token"` and `"access_token"` and take, which ever is not empty .... what do you think?

helm/values.yaml
phbelitz: Nope, especially if we change the `notary` field into `notaries`. I'll go with version 2.0.0

adr/ADR-3_multi_notary_config.md
```diff
@@ -0,0 +1,103 @@
...
+
+When no `notary` is specified in a image policy rule, the first entry in the `notary` configuration list is taken. The same goes for the public root key list, should option 1 for choice 1 be chosen.
+
+Problem: Might get inconsistent, should list ordering in python get shuffled around
+
+#### Option 2 (Explicit default)
+
+One of the notary configuration will be given a `default` field, which marks it as the default value.
+
+Problem: Hard to validate that at least on notary configuration has a `default` field ...?
```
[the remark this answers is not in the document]
phbelitz: Aight, thanks to stackoverflow i found a jsonschema to acurately validate this. So shouldn't be hard anymore.

connaisseur/config.py
phbelitz: Yea unsure about that, i found the `Notary` class to be logically close enough to the `Config` class, that I put both in one file. If you + maybe peter don't think this is rather important, I would leave it as is.

peterthomassen: I have no stakes in this.

connaisseur/util.py
```diff
@@ -1,5 +1,15 @@
...
     prefix = "targets/"
     if not delegation_role.startswith(prefix):
         delegation_role = prefix + delegation_role
     return delegation_role
+
+
+def safe_path_exists(base_dir: str, path: str):
+    if os.path.commonprefix((os.path.realpath(path), base_dir)) != base_dir:
+        raise InvalidFormatException("potential path traversal.", {"path": path})
+    return os.path.exists(path)
```
peterthomassen, line 15: nice!

Just a random thought, a generalization of this would be:

```python
def safe_path_func(callback, base_dir: str, path: str, *args, **kwargs):
    if os.path.commonprefix((os.path.realpath(path), base_dir)) != base_dir:
        raise InvalidFormatException("potential path traversal.", {"path": path})
    return callback(path, *args, **kwargs)
```

... to be called with `safe_path_func(os.path.exists, ...)`. You could also replace `with open(...): by `with safe_path_func(open, ...):` etc.

Note that this is very much like a function decorator (or context manager), so there may be more elegant ways.

---

I also just realized that `InvalidFormatException(BaseConnaisseurException)` takes a `context` argument. It would be more pythonic to declare `BaseConnaisseurException.__init__(self, message: str, **kwargs)`, then set `self.context = kwargs`.

As a result, the exception here could be raised as `raise InvalidFormatException(..., path=path)`.

connaisseur/config.py
```diff
@@ -0,0 +1,110 @@
...
+    def get_key(self, key_name: str = None):
+        try:
+            if key_name:
+                key = next(
+                    key.get("key")
+                    for key in self.root_keys
+                    if key.get("name") == key_name
+                )
+            else:
+                key = next(iter(self.root_keys)).get("key")
```
peterthomassen, line 123: If I understand correctly, `key` and `name` are expected to be present. So let's better make that explicit by using subscripts `[]`.

connaisseur/config.py
```diff
@@ -0,0 +1,110 @@
...
+            return "".join(key.split("\n")[1:-2])
```
peterthomassen, line 124: What do the numbers in `[1:-2]` depend on (especially the `2`)? (Should we be trimming whitespace, for example?)

connaisseur/config.py
```diff
@@ -0,0 +1,110 @@
...
+            with open(self.AUTH_PATH.format(self.name), "r") as cred_file:
+                self.auth = yaml.safe_load(cred_file)
+
+                if not (self.auth.get("USER") and self.auth.get("PASS")):
+                    raise InvalidFormatException(
+                        (
+                            "credentials for host configuration "
+                            "{} are in a wrong format."
+                        ).format(self.name)
+                    )
```
peterthomassen, line 148: You are loading arbitrary stuff into a local variable. Should we be more explicit, e.g.

```python
auth = yaml.safe_load(cred_file)
try:
    return auth["USER"], auth["PASS"]
except KeyError:
    msg = f"credentials for host configuration {self.name} are in a wrong format."
    raise InvalidFormatException(msg)
```

Note that I also changed the return value to a tuple. This has the advantage that down below, when you use this to construct the basic auth header, you can simply use tuple unpacking (`*notary_config.auth`).

connaisseur/config.py
```diff
@@ -0,0 +1,110 @@
...
+        return self.auth
+
+    def get_selfsigned_cert(self):
+        if self.is_selfsigned and not self.selfsigned_cert:
+            self.selfsigned_cert = self.SELFSIGNED_PATH.format(self.name)
```
peterthomassen, line 157: path traversal if `self.name` contains `..`? -- This is an instance of a TOC/TOU race condition: The common path prefix check has been done during initialization, and now we're assuming that `self.name` is still the same. Of course, that is currently the case, but who knows whether we'll remember in the future to not modify `self.name`?

The prefix check is done during the computation of the `has_auth` and `is_selfsigned` flags. If we remove those (during restructuring according to "beg for forgiveness", so we don't need the flags anymore), the opportunity for the check during initialization will also not be as obvious. I suggest running the check closer to where the path is used (e.g. here).

connaisseur/config.py
```diff
@@ -0,0 +1,110 @@
...
+            else:
+                key = next(iter(self.root_keys)).get("key")
+            return "".join(key.split("\n")[1:-2])
+        except StopIteration:
+            raise NotFoundException(
+                "the give public key could not be found.", {"key_name": key_name}
+            )
+
+    def get_auth(self):
+        if not self.auth and self.has_auth:
```
peterthomassen, line 138: No need to check here (we can get rid of the whole `has_auth` attribute). Just give it a try and see if you can open the file. If you can, return the auth information. Otherwise, catch the exception from `open` and `return None`.

This is the Python principle of "begging for forgiveness" instead of "asking for permission". Take a look here: https://docs.quantifiedcode.com/python-anti-patterns/readability/asking_for_permission_instead_of_forgiveness_when_working_with_files.html It is also closely related to the idea of [duck typing](https://en.wikipedia.org/wiki/Duck_typing). It makes code more readable and less error-prone. For example, you can't run into [TOC/TOU race conditions](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use).

connaisseur/notary_api.py
```diff
@@ -160,21 +125,23 @@ def parse_auth(auth_header: str):
...
-    if user:
-        request_kwargs["auth"] = requests.auth.HTTPBasicAuth(user, password)
 
-    if is_notary_selfsigned():
-        request_kwargs["verify"] = "/etc/certs/notary.crt"
+    if notary_config.has_auth:
+        auth = notary_config.get_auth()
+        request_kwargs["auth"] = requests.auth.HTTPBasicAuth(
+            auth.get("USER"), auth.get("PASS")
+        )
```
peterthomassen, line 141: If you get rid of `has_auth` as suggested above, you may simply rewrite this as:

```python
try:
    request_kwargs["auth"] = requests.auth.HTTPBasicAuth(*notary_config.auth)
except TypeError:  # happens if .auth is None
    pass  # just ignore the failure
```

This is another instance of begging for forgiveness instead of asking for permission.

connaisseur/config.py
```diff
@@ -0,0 +1,110 @@
...
+                )
+            else:
+                key = next(iter(self.root_keys)).get("key")
+            return "".join(key.split("\n")[1:-2])
+        except StopIteration:
+            raise NotFoundException(
+                "the give public key could not be found.", {"key_name": key_name}
+            )
+
+    def get_auth(self):
```
peterthomassen, line 130: As this takes no arguments, you can make this a property:

```python
@property
def auth(self):
    ...
```

As a result, you can access this as `.auth` (without `()`).

connaisseur/notary_api.py
```diff
@@ -160,21 +125,23 @@ def parse_auth(auth_header: str):
...
+
+    if notary_config.is_selfsigned:
+        request_kwargs["verify"] = notary_config.get_selfsigned_cert()
```
peterthomassen, line 144: This can be similarly restructured.

requirements.txt
```diff
@@ -6,3 +6,4 @@ jsonschema~=3.2.0
 parsedatetime~=2.6
 pytz~=2020.1
 python-dateutil~=2.8.1
+PyYAML~=5.3.1
```
peterthomassen, line 9: 5.4.1 is current. Also, consider bumping other dependencies every once in a while.

peterthomassen (changes requested):

Makefile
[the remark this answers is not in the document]
phbelitz: Nothing in particular, just disrupted the workflow. I can revert it back and apply the changes in a alter refactor commit.

connaisseur/util.py
phbelitz: The exception stuff would've to be done for all exceptions, so I think it's best to do it in a seperate refactoring PR including the deletion of the namespeace from above.

connaisseur/config.py
phbelitz: this is a yaml snippet that gets transformed into json:

```yaml
root_keys:
  - name: library
    key: |
      -----BEGIN PUBLIC KEY-----
      MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEOXYta5TgdCwXTCnLU09W5T4M4r9f
      QQrqJuADP6U7g5r9ICgPSmZuRHP/1AYUfOQW3baveKsT969EfELKj1lfCA==
      -----END PUBLIC KEY-----
```

here the json part of this:

```json
{
  "root_keys": [
    {
      "name": "library",
      "key": "-----BEGIN PUBLIC KEY-----\nMFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEOXYta5TgdCwXTCnLU09W5T4M4r9f\nQQrqJuADP6U7g5r9ICgPSmZuRHP/1AYUfOQW3baveKsT969EfELKj1lfCA==\n-----END PUBLIC KEY-----\n"
    }
  ]
}
```

the `[1:-2]` will get us the key in DER format.

peterthomassen: I guess that it's not a very safe to assume that this will always work. For example, what happens if there's an extra blank line in the beginning of `key:`?

phbelitz: what do you recommend? regex for "-----BEGIN CERTIFICATE-----", "-----END CERTIFICATE-----" and whitespace; then remove those?

peterthomassen: We can use a suitable parsing function that takes care of everything, including whitespace formatting and base64 decoding. That means we have to use a function that decodes PEM (the format including the boundaries) and not DER (a binary representation of the actual key).

```python
return cryptography.hazmat.primitives.serialization.load_pem_public_key(key)
```

However, as there are other places where that applies, too, it may be cleaner to just `return key` here and let the `KeyStore` handle it. Specifically, we can add the call to `load_pem_public_key` in `KeyStore.__init__` and `KeyStore.update`.

As a result, the key will be an object, not a string, so the first argument of `crypto.verify_signature` should become just `pub_key`.

connaisseur/config.py
```diff
@@ -40,7 +41,10 @@ def __init__(self):
                 "invalid format for Connaisseur configuration."
             )
 
-        self.notaries = [Notary(notary) for notary in config_content]
+        self.notaries = [
+            Notary(**{k: v for k, v in notary.items() if v is not None})
+            for notary in config_content
+        ]
```
peterthomassen, line 47: First, why are `None` values a problem?
Second, why take care of them here? I think `Notary.__init__` should handle that.

We can discuss off-band if you like.

connaisseur/config.py
```diff
@@ -115,44 +103,47 @@ def get_key(self, key_name: str = None):
...
+    def selfsigned_cert(self):
         """
         Returns the path to a selfsigned certificate, should it exist. Otherwise None is
         returned.
         """
-        if self.is_selfsigned and not self.selfsigned_cert:
-            self.selfsigned_cert = self.SELFSIGNED_PATH.format(self.name)
-        return self.selfsigned_cert
+        path = self.SELFSIGNED_PATH.format(self.name)
+        return path if safe_path_func(os.path.exists, "/etc/certs/", path) else None
```
peterthomassen, line 149: nice!

peterthomassen (dismissed):

adr/ADR-3_multi_notary_config.md
```diff
@@ -0,0 +1,103 @@
...
+
+The entries will look identical to the one's from option 1, with two exceptions.
+
+1. The `root_keys` field of the notary configurations won't be a list and only has a single entry, without needing to specify a key name.
+
+2. The image policy will only address the notary configuration to be chosen with the `notary` field, without the need for a `key` field.
+
+### Choice 2
+
+Default values for `notary` (and `key`) inside the image policy.
```
xopham, line 59: I think there might be a 4th option for setting a default:
The image policy was naturally built to allow for setting an overall default via:
```
  - pattern: "*:*"
    verify: true
```
If that specifies the default configuration to use via `notary: default` if not otherwise configured in a more specific rule. If then the preset notary config goes by the name `default` it might be sufficiently explicit:
```
- pattern: "*:*"
  verify: true
  notary: default
- pattern: "k8s.gcr.io/*:*"
  verify: false
  verify: false
- pattern: "docker.io/securesystemsengineering/testimage:*"
  verify: true
```
here the last pattern should then just test against the next less specific rule that specifies a notary...

setup/README.md
```diff
@@ -248,7 +244,7 @@ Your signed images were allowed through, in contrast to those unsigned ones with
 
 ## Cleanup
 
-If you want to remove Connaisseur, run `make uninstall`.
+If you want to remove Connaisseur, run `make uninstall`. We do not delete the namespace, since you might have deployed things in there, which we would inadvertently delete. If you want to get rid of the namespace, run `kubectl delete ns WHATEVEROURDEFAULTIS`.
```
Starkteetje, line 247:
```suggestion
If you want to remove Connaisseur, run `make uninstall`. We do not delete the namespace, since you might have deployed things in there, which we would inadvertently delete. If you want to get rid of the namespace, run `kubectl delete ns connaisseur` (or whatever other namespace you installed Connaisseur to).
```

connaisseur/config.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+
+    path = "/etc/connaisseur-config/config.yaml"
+    schema_path = "/app/connaisseur/res/config_schema.json"
+    notaries: list = []
+
+    def __init__(self):
+        """
+        Creates a Config object, containing all notary configurations. It does so by
+        reading a config file, doing input validation and then creating Notary object,
```
Starkteetje, line 21:
```suggestion
        reading a config file, doing input validation and then creating Notary objects,
```

connaisseur/config.py
```diff
@@ -0,0 +1,152 @@
...
+            raise InvalidFormatException(
+                "invalid format for Connaisseur configuration."
+            )
+
+        self.notaries = [Notary(**notary) for notary in config_content]
+
+    def get_notary(self, notary_name: str = None):
+        """
+        Returns the notary configuration with the given `notary_name`. If `notary_name`
+        is None, the top most element of the notary configuration list is returned.
```
Starkteetje, line 49: This comment is no longer accurate with regards to None, right?

connaisseur/config.py
```diff
@@ -0,0 +1,152 @@
...
+        self.name = name
+        self.host = host
+        self.root_keys = root_keys
+        self.is_acr = kwargs.get("is_acr", False)
+
+    def get_key(self, key_name: str = None):
+        """
+        Returns the public root key with name `key_name` in DER format, without any
+        whitespaces. If `key_name` is None, the top most element of the public root key
+        list is returned.
```
Starkteetje, line 100: This comment is no longer accurate with regards to None, right?

connaisseur/config.py
```diff
@@ -0,0 +1,152 @@
...
+            )
+
+        self.notaries = [Notary(**notary) for notary in config_content]
+
+    def get_notary(self, notary_name: str = None):
+        """
+        Returns the notary configuration with the given `notary_name`. If `notary_name`
+        is None, the top most element of the notary configuration list is returned.
+
+        Raises `NotFoundException` if no top most element can be found.
```
Starkteetje, line 51:
```suggestion
        Raises `NotFoundException` if no matching or default element can be found.
```

connaisseur/config.py
```diff
@@ -0,0 +1,152 @@
...
+    is_acr: bool = False
+
+    SELFSIGNED_PATH = "/etc/certs/{}.crt"
+    AUTH_PATH = "/etc/creds/{}/cred.yaml"
+
+    def __init__(self, name: str, host: str, root_keys: list, **kwargs):
+        """
+        Creates a Notary object from a dictionary.
+
+        Raises `InvalidFormatException` should teh mandatory fields be missing.
```
Starkteetje, line 85:
```suggestion
        Raises `InvalidFormatException` should the mandatory fields be missing.
```

connaisseur/config.py
```diff
@@ -0,0 +1,152 @@
...
+        self.root_keys = root_keys
+        self.is_acr = kwargs.get("is_acr", False)
+
+    def get_key(self, key_name: str = None):
+        """
+        Returns the public root key with name `key_name` in DER format, without any
+        whitespaces. If `key_name` is None, the top most element of the public root key
+        list is returned.
+
+        Raises `NotFoundException` if no top most element can be found.
```
Starkteetje, line 102:
```suggestion
        Raises `NotFoundException` if no matching or default element can be found.
```

connaisseur/config.py
```diff
@@ -0,0 +1,152 @@
...
+                key = next(iter(self.root_keys))["key"]
+            else:
+                key_name = key_name or "default"
+                key = next(
+                    key["key"] for key in self.root_keys if key["name"] == key_name
+                )
+            return "".join(key)
+        except StopIteration:
+            raise NotFoundException(
+                "the give public key could not be found.", {"key_name": key_name}
```
Starkteetje, line 116:
```suggestion
                "the given public key could not be found.", {"key_name": key_name}
```

connaisseur/key_store.py
```diff
@@ -60,9 +56,27 @@ def update(self, trust_data):
         """
 
         # update keys
-        keys = trust_data.get_keys()
-        for key in keys:
-            self.keys.setdefault(key, keys[key]["keyval"]["public"])
+        keys = dict(trust_data.get_keys())
+        signature_keys = [sig.get("keyid") for sig in trust_data.signatures]
```
Starkteetje, line 60: Would move this line below the comment as it only pertains to the very case described in the comment

Starkteetje (changes requested):

README.md
```diff
@@ -172,7 +172,7 @@ The policy consists of a set of rules. Each rule starts with a pattern that need
...
 
 1. Should there only be one notary configuration or only one key defined within a given configuration, this configuration or key will be taken by default for verification purposes.
-2. Should there be more then one notary configuration or key, Connaisseur assumes that one of those has the name "default" and take this one by default.
+2. Should there be more then one notary configuration or key, Connaisseur assumes that one of those has the `name` field set to `default` and take this one by default.
```
Starkteetje, line 175:
```suggestion
2. Should there be more then one notary configuration or key and if one of the `name` fields is set to `default`, Connaisseur will take this one. If no default is specified and no specific entry defined, Connaisseur will abort the validation with an error.
```

I'd like this to be a bit more specific, along the lines of the above, if that's fine with you.

Starkteetje: Would then do the same change in the setup/README

connaisseur/tests/data/sample_config.yaml
```diff
@@ -0,0 +1,47 @@
+- name: default
```
Starkteetje, line 1: Would maybe also make an explicit test with a config that does not set the `default`s

setup/README.md
```diff
@@ -244,7 +244,7 @@ Your signed images were allowed through, in contrast to those unsigned ones with
 
 ## Cleanup
 
-If you want to remove Connaisseur, run `make uninstall`. We do not delete the namespace, since you might have deployed things in there, which we would inadvertently delete. If you want to get rid of the namespace, run `kubectl delete ns WHATEVEROURDEFAULTIS`.
+If you want to remove Connaisseur, run `make uninstall`.
```
Starkteetje: Why are you dropping the namespace cleanup hint again?

phbelitz: cuz of this:
https://github.com/sse-secure-systems/connaisseur/pull/90#discussion_r576136886

Starkteetje (dismissed):

Starkteetje (dismissed):
lgtm
might wanna wait with merging until we discussed potential schema changes given nv2 (see open school), such that we can still potentially release other stuff in the meantime

peterthomassen (dismissed):
All (relevant) conversations have been resolved, and I reviewed the changes that came along with fixing the conversations. I did not do a full review again, as the PR is too large.

I'm confident that the code is good for production now; it needs a rebase though.

helm/values.yaml
```diff
@@ -21,47 +21,48 @@ service:
...
-  # selfsignedCert: |
-  #   -----BEGIN CERTIFICATE-----
-  #   -----END CERTIFICATE-----
+  root_keys:
+  - name: library
+    key: |
+      -----BEGIN PUBLIC KEY-----
+      MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEOXYta5TgdCwXTCnLU09W5T4M4r9f
+      QQrqJuADP6U7g5r9ICgPSmZuRHP/1AYUfOQW3baveKsT969EfELKj1lfCA==
+      -----END PUBLIC KEY-----
```
xopham, line 34: I like, but please add a comment what this key is

xopham: you should actually be able to do the same for connaisseur and the `k8s.gcr.io` images by providing the keys here and setting `verify: true` below, right? That would be nice, as it might allow us to verify all signatures even our own :-)

adr/ADR-1_connaisseur-bootstrap-sentinel.md
```diff
@@ -28,6 +28,8 @@ Deploy MutatingWebhookConfiguration through Helm when Connaisseur Pods are healt
 
 We chose option 1 over option 2, because it was important to us that a brief glance at Connaisseur's Namespace allows one to judge whether it is running properly. Option 3 was not chosen as the readiness status of Pods can be easily seen from the Service, whereas the health status would require querying every single Pod individually. We deemed that to be a very ugly, non-kubernetes-y solution and hence decided against it.
 
+**edit**: the outcome of this decision was further changed in [ADR-3](ADR-3_multi_notary_config.md).
```
Starkteetje, line 31: I would put this up in the `Status` of the ADR as "Amended in [ADR-3](ADR-3_multi_notary_config.md)" instead of in the outcome

adr/ADR-1_connaisseur-bootstrap-sentinel.md
```diff
@@ -4,6 +4,8 @@
 
 Accepted
```
Starkteetje, line 5:
```suggestion
Amended in [ADR 3](ADR-3_multi_notary_config.md).
```

adr/ADR-1_connaisseur-bootstrap-sentinel.md
```diff
@@ -4,6 +4,8 @@
...
 
+**edit**: the outcome of this decision was amended in [ADR-3](ADR-3_multi_notary_config.md).
```
Starkteetje, line 7:
```suggestion
```

Starkteetje (approved):
you did it :+1: 
rebaseodySSE survived

devopstales:
Thank you for your great work @phbelitz.  Is there a documentation how to use multiple notary server?

xopham:
Hey @devopstales! Thank you for your feedback. The documentation for how to configure multiple keys and notaries in Connaisseur can be found in the setup guide: https://github.com/sse-secure-systems/connaisseur/blob/develop/setup/README.md#configure-notary
Please note that the feature is currently only on the `develop` branch not yet released which means that the current version (and public image) of Connaisseur does not yet implement it. However, you can obviously build Connaisseur from the `develop` branch yourself via `make docker` and start using the feature. We'd be excited about your feedback. We might take another few weeks until this is ultimately released because it introduces a breaking change to the configuration and we plan to roll some further changes with it. Feel to checkout our GitHub Discussion announcement of **Connaisseur 2.0**: 
[MAJOR REALEASE: Multi-Key, Multi-Notary, Multi-Validator #151](https://github.com/sse-secure-systems/connaisseur/discussions/151)

PR #95, opened 2021-02-15
Fix/key validation

phbelitz:
Whenever a public key is loaded into the key store, it's immediatly converted into a ecdsa.VerifyingKey to validated it's in a correct format. Otherwise an error is desplayed, informing which excat key is in an invalid format.

Includes commits from #90


phbelitz:
will be fixed as part of PR #90

PR #100, opened 2021-03-07
Enable Per-Namespace verification

operatorequals:
This commit aims to add the feature to enable the Connaiseur
Webhook validation in *some* K8s namespaces and not the whole
cluster.

This is done by adding a variable in the `values.yaml` file
called `targetNamespaces`, that contains a list of namespace names.
The webhook will then be used by container creations ONLY in these
namespaces.

In case the names in the list are not available as namespaces,
they are auto-created.

Finally, the default value of `['*']`, deploys the default behavior -
validating all the cluster containers.

### DIFF
--- a/helm/templates/namespace.yaml
+++ b/helm/templates/namespace.yaml
@@ -1,0 +1,11 @@
+{{- if not (has "*" .Values.targetNamespaces) }}
+{{- range .Values.targetNamespaces }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ . | quote }}
+  labels:
+    {{- include  "connaisseur.namespaceLabels" $ | indent 4 }}
+---
+{{- end }}        
+{{- end }}        
--- a/helm/templates/webhook/_webhook.tpl
+++ b/helm/templates/webhook/_webhook.tpl
@@ -1,0 +1,5 @@
+{{- define "connaisseur.namespaceLabels" }}
+de.securesystems.connaisseur/enabled: "true"
+de.securesystems.connaisseur/release: "{{ .Release.Name }}"
+de.securesystems.connaisseur/rootPubKeyHash: "{{ trim .Values.notary.rootPubKey | sha1sum }}"
+{{ end -}}
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -68,4 +68,8 @@ policy:
 # interrupting operation.
 detection_mode: false
 
+# A list of namespaces that will be subsequent to Connaisseur verification.
+# If the list contains '*' - all namespaces will be monitored.
+targetNamespaces: ['*']
+
 # debug: true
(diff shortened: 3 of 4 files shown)


operatorequals:
Hello people!
This is a really beautiful lead for Image verification in Kubernetes!

I am using this modified chart for the cluster I maintain at work so I thought you should have it too :)

Thanks again for the great software and verbose guide!

helm/values.yaml
```diff
@@ -68,4 +68,8 @@ policy:
 # interrupting operation.
 detection_mode: false
 
+# A list of namespaces that will be subsequent to Connaisseur verification.
```
peterthomassen, line 71:
```suggestion
# A list of namespaces that will be subject to Connaisseur verification.
```

operatorequals: My bad :)

peterthomassen:
Thanks for this nice PR, @operatorequals :-) I'll hand this over to people who understand more of K8s than I do.

helm/templates/webhook/_webhook.tpl
```diff
@@ -0,0 +1,5 @@
+{{- define "connaisseur.namespaceLabels" }}
```
phbelitz, line 1: heh. I think it's time we refactor our helm charts. This is much better than the current setup.

helm/templates/namespace.yaml
```diff
@@ -0,0 +1,11 @@
+{{- if not (has "*" .Values.targetNamespaces) }}
+{{- range .Values.targetNamespaces }}
+apiVersion: v1
+kind: Namespace
+metadata:
+  name: {{ . | quote }}
```
phbelitz, line 6: If `targetNamespaces` includes namespaces that already exist, the chart will fail, making the installation of connaissuer in an existing cluster a headache. On the other hand I do realize that we need the labels for matching to the webhook. 

My idea would be to add something like a `freshInstallation` field (boolean) to the `values.yaml`, indicating whether connaisseur is being installed in an empty cluster and can create the namespaces or not. Default would be `false`, so we do not create the namespaces ourselves, but assume their existance. Then we ideally need some kind of bash script, that can automatically update all the needed namespaces with the required labels.

It's a bit hack-y and not the clean "just do `helm install`" way, but it's the only thing I can come up with right now. Maybe you have a better idea?

operatorequals: I think the wisest thing to do is to change the requirement. Let's just make `targetNamespaces` a list of new Namespaces, admitting that it will fail for existing ones with the same name. Chances are that if you are about to initialize such a security measure to a Namespace, it'd better be empty.

phbelitz: Fair point.

phbelitz:
@operatorequals please do a rebase with the `develop` branch, then I'll merge the changes.

operatorequals:
Happily!

phbelitz (approved):

phbelitz:
@operatorequals oh well, you need to sign your commit :sweat_smile:

peterthomassen:
@phbelitz fyi, you can also do that if you want :) Clone @operatorequals's repo, check out his branch, `git commit --amend --no-edit -S`, `push -f`, and it should show up here. (You usually can write to other people's branches if that branch is connected to a pull request here.) The commit will then show up as authored by @operatorequals and committed by @phbelitz.

phbelitz:
Aha! TIL

phbelitz (approved):

PR #79, opened 2020-12-22
docs/test: Move test images into securesystems DockerHub repository

Starkteetje:
Previously, we had test images at docker.io/connytest/testimage but decided we wanted to have have the securesystems repository instead. This PR follows up on that decision.

Fix #41

PR #102, opened 2021-03-31
docs/fix: adjust public keys in READMEs to DockerHub Repo

xopham:
the issue was introduced in #79 when all testimages were moved to `docker.io/securesystemsengineering/testimage` but only the public key for integration tests was changed. Consequently, demo and other setup guides would not work anymore


Starkteetje (approved):
thx

PR #104, opened 2021-03-31
feat(helm): use helm to create certs

mkilchhofer:
I tried this project in my lab and I thought that maybe this is interesting for you.

As I use a [GitOps workflow](https://github.com/mkilchhofer/k8s-gitops/tree/main/clusters/k3s) to install things into my lab cluster, I could not use the Makefile approach. Helm is able to generate the certificates for you. The only downside is that every K8s resource using these certs must be defined in the same file as helm cannot share generated certificates across multiple files: https://github.com/Masterminds/sprig/issues/262

This was a quick-and-dirty way to fulfil this requirement.
Feel free to reject or extend this PR. :-)


phbelitz (approved):
@mkilchhofer Thank you very much for your contribution! I like this solution more than the previous implementation so I'm gonna merge this into our branch :)

phbelitz (approved):
gnar, signatures ...

mkilchhofer:
👍
Shall I add a GPG signature to my commit?

phbelitz:
I don't think this is necessary

PR #107, opened 2021-03-31
feat: sigstore/cosign support

xopham:
[sigstore](https://github.com/sigstore) project for software supply chain transparency and [sigstore/cosign](https://github.com/sigstore/cosign) for container signing and verification offer an alternative to Notary v1 and Notary v2. This is an initial tentative integration into Connaisseur.

Currently a rough first hack to get `cosign` on board and allow playing around with it as an experimental feature. Important next steps:

- [x] improve design and architectural approach
   - move signed digest identification into module
   - only use cosign when configured
   - fix parsing of stdout return
- [x] consider building signature verification ourselves to avoid having to inject cosign
   - [x] :arrow_right: create issue 
- [x] if injecting cosign: pull by version
   - [x] :arrow_right: create issue to verify signature
- [x] use `subprocess` module securely or avoid completely
   - rewrite the popen echo part properly
- [x] allow configuration via `values.yaml`
   - [x] :arrow_right: create separate issue/PR to handle additional/experimental features
- [x] how to handle same image with multiple signatures / different images with same tag but different signed digests
- [x] refactor and improve code quality
- [x] add error handling
- [x] tests, tests and more tests
   - [x] unit tests
   - [x] integration tests
   - [x] add `testimage:cosigned` to docker
- [x] README.md for cosign
- [x] comments

In case you want to test it, you will have to build the Conny image locally. To generate a key pair and sign your test image:
```
cosign generate-key-pair
cosign sign -key cosign.key <reponame>/<imagename>:signed
```
More details on cosign can be found in the docs: https://github.com/sigstore/cosign
The `cosign generate-key-pair` will yield a public key that has to be copied into `helm/values.yaml` as a trust root just like for Nv1.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,7 +6,9 @@
 
 <!-- # Connaisseur -->
 
-Connaisseur is an admission controller for Kubernetes that integrates Image Signature Verification into a cluster, as a means to ensure that only valid images are being deployed.
+Connaisseur is an admission controller for Kubernetes that integrates Image Signature Verification and Trust Pinning into a cluster, as a means to ensure that only valid images are being deployed.
+
+🚨 🚨 We have added experimental support for [Cosign](https://github.com/sigstore/cosign) from the [Sigstore Project](https://sigstore.dev/). You can find instructions [here](setup/COSIGN.md). 🚨 🚨
 
 Have feedback? Feel free to [reach out to us](https://github.com/sse-secure-systems/connaisseur/discussions)!
 
--- a/connaisseur/crypto.py
+++ b/connaisseur/crypto.py
@@ -2,6 +2,8 @@ import hashlib
 import base64
 import ecdsa
 
+from connaisseur.exceptions import InvalidPublicKey
+
 
 def verify_signature(public_base64: str, signature_base64: str, message: str):
     """
@@ -10,11 +12,24 @@ def verify_signature(public_base64: str, signature_base64: str, message: str):
 
     Raises ValidationError if unsuccessful.
     """
-    public = base64.b64decode(public_base64)
-    pub_key = ecdsa.VerifyingKey.from_der(public)
+    pub_key = decode_and_verify_ecdsa_key(public_base64)
 
     signature = base64.b64decode(signature_base64)
 
     msg_bytes = bytearray(message, "utf-8")
 
     return pub_key.verify(signature, msg_bytes, hashfunc=hashlib.sha256)
+
+
+def decode_and_verify_ecdsa_key(public_base64: str):
+    """
+    Verifies that the provided public key in base64 encoding qualifies as a
+    proper ecdsa key and throws if not.
+    """
+    try:
+        public = base64.b64decode(public_base64)
+        return ecdsa.VerifyingKey.from_der(public)
+    except Exception as err:
+        raise InvalidPublicKey(
+            f"The public key provided is not a base64-encoded ECDSA key: {err}."
+        ) from err
--- a/connaisseur/exceptions.py
+++ b/connaisseur/exceptions.py
@@ -29,6 +29,10 @@ class BaseConnaisseurException(Exception):
         return msg
 
 
+class InvalidPublicKey(BaseConnaisseurException):
+    pass
+
+
 class InvalidFormatException(BaseConnaisseurException):
     pass
 
@@ -54,6 +58,18 @@ class UnknownVersionError(Exception):
 
 
 class AmbiguousDigestError(BaseConnaisseurException):
+    pass
+
+
+class CosignError(BaseConnaisseurException):
+    pass
+
+
+class CosignTimeout(BaseConnaisseurException):
+    pass
+
+
+class UnexpectedCosignData(BaseConnaisseurException):
     pass
 
 
--- a/connaisseur/sigstore_validator.py
+++ b/connaisseur/sigstore_validator.py
@@ -1,0 +1,98 @@
+import json
+import logging
+import re
+import subprocess  # nosec
+
+from connaisseur.crypto import decode_and_verify_ecdsa_key
+from connaisseur.exceptions import (
+    CosignError,
+    CosignTimeout,
+    NotFoundException,
+    ValidationError,
+    UnexpectedCosignData,
+)
+
+
+def get_cosign_validated_digests(image: str, pubkey: str):
+    """
+    Gets and processes cosign validation output for a given `image` and `pubkey`
+    and either returns a list of valid digests or raises a suitable exception
+    in case no valid signature is found or cosign fails.
+    """
+    returncode, stdout, stderr = invoke_cosign(image, pubkey)
+    logging.info(
+        "COSIGN output for image: %s; RETURNCODE: %s; STDOUT: %s; STDERR: %s",
+        image,
+        returncode,
+        stdout,
+        stderr,
+    )
+    digests = []
+    if returncode == 0:
+        for sig in stdout.splitlines():
+            try:
+                sig_data = json.loads(sig)
+                try:
+                    digest = sig_data["Critical"]["Image"].get(
+                        "Docker-manifest-digest", ""
+                    )
+                    if re.match(r"sha256:[0-9A-Fa-f]{64}", digest) is None:
+                        raise Exception(
+                            f"digest '{digest}' does not match expected digest pattern."
+                        )
+                except Exception as err:
+                    raise UnexpectedCosignData(
+                        "could not retrieve valid and unambiguous digest "
+                        f"from data received by cosign: {type(err).__name__}: {err}"
+                    ) from err
+                # remove prefix 'sha256'
+                digests.append(digest[7:])
+            except json.JSONDecodeError:
+                logging.info("non-json signature data from cosign: %s", sig)
+                pass
+    elif "error: no matching signatures:\nunable to verify signature\n" in stderr:
+        raise ValidationError(
+            "failed to verify signature of trust data.",
+            {"trust_data_type": "dev.cosignproject.cosign/signature", "stderr": stderr},
+        )
+    elif re.match(r"^error: GET https://[^ ]+ MANIFEST_UNKNOWN:.*", stderr):
+        raise NotFoundException(
+            f'no trust data for image "{image}".',
+            {"trust_data_type": "dev.cosignproject.cosign/signature", "stderr": stderr},
+        )
+    else:
+        raise CosignError(
+            f'unexpected cosign exception for image "{image}": {stderr}.',
+            {"trust_data_type": "dev.cosignproject.cosign/signature"},
+        )
+    if not digests:
+        raise UnexpectedCosignData(
+            "could not extract any digest from data received by cosign "
+            "despite successful image verification."
+        )
+    return digests
+
+
+def invoke_cosign(image, pubkey):
+    """
+    Invokes a cosign binary in a subprocess for a specific `image` given a `pubkey` and
+    returns the returncode, stdout and stderr. Will raise an exception if cosign times out.
+    """
+
+    decode_and_verify_ecdsa_key(pubkey)  # raises if invalid; return value not used
+    cmd = ["/app/cosign/cosign", "verify", "-key", "/dev/stdin", image]
+    stdinput = f"-----BEGIN PUBLIC KEY-----\n{pubkey}\n-----END PUBLIC KEY-----"
+
+    with subprocess.Popen(  # nosec
+        cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    ) as process:
+        try:
+            stdout, stderr = process.communicate(bytes(stdinput, "utf-8"), timeout=60)
+        except subprocess.TimeoutExpired as err:
+            process.kill()
+            raise CosignTimeout(
+                "cosign timed out.",
+                {"trust_data_type": "dev.cosignproject.cosign/signature"},
+            ) from err
+
+    return process.returncode, stdout.decode("utf-8"), stderr.decode("utf-8")
--- a/connaisseur/tests/integration/cosign-integration-test.sh
+++ b/connaisseur/tests/integration/cosign-integration-test.sh
@@ -1,0 +1,52 @@
+#! /bin/bash
+set -euo pipefail
+
+# This script is expected to be called from the root folder of Connaisseur
+
+echo 'Preparing Connaisseur config...'
+yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' helm/values.yaml connaisseur/tests/integration/cosign-update.yaml
+echo 'Config set'
+
+echo 'Installing Connaisseur...'
+make install || { echo 'Failed to install Connaisseur'; exit 1; }
+echo 'Successfully installed Connaisseur'
+
+echo 'Testing unsigned image...'
+kubectl run pod --image=securesystemsengineering/testimage:co-unsigned >output.log 2>&1 || true
+
+if [[ "$(cat output.log)" != 'Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: no trust data for image "docker.io/securesystemsengineering/testimage:co-unsigned".' ]]; then
+  echo 'Failed to deny unsigned image or failed with unexpected error. Output:'
+  cat output.log
+  exit 1
+else
+  echo 'Successfully denied usage of unsigned image'
+fi
+
+echo 'Testing image signed under different key...'
+kubectl run pod --image=securesystemsengineering/testimage:co-signed-alt >output.log 2>&1 || true
+
+if [[ "$(cat output.log)" != 'Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: failed to verify signature of trust data.' ]]; then
+  echo 'Failed to deny image signed with different key or failed with unexpected error. Output:'
+  cat output.log
+  exit 1
+else
+  echo 'Successfully denied usage of image signed under different key'
+fi
+
+echo 'Testing signed image...'
+kubectl run pod --image=securesystemsengineering/testimage:co-signed >output.log 2>&1 || true
+
+if [[ "$(cat output.log)" != 'pod/pod created' ]]; then
+  echo 'Failed to allow signed image. Output:'
+  cat output.log
+  exit 1
+else
+  echo 'Successfully allowed usage of signed image'
+fi
+
+echo 'Uninstalling Connaisseur...'
+make uninstall || { echo 'Failed to uninstall Connaisseur'; exit 1; }
+echo 'Successfully uninstalled Connaisseur'
+
+rm output.log
+echo 'Passed integration test'
--- a/connaisseur/tests/integration/cosign-update.yaml
+++ b/connaisseur/tests/integration/cosign-update.yaml
@@ -1,0 +1,22 @@
+deployment:
+  imagePullPolicy: Never
+notary:
+  host: notary.docker.io
+  selfsigned: false
+  auth:
+    enabled: false
+  rootPubKey: |
+    -----BEGIN PUBLIC KEY-----
+    MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEvtc/qpHtx7iUUj+rRHR99a8mnGni
+    qiGkmUb9YpWWTS4YwlvwdmMDiGzcsHiDOYz6f88u2hCRF5GUCvyiZAKrsA==
+    -----END PUBLIC KEY-----
+  isCosign: true
+policy:
+  - pattern: "*:*"
+    verify: true
+  - pattern: "k8s.gcr.io/*:*"
+    verify: false
+  - pattern: "docker.io/securesystemsengineering/connaisseur:*"
+    verify: true
+  - pattern: "docker.io/securesystemsengineering/connaisseur:helm-hook-*"
+    verify: false
--- a/docker/Dockerfile
+++ b/docker/Dockerfile
@@ -8,6 +8,14 @@ WORKDIR /install
 COPY requirements.txt /requirements.txt
 RUN pip install --no-cache-dir --prefix=/install -r /requirements.txt
 
+# build cosign
+FROM golang:buster as go_builder
+
+ARG COSIGN_VERSION
+RUN git clone -b "v${COSIGN_VERSION}" --single-branch --depth 1 https://github.com/sigstore/cosign.git
+WORKDIR /go/cosign
+RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./cmd/cosign/main.go
+
 # build connaisseur image
 FROM base
 
@@ -19,6 +27,7 @@ RUN sh /harden.sh && rm /harden.sh
 
 # Copy source code and install packages
 COPY --from=builder /install /usr/local
+COPY --from=go_builder /go/cosign/main /app/cosign/cosign
 COPY connaisseur /app/connaisseur
 
 USER 1000:2000
--- a/setup/COSIGN.md
+++ b/setup/COSIGN.md
@@ -1,0 +1,111 @@
+# Using Connaisseur with Cosign signatures
+
+[Sigstore](https://sigstore.dev/) is a [Linux Foundation](https://linuxfoundation.org/) project that aims to provide public software signing and transparency to improve open source supply chain security. As part of the Sigstore project, [Cosign](https://github.com/sigstore/cosign) allows seamless container signing, verification and storage. You can read more about it [here](https://blog.sigstore.dev/cosign-signed-container-images-c1016862618).
+
+Connaisseur currently supports the elementary function of verifying Cosign-generated signatures against the locally created corresponding public keys. We plan to expose further features of Cosign and Sigstore in upcoming releases, so stay tuned!
+
+> :warning: Sigstore and Cosign are currently in *pre-release* state and under heavy development and so is our support for them. We therefore consider this an *experimental feature* that might unstable over time. As such, it is not part of our semantic versioning guarantees and we take the liberty to adjust or remove it with any version at any time without incrementing MAJOR or MINOR.
+
+## Demo
+![](../img/connaisseur_cosign.gif)
+
+## Signing Container Images with Cosign
+
+> **NOTE**: You can also do a minimal test without installing Cosign locally. In that case, skip this step and use our provided public key and `testimage`s below.
+
+Getting started with Cosign is very well described in the [docs](https://github.com/sigstore/cosign). Please check there for detailed instructions. The currently supported version can be found in our [Makefile](https://github.com/sse-secure-systems/connaisseur/blob/master/Makefile/#L5). In short: After installation, a keypair is generated via:
+
+```bash
+cosign generate-key-pair
+```
+
+You will be prompted to set a password, after which a private (`cosign.key`) and public (`cosign.pub`) key are created. You can then use Cosign to sign a container image using:
+
+```bash
+# Here, $IMAGE is REPOSITORY/IMAGE_NAME:TAG
+cosign sign -key cosign.key $IMAGE
+```
+
+The created signature can be verfied via:
+
+```bash
+cosign verify -key cosign.pub $IMAGE
+```
+
+
+
+## Configuring Connaisseur for Cosign Signatures
+
+Setting up Connaisseur for Cosign signatures only requires minor changes. In case of questions, please refer to the [default guide](README.md). In essence, you can just clone this repository:
+
+```bash
+git clone https://github.com/sse-secure-systems/connaisseur.git
+cd connaisseur
+```
+
+Next, configure Connaisseur to use Cosign and the previously created public key for validation via the `helm/values.yaml`.  To do so, copy your `cosign.pub` key into `notary.rootPubkey`. Here, at the example of our public key for our `testimage`s (you can also use this key and test with our images):
+
+```yaml
+# Replace the actual key part with your own key
+  rootPubKey: |
+    -----BEGIN PUBLIC KEY-----
+    MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEvtc/qpHtx7iUUj+rRHR99a8mnGni
+    qiGkmUb9YpWWTS4YwlvwdmMDiGzcsHiDOYz6f88u2hCRF5GUCvyiZAKrsA==
+    -----END PUBLIC KEY-----
+```
+
+Next, set `notary.isCosign` to `true`:
+
+```yaml
+  isCosign: true
+```
+
+And finally, install Connaisseur via:
+
+```bash
+make install
+```
+
+
+
+## Test Signature Verification
+
+You can now test signature verification by deploying the signed image from above:
+
+```bash
+kubectl run signed --image=$IMAGE
+```
+
+Or attempt to deploy any another unsigned image which will fail.
+
+For the lazy ones, you can configure our public key provided in the previous section and test with our signed `testimage`:
+
+```bash
+kubectl run signed --image=docker.io/securesystemsengineering/testimage:co-signed
+```
+
+and compare to the unsigned `testimage`:
+
+```bash
+kubectl run unsigned --image=docker.io/securesystemsengineering/testimage:co-unsigned
+```
+
+or a `testimage` signed with a different key:
+
+```bash
+kubectl run unsigned --image=docker.io/securesystemsengineering/testimage:co-signed-alt
+```
+
+Once finished, you can clean up your cluster via:
+
+```bash
+make uninstall
+```
+
+
+
+## End
+
+Hope you enjoy the new feature and let us [know what you think](https://github.com/sse-secure-systems/connaisseur/discussions/137)!
+
+We are working on improving support for Cosign and Sigstore features. So stay tuned!
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -224,6 +224,35 @@ jobs:
         run: |
           kind load docker-image $(yq e '.deployment.image' helm/values.yaml)
           kind load docker-image $(yq e '.deployment.helmHookImage' helm/values.yaml)
-
       - name: Run actual integration test
         run: bash connaisseur/tests/integration/namespaced-integration-test.sh
+
+  cosign-integration-test:
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/download-artifact@v2
+        with:
+          name: images
+      - name: Load Docker images
+        run: |
+          docker load -i ${GITHUB_SHA}_image.tar
+          docker load -i ${GITHUB_SHA}_hook.tar
+      - name: Install yq and bash
+        run: |
+          sudo snap install yq
+          sudo apt update
+          sudo apt install bash -y
+      - name: Create KinD cluster
+        run: |
+          GO111MODULE="on" go get sigs.k8s.io/kind
+          kind create cluster --wait 120s
+      - name: Check KinD cluster
+        run: kubectl get nodes
+      - name: Add images to KinD
+        run: |
+          kind load docker-image $(yq e '.deployment.image' helm/values.yaml)
+          kind load docker-image $(yq e '.deployment.helmHookImage' helm/values.yaml)
+      - name: Run actual integration test
+        run: bash connaisseur/tests/integration/cosign-integration-test.sh
--- a/.github/workflows/dockerhub-check.yml
+++ b/.github/workflows/dockerhub-check.yml
@@ -17,5 +17,11 @@ jobs:
         run: DOCKER_CONTENT_TRUST=1 docker pull docker.io/securesystemsengineering/testimage:signed
       - name: Check unsigned test image
         run: DOCKER_CONTENT_TRUST=0 docker pull docker.io/securesystemsengineering/testimage:unsigned
+      - name: Check cosign signed test image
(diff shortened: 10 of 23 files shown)


connaisseur/tests/integration/cosign-integration-test.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if [[ "$(cat output.log)" != 'pod/pod created' ]]; then
+  echo 'Failed to allow signed image. Output:'
+  cat output.log
+  exit 1
+else
+  echo 'Successfully allowed usage of signed image'
+fi
+
+#echo 'Testing deployment of unsigned init container along with a valid container...'
```
annekebr, line 47: Why is it commented out? Could simply be removed if not needed couldn't it?

connaisseur/tests/integration/cosign-update.yaml
```diff
@@ -0,0 +1,22 @@
...
+  host: notary.docker.io
+  selfsigned: false
+  auth:
+    enabled: false
+  rootPubKey: |
+    -----BEGIN PUBLIC KEY-----
+    MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEUU2/8oNKPRTVjnTPY4WwUNFXYke3
+    CmpJpSaqBOFh92bdXZENv0AdR5qipqWxjGWO28Miyc2O6vGuoIyzzhQvuA==
+    -----END PUBLIC KEY-----
+  isCosign: true
```
annekebr, line 13: Would mark the experimental in the name of the var e.g. `isCosign(experimental)` to be on the very safe side

setup/COSIGN.md
```diff
@@ -0,0 +1,98 @@
...
+```bash
+kubectl run signed --image=<IMAGE>
+```
+
+Or attempt to deploy another unsigned image.
+
+For the lazy ones, you can just configure our public key provided in the previous section and test with:
+
+```bash
+kubectl run signed --image=docker.io/xoph/ctest:signed
```
annekebr, line 80: We should move the image to the other hub before we release

setup/COSIGN.md
```diff
@@ -0,0 +1,101 @@
...
+
+```bash
+cosign verify -key cosign.pub <IMAGE>
+```
+
+
+
+## Verification with Connaisseur
+
+Setting up Connaisseur (or short: `conny`) for `cosign` signatures only requires minor changes. In case of questions, please refer to the [default guide](README.md). In essence, you can just clone this repository:
```
annekebr, line 35: wouldn't name Connaisseur by it's nickname in the official docs

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,76 @@
...
+    digests = []
+    if not failed:
+        for sig in stdout.splitlines():
+            try:
+                sig_data = json.loads(sig)
+                digests.append(sig_data["Critical"]["Image"]["Docker-manifest-digest"])
+            except json.JSONDecodeError:
+                logging.info("Non-json signature data from Cosign: %s", sig)
+                pass
+    elif stderr == "error: no matching signatures:\nunable to verify signature\n":
```
annekebr, line 37: Do these errors differ in their return codes? Maybe we could go for the return code as it might be stabler than the message itself.

docker/Dockerfile
```diff
@@ -19,6 +27,7 @@ RUN sh /harden.sh && rm /harden.sh
 
 # Copy source code and install packages
 COPY --from=builder /install /usr/local
+COPY --from=go_builder /go/cosign/main /app/cosign/cosign
```
annekebr, line 30: We must be aware that we change the official docker image to include an experimental binary. Maybe we should release a version with and without cosign as long as cosign is experimental/we support this feature experimentally, what do you think?

setup/COSIGN.md
xopham: actually a bigger fan of the nickname than the real name bc it is way too long and confuses people, but will do

connaisseur/sigstore_validator.py
xopham: I wanted to keep it compatible with the standard connaisseur error messages to make it more understandable once we have multiple validators

docker/Dockerfile
xopham: thought about it and felt like it will cause significant extra work. If you do not choose to use it, it will never be called. So I feel it is safe enough :shrug:

connaisseur/sigstore_validator.py
annekebr: That's not what I meant, sorry. I was referring to line 37 and the error message that comes from `cosign`. If the binary returns different error codes we could use these instead of the error message to avoid having to change code if they change their error messages (which I could imagine to change rather than the codes)

setup/COSIGN.md
xopham: yes, mentioned it as a task in the PR

annekebr: ok :+1: I did see that but wanted to doublecheck whith regards to the time frame ;)

connaisseur/tests/integration/cosign-update.yaml
xopham: you meant in the `helm/values.yaml` right? There is a comment above.

connaisseur/sigstore_validator.py
xopham: Gotcha. There is unfortunately no separate error codes. The output is still a bit hard to parse...

connaisseur/tests/integration/cosign-update.yaml
annekebr: I know, that's why I said "very" :-)

xopham: I made a comment in the helm/values.yaml just above the `isCosign`. The advantage of a comment would be that in case someone is using it, they will not have to change the flag when it becomes non-experimental, but I am open about it. Opinions?

Starkteetje: I'd say with the comment in README, COSIGN and values.yml and the fact that cosign itself is experimental it's fine to not have it in the name

annekebr: fine with me :-)

connaisseur/sigstore_validator.py
annekebr: alrighty then :-)

docker/Dockerfile
Starkteetje: It's not only about security though, image size also goes up from 56,5 to 82,4 MB, which is quite significant for at least currently being an experiment of an experiment. What about having an experimental Dockerfile (for connaisseur:$VERSION-experimental), which in this case would include cosign and e.g. replacing the cosign binary with the `false` binary in the main image?

connaisseur/tests/integration/cosign-integration-test.sh
xopham: done

setup/COSIGN.md
xopham: it is done :rocket:

setup/COSIGN.md
xopham: duly done :disappointed:

connaisseur/crypto.py
```diff
@@ -10,11 +12,26 @@ def verify_signature(public_base64: str, signature_base64: str, message: str):
...
+    Verifies that the provided public key in base64 encoding qualifies as a
+    proper ecdsa key and throws if not.
+    """
+    public = base64.b64decode(public_base64)
+    try:
+        pubkey = ecdsa.VerifyingKey.from_der(public)
+    except ecdsa.keys.MalformedPointError as err:
+        raise InvalidPublicKey(
+            f"The public key provided is no valid ecdsa key: {err}."
+        ) from err
```
peterthomassen, line 35:
```suggestion
        )
```

> Exception chaining happens automatically when an exception is raised inside an `except` or `finally` section.
Source: https://docs.python.org/3/tutorial/errors.html

connaisseur/crypto.py
```diff
@@ -10,11 +12,26 @@ def verify_signature(public_base64: str, signature_base64: str, message: str):
...
+    """
+    Verifies that the provided public key in base64 encoding qualifies as a
+    proper ecdsa key and throws if not.
+    """
+    public = base64.b64decode(public_base64)
+    try:
+        pubkey = ecdsa.VerifyingKey.from_der(public)
+    except ecdsa.keys.MalformedPointError as err:
+        raise InvalidPublicKey(
+            f"The public key provided is no valid ecdsa key: {err}."
```
peterthomassen, line 34:
```suggestion
            f"The public key provided is not a base64-encoded ECDSA key: {err}."
```

connaisseur/exceptions.py
```diff
@@ -57,6 +61,14 @@ class AmbiguousDigestError(BaseConnaisseurException):
     pass
 
 
+class CosignError(BaseConnaisseurException):
+    pass
+
+
+class CosignTimeout(BaseConnaisseurException):
+    pass
```
peterthomassen, line 69: Maybe create an exception hierarchy, with `BaseConaisseurException` --> `CosignException` --> `CosignError` etc., to avoid polluting the namespace once several attestation mechanisms are supported.

Speaking of disentangling things, it may also be worth thinking about whether stuff specific to an attestation method should live inside a subdirectory (here: `cosign/exceptions.py`, `cosign/validate.py` etc., similar for Notary v1/v2).

docker/Dockerfile
```diff
@@ -8,6 +8,14 @@ WORKDIR /install
 COPY requirements.txt /requirements.txt
 RUN pip install --no-cache-dir --prefix=/install -r /requirements.txt
 
+# build cosign
+FROM golang:buster as go_builder
+
+ARG COSIGN_VERSION
+RUN git clone -b "v${COSIGN_VERSION}" --single-branch --depth 1 https://github.com/sigstore/cosign.git
+WORKDIR /go/cosign
+RUN GO111MODULE=on  CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./cmd/cosign/main.go
```
peterthomassen, line 17:
```suggestion
RUN GO111MODULE=on CGO_ENABLED=0 GOOS=linux GOARCH=amd64 go build ./cmd/cosign/main.go
```

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,78 @@
...
+        )
+    return digests
+
+
+def invoke_cosign(image, pubkey):
+    """
+    Invokes a cosign binary in a subprocess for a specific `image` given a `pubkey` and
+    returns the returncode, stdout and stderr. Will raise an exception if cosign times out.
+    """
+    decode_and_verify_ecdsa_key(pubkey)
```
peterthomassen, line 63:
```suggestion
    decode_and_verify_ecdsa_key(pubkey)  # raises if invalid; return value not used
```

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,78 @@
...
+    cmd = ["/app/cosign/cosign", "verify", "-key", "/dev/stdin", image]
+    stdinput = "-----BEGIN PUBLIC KEY-----\n{pubkey}\n-----END PUBLIC KEY-----".format(
+        pubkey=pubkey
+    )
```
peterthomassen, line 67:
```suggestion
    stdinput = f"-----BEGIN PUBLIC KEY-----\n{pubkey}\n-----END PUBLIC KEY-----"
```

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,78 @@
...
+    process = subprocess.Popen(  # nosec
+        cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    )
```
peterthomassen, line 70:
```suggestion
    with subprocess.Popen(  # nosec
        cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE
    ) as process:
    # indent the following lines
```

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,78 @@
...
+    try:
+        stdout, stderr = process.communicate(bytes(stdinput, "utf-8"), timeout=120)
+    except subprocess.TimeoutExpired as err:
+        raise CosignTimeout(
+            "Cosign timed out.",
+            {"trust_data_type": "dev.cosignproject.cosign/signature"},
+        ) from err
```
peterthomassen, line 77: remove `err`, also from `except` line

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,78 @@
...
+    stdinput = "-----BEGIN PUBLIC KEY-----\n{pubkey}\n-----END PUBLIC KEY-----".format(
+        pubkey=pubkey
+    )
+    process = subprocess.Popen(  # nosec
+        cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    )
+    try:
+        stdout, stderr = process.communicate(bytes(stdinput, "utf-8"), timeout=120)
+    except subprocess.TimeoutExpired as err:
+        raise CosignTimeout(
```
peterthomassen, line 74:
```suggestion
        process.kill()
        raise CosignTimeout(
```

Otherwise, process may linger, see https://docs.python.org/3/library/subprocess.html#subprocess.Popen.communicate.

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,78 @@
...
+)
+
+
+def get_cosign_validated_digests(image: str, pubkey: str):
+    """
+    Gets and processes cosign validation output for a given `image` and `pubkey`
+    and either returns a list of valid digests or raises a suitable exception
+    in case no valid signature is found or cosign fails.
+    """
+    failed, stdout, stderr = invoke_cosign(image, pubkey)
```
peterthomassen, line 21: `failed` is a misnomer, as it sounds boolean, but it is an int.

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,78 @@
...
+    logging.info(
+        "COSIGN output for image: %s; RETURNCODE: %s; STDOUT: %s; STDERR: %s",
+        image,
+        failed,
+        stdout,
+        stderr,
+    )
+    digests = []
+    if not failed:
```
peterthomassen, line 30:
```suggestion
    if failed == 0:
```

`failed` is `None` if for some reason the process has no return code (not exited yet, but maybe we arrived here by mishandling a timeout?). In this case, `not failed` evaluates to `True`, and we're off on the wrong path. (Explicit is better than implicit!)

setup/COSIGN.md
```diff
@@ -0,0 +1,101 @@
+# Using Connaisseur with Cosign signatures
+
+[Sigstore](https://sigstore.dev/) is a [Linux Foundation](https://linuxfoundation.org/) project that aims to provide public software signing and transparency to improve open source supply chain security. As part of the Sigstore project, [Cosign](https://github.com/sigstore/cosign) allows seamless container signing, verification and storage. You can read more about it [here](https://blog.sigstore.dev/cosign-signed-container-images-c1016862618?gi=95796f563c9f).
```
peterthomassen, line 3:
```suggestion
[Sigstore](https://sigstore.dev/) is a [Linux Foundation](https://linuxfoundation.org/) project that aims to provide public software signing and transparency to improve open source supply chain security. As part of the Sigstore project, [Cosign](https://github.com/sigstore/cosign) allows seamless container signing, verification and storage. You can read more about it [here](https://blog.sigstore.dev/cosign-signed-container-images-c1016862618).
```

Remove tracking parameter.

setup/COSIGN.md
```diff
@@ -0,0 +1,101 @@
...
+
+As a first step, we have implemented the elementary function of verifying `cosign` generated signatures against the locally created corresponding public keys. We plan to expose further features of cosign in upcoming releases, so stay tuned!
```
peterthomassen, line 5:
```suggestion
Connaisseur supports the elementary function of verifying `cosign`-generated signatures against the locally created corresponding public keys. We plan to expose further features of cosign in upcoming releases, so stay tuned!
```

setup/COSIGN.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Demo
+![](../img/co-nny-sign.gif)
```
peterthomassen, line 8: I wonder if this filename will later be understood.

setup/COSIGN.md
```diff
@@ -0,0 +1,101 @@
...
+
+## Signing Container Images with Cosign
+
+Getting started with `cosign` is very well described in the [docs](https://github.com/sigstore/cosign). Please check there for detailed instructions. In short: After installation a keypair is generated via:
+
+```bash
+cosign generate-key-pair
+```
+
+You will be prompted to enter a password and a private (`cosign.key`) and public (`cosign.pub`) key are created. You can then use cosign to sign a container image using:
```
peterthomassen, line 18:
```suggestion
You will be prompted to enter a password, after which a private (`cosign.key`) and public (`cosign.pub`) key are created. You can then use cosign to sign a container image using:
```

setup/COSIGN.md
```diff
@@ -0,0 +1,101 @@
...
+
+```bash
+# Here, <IMAGE> is REPOSITORY/IMAGE_NAME:TAG
+cosign sign -key cosign.key <IMAGE>
+```
+
+You can then verify the created signature via:
+
+```bash
+cosign verify -key cosign.pub <IMAGE>
```
peterthomassen, line 28: I'd prefer `$IMAGE`, as `<` and `>` have special shell meaning.

setup/COSIGN.md
```diff
@@ -0,0 +1,101 @@
...
+
+```bash
+git clone https://github.com/sse-secure-systems/connaisseur.git
+cd connaisseur
+```
+
+Next, configure `Connaisseur` to use cosign and the previously created public key for validation via the `helm/values.yaml`.  To do so, copy your `cosign.pub` key into `notary.rootPubkey`:
+
+```bash
+# the public part of the root key, for verifying notary's signatures
```
peterthomassen, line 45:
```suggestion
# The public part of the root key, for verifying notary's signatures. Example:
```

setup/COSIGN.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+
+
+## Test Signature Verification
+
+You can now test signature verification by deploying the signed image from above:
+
+```bash
+kubectl run signed --image=<IMAGE>
```
peterthomassen, line 72: `$IMAGE`

setup/COSIGN.md
```diff
@@ -0,0 +1,101 @@
...
+```
+
+Or attempt to deploy another unsigned image.
+
+For the lazy ones, you can just configure our public key provided in the previous section and test with:
```
peterthomassen, line 77:
```suggestion
For the lazy ones, you can configure our public key provided in the previous section and test with:
```

peterthomassen (dismissed):
nice work!

docker/Dockerfile
Starkteetje: With the base image being 44,9MB, the cosign update is double the size of the Connaisseur footprint. In the past, we've had the issue of A's Internet connection not being able to download the image in time for a two-minute sentinel timeout (and also had tickets regarding that) and the increase for an experimental inclusion of an experimental signing method would increase this size by 46%. I see your reasoning with feature flags, but I feel like the size increase is unwarranted given the very experimental nature

setup/COSIGN.md
xopham: that is actually a copy/paste from the `helm/values.yaml`. I might just drop that line here, as it does not add value and is only confusing

peterthomassen: We should in any case emphasize that the key given is an example.

connaisseur/exceptions.py
xopham: Maybe we should move this into the overall refactoring: https://github.com/sse-secure-systems/connaisseur/pull/101. Like keep it flat for now and then choose a proper solution with the central refactoring? Otherwise, I would feel that part of the code base is going ahead in one of the directions and it will be changed later anyways...

connaisseur/crypto.py
xopham: pylint does not like that: https://github.com/sse-secure-systems/connaisseur/runs/2295833587?check_suite_focus=true

> crypto.py:33:8: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
They like explicit better apparently...

xopham: Think I should tell pylint to ignore it or make it explicit? In general, explicit is better than implicit, but at this point I am not sure whether that rule is good...

xopham: well actually there is apparently a difference: https://stefan.sofa-rockers.org/2020/10/28/raise-from/
will re-add the `from err`

connaisseur/exceptions.py
xopham: @phbelitz what do you think? Should I delay this until the refactoring is done to consider multi-validator support?

phbelitz: if this is merged before the refactoring, then yes keep it flat and i will make things consistent in the refactoring PR

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,76 @@
...
+        for sig in stdout.splitlines():
+            try:
+                sig_data = json.loads(sig)
+                digests.append(sig_data["Critical"]["Image"]["Docker-manifest-digest"])
```
annekebr, line 34: This means digests will be a list with entries of the form `sha256:<hash>`. The digests we obtain from notary are handled as plain `<hash>` without `sha256`. So, if I am not mistaken the image name is going to  be changed to `<image>@sha256:sha256:<hash>` [here](https://github.com/sse-secure-systems/connaisseur/blob/master/connaisseur/image.py#L86) which I would assume to result in an invalid image name, which cannot be pulled by k8s.

annekebr: Yep, seems so. Connaisseur admits this image in my test setup but then the image cannot be pulled:

```
  Warning  InspectFailed  4s (x3 over 6s)  kubelet, minikube  Failed to apply default image tag "docker.io/securesystemsengineering/testimage@sha256:sha256:c5327b291d702719a26c6cf8cc93f72e7902df46547106a9930feda2c002a4a7": couldn't parse image reference "docker.io/securesystemsengineering/testimage@sha256:sha256:c5327b291d702719a26c6cf8cc93f72e7902df46547106a9930feda2c002a4a7": invalid reference format
  Warning  Failed         4s (x3 over 6s)  kubelet, minikube  Error: InvalidImageName
```

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,76 @@
...
+            "failed to verify signature of trust data.",
+            {"trust_data_type": "dev.cosignproject.cosign/signature", "stderr": stderr},
+        )
+    elif re.match(r"^error: GET https://[^ ]+ MANIFEST_UNKNOWN:.*", stderr):
+        raise NotFoundException(
+            f'no trust data for image "{image}".',
+            {"trust_data_type": "dev.cosignproject.cosign/signature", "stderr": stderr},
+        )
+    else:
+        raise CosignError(
```
annekebr, line 49: For the sake of completeness: If the process dies for other reasons (e.g. there is an issue with the docker image and there is a dysfunctional `cosign` binary) this would result in a `CosignError` which one could argue it not really is. As we also include the error message that could be fine, but I could also image that `CosignError` and `image` could mislead the user in that case. I don't know whether these types could be differentiated at all, but still mentioning ;-)

connaisseur/sigstore_validator.py
xopham: oh wow...how could I miss this :disappointed: great catch!!! Will fix it

connaisseur/sigstore_validator.py
xopham: yes, thought about that too, but would not know how to distinguish. Was hoping the error message solves it

connaisseur/sigstore_validator.py
xopham: besides me being feeling bad for missing this, we should start checking in the integration tests whether we actually derived a valid digest

README.md
```diff
@@ -6,7 +6,9 @@
 
 <!-- # Connaisseur -->
 
-Connaisseur is an admission controller for Kubernetes that integrates Image Signature Verification into a cluster, as a means to ensure that only valid images are being deployed.
+Connaisseur is an admission controller for Kubernetes that integrates Image Signature Verification and Trust Pinning into a cluster, as a means to ensure that only valid images are being deployed.
+
+🚨 🚨 We have added experimental support for [Cosign](https://github.com/sigstore/cosign) from the [Sigstore Project](https://sigstore.dev/). You can find instructions [here](setup/COSIGN.md). 🚨 🚨
```
Starkteetje, line 11: Maybe add here that experimental does mean that it is not part of our semantic versioning guarantees and we take the liberty to remove it with any version at any time without incrementing major

connaisseur/crypto.py
```diff
@@ -10,11 +12,26 @@ def verify_signature(public_base64: str, signature_base64: str, message: str):
...
 
     return pub_key.verify(signature, msg_bytes, hashfunc=hashlib.sha256)
+
+
+def decode_and_verify_ecdsa_key(public_base64: str):
+    """
+    Verifies that the provided public key in base64 encoding qualifies as a
+    proper ecdsa key and throws if not.
+    """
+    public = base64.b64decode(public_base64)
```
annekebr, line 29: Is there a reason for this statement not being in the `try` block? What happens if it's invalid base64?

connaisseur/sigstore_validator.py
xopham: renamed to `returncode` as in the subprocess module

connaisseur/sigstore_validator.py
xopham: done

connaisseur/sigstore_validator.py
xopham: kept as it exception chaining is not fully explicit, as discussed in the other thread

connaisseur/crypto.py
```diff
@@ -10,11 +12,26 @@ def verify_signature(public_base64: str, signature_base64: str, message: str):
...
+    try:
+        pubkey = ecdsa.VerifyingKey.from_der(public)
+    except ecdsa.keys.MalformedPointError as err:
```
annekebr, line 32: Is there a reason for only catching this kind of Error? I encountered a `ecdsa.der.UnexpectedDER` error - wouldn't we want to catch that too?

README.md
xopham: Will do that on the actual `COSIGN.md`

xopham: done

setup/COSIGN.md
xopham: duh...newb mistake :facepalm:

connaisseur/crypto.py
xopham: done. Now catching all.

connaisseur/crypto.py
xopham: done. Now catching all.

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,93 @@
...
+    logging.info(
+        "COSIGN output for image: %s; RETURNCODE: %s; STDOUT: %s; STDERR: %s",
+        image,
+        returncode,
+        stdout,
+        stderr,
+    )
+    digests = []
+    if returncode == 0:
+        for sig in stdout.splitlines():
```
Starkteetje, line 32: If stdout.splitlines is empty, you won't catch that here. It will fail later, because then `digests = []`. Dunno, whether that is something you'd want to check for here to have a more specific error or whether you're fine with it failing later due to no digests.

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,93 @@
...
+                except Exception as err:
+                    raise UnexpectedCosignData(
+                        f"could not retrieve valid digest from data received by cosign: {err}"
+                    ) from err
+
+                # remove prefix 'sha256' in case it exists
+                digests.append(digest[7:] if digest[:7] == "sha256:" else digest)
+            except json.JSONDecodeError:
+                logging.info("Non-json signature data from Cosign: %s", sig)
+                pass
```
Starkteetje, line 52: Why do you pass here instead of raising an exception? Given that all other error conditions raise, I'd have expected that here as well

connaisseur/sigstore_validator.py
```diff
@@ -0,0 +1,93 @@
...
+
+    decode_and_verify_ecdsa_key(pubkey)  # raises if invalid; return value not used
+    cmd = ["/app/cosign/cosign", "verify", "-key", "/dev/stdin", image]
+    stdinput = f"-----BEGIN PUBLIC KEY-----\n{pubkey}\n-----END PUBLIC KEY-----"
+
+    with subprocess.Popen(  # nosec
+        cmd, stdin=subprocess.PIPE, stdout=subprocess.PIPE, stderr=subprocess.PIPE
+    ) as process:
+        try:
+            stdout, stderr = process.communicate(bytes(stdinput, "utf-8"), timeout=120)
```
Starkteetje, line 85: 120 seems to be a very long timeout given that it halts the deployment in the meantime. Would vote for max 60, or is it an expectation that verifying a cosign signature could take longer than a minute?

setup/COSIGN.md
```diff
@@ -0,0 +1,111 @@
...
+make uninstall
+```
+
+
+
+## End
+
+Hope you enjoy the new feature and let us [know what you think](../CONTRIBUTING.md)!
+
+We are working on improving support for `cosign` and `sigstore` features and will also be introducing [multi-key and multi-validator support](https://github.com/sse-secure-systems/connaisseur/pull/90) in the near future. So stay tuned!
```
Starkteetje, line 111: If you link to #90 here, it becomes obsolete with merging #90, so #90 would need to touch this file, which is completly unrelated to its content, wouldn't link to specific stuff. If you want to do advertising for new features, I'd keep it more general

connaisseur/sigstore_validator.py
xopham: I sort of actively decided that the check that there is at least 1 valid digest is handled centrally later to also avoid code duplication. Would keep it, as there seems to be no disadvantage in doing later centrally, or is there?

connaisseur/sigstore_validator.py
xopham: Well, I assumed from the cosing code that there might be non-json lines, e.g. when there is a certificate: https://github.com/sigstore/cosign/blob/v0.2.0/cmd/cosign/cli/verify.go#L144
However, it might be a valid question whether we should or not (and just log) raise in the other cases, given that one unexpected json does not mean there is no valid json line. What do you think @annekebr ?

connaisseur/sigstore_validator.py
xopham: I would absolutely not assume that it may take longer than a minute. Should typically be done in a few seconds. The timeout is however set to be vastly above expected timing because we want to make sure it only ever happens when there is a real timeout. For it to take longer than 2 minutes, would require an impossibly bad internet connection which should avoid that we ever run into this by accident.
I have never encountered the timeout condition. It is more meant as a safety precaution...

setup/COSIGN.md
xopham: Actually, I thought we would specifically in that case not have to touch this file, since if you open the closed PR, you know that it has been done. Can also remove it if it's unwanted...

Starkteetje: The formulation with `will also be introducing` would still become wrong, I'd remove it, but also don't feel strongly about it, just noticed

connaisseur/sigstore_validator.py
peterthomassen: I think it should be a small multiple your expectation's upper bound. If that's a few seconds, 60s should certainly be enough.

setup/COSIGN.md
peterthomassen: I agree with @Starkteetje

connaisseur/crypto.py
```diff
@@ -10,11 +12,26 @@ def verify_signature(public_base64: str, signature_base64: str, message: str):
...
+
+
+def decode_and_verify_ecdsa_key(public_base64: str):
+    """
+    Verifies that the provided public key in base64 encoding qualifies as a
+    proper ecdsa key and throws if not.
+    """
+    try:
+        public = base64.b64decode(public_base64)
+        pubkey = ecdsa.VerifyingKey.from_der(public)
```
peterthomassen, line 31:
```suggestion
        return ecdsa.VerifyingKey.from_der(public)
```

Remove line 37.

connaisseur/sigstore_validator.py
peterthomassen, line 52: This `except` clause should be right after `json.loads` (line 34), right? The (currently) inner `try`-`except` block could then go up one level, streamlining the whole thing.

peterthomassen (dismissed):
lgtm, some suggestions

connaisseur/sigstore_validator.py
Starkteetje: If you do it here, you might get a more detailed error. If `stdout.splitlines()` is empty here but the return code is 0, then there was an error with cosign, which is different from there being no valid digest. If you assume that's not gonna happen, you can also leave it out, but I'd say that would be the more defensive approach of calling another binary (or function in general)

connaisseur/sigstore_validator.py
annekebr: True, we shouldn't be raising if there is a signature that can be parsed. I'd go for logging and raising only if `digests` is the empty list in the end. If that is the case there is definitely something odd/new with the cosign output because cosign indicated that it was able to verify the image against the public key but we cannot handle the output correctly.

annekebr: Ah, that also ties into @Starkteetje s comment [above](https://github.com/sse-secure-systems/connaisseur/pull/107/files#r611461294) as we basically reduce everything to the case `digests` is the empty list and would throw a specific error in this case only.

setup/COSIGN.md
xopham: will drop it

connaisseur/sigstore_validator.py
xopham: will adjust to 60s

connaisseur/crypto.py
xopham: will do

connaisseur/sigstore_validator.py
annekebr: I just tried but noticed that we need `sig_data` in the inner block which is not defined if a `JSONDecodeError` is thrown (and caught by us).

annekebr: So, as far as I see it we have these two options: 
1) If cosign exits with status code one, thus indicating successful signature verification against the given public key, and Connaisseur can find at least one line in the output that is a) valid json and b) this valid json is in the format known to Connaisseur such that it can parse the digest we mutate the admission request to point to the digest we found. We fail with UnexpectedCosignData if cosign returns with 0 but Connaisseur cannot find a single parsable digest in the output.
This would the like in https://github.com/sse-secure-systems/connaisseur/pull/136
2) If cosign exits with status code one, thus indicating successful signature verification against the given public key, but Connaisseur finds at least one line that it cannot parse, we fail with UnexpectedCosignData, thus not admitting the image because we don't trust the cosign output any longer.

@xopham favors option 1. Any other opinions @Starkteetje , @peterthomassen ? 
If option 1 is fine, please approve the above mentioned PR. :)

Starkteetje: Cosign exits with status code zero, right?

I'd be very strongly in favor of option 2 if there's no reason to expect non JSON data. Trying to accept partial, wrongly formatted data as valid is a great recipe for disaster. If cosign should give us JSON but doesn't give us JSON, either it is broken, there's an attack or we're not compatible. In each of these cases, we should fail

xopham: Well, we are currently using Cosign v0.2.0. It will only output all valid signatures for a given key in a separate line. However, it seems to be possible that a certificate common name is returned that would probably not be valid json, as mentioned above: https://github.com/sigstore/cosign/blob/v0.2.0/cmd/cosign/cli/verify.go#L144 . Furthermore, it is not prohibited to sign multiple times with a valid public key and one might also sign with the same public key in a future version of cosign that could potentially have a different output format, but might theoretically still pass cosign. I do not think this is a strong case and am not sure if it exists.
However, if there is a valid signature line that fullfils our spec, I feel we are doing exactly what was expected, right?

Starkteetje: Ah, sorry. I didn't see the link before. I guess that does indeed warrant assuming there might be other stuff in the output. It kind of makes my point still, since e.g. now, we have to trust that cosign doesn't allow linebreaks in `vp.Cert.Subject.CommonName`, which is probably a fair assumption in this instance. Still, I don't like it.
But tbh I also don't see how we could work around that, so I guess, I'm fine with this going forward as is

annekebr: Went for option 2 to be on the safe side.

We probably need to touch this anyway as soon as something changes in the output format; so we probably don't gain much by option in the first place. 

Changed #136

annekebr (approved):
Great work, @xopham !

spielkind:
hey, great work, I'm just wondering if it would be possible to verify images signed by different keys (one for each vendor/project group), or if this is not possible with the architecture of connaisseur?

xopham:
Hi @spielkind :wave: Yes! We are currently working on adding support for multiple keys to allow more granular control. I was actually just playing around with it's first version just now. I made a brief description of the plans in the discussions: https://github.com/sse-secure-systems/connaisseur/discussions/151

PR #108, opened 2021-04-01
ci: codecov fix

phbelitz:
The code coverage silently fails, since the python:slim image doesn't contain git and curl for it to function. This is fixed.


Starkteetje:
Maybe we should also be running the codecov action with `fail_ci_if_error: true # optional (default = false)` such that it will visibly fail in the future

xopham (approved):
lgtm

xopham (approved):

PR #109, opened 2021-04-01
ci/test: Test Connaisseur verification with target namespace

### LINKED ISSUE #103: Integration test for validation in single namespace
We got #100 for enabling verification in namespaces (in contrast to for all namespaces). Functionality was manually tested, but we should add an integration test for that

Starkteetje:
This commit adds an integration test for the namespaced verification behaviour of Connaisseur.

Fix #103

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -54,3 +54,34 @@ jobs:
 
       - name: Run actual integration test
         run: bash connaisseur/tests/integration/integration-test.sh
+
+  namespaced-integration-test:
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/download-artifact@v2
+        with:
+          name: images
+      - name: Load Docker images
+        run: |
+          docker load -i ${GITHUB_SHA}_image.tar
+          docker load -i ${GITHUB_SHA}_hook.tar
+      - name: Install yq and bash
+        run: |
+          sudo snap install yq
+          sudo apt update
+          sudo apt install bash -y
+      - name: Create KinD cluster
+        run: |
+          GO111MODULE="on" go get sigs.k8s.io/kind
+          kind create cluster --wait 120s
+      - name: Check KinD cluster
+        run: kubectl get nodes
+      - name: Add images to KinD
+        run: |
+          kind load docker-image $(yq e '.deployment.image' helm/values.yaml)
+          kind load docker-image $(yq e '.deployment.helmHookImage' helm/values.yaml)
+
+      - name: Run actual integration test
+        run: bash connaisseur/tests/integration/namespaced-integration-test.sh
--- a/connaisseur/tests/integration/namespaced-integration-test.sh
+++ b/connaisseur/tests/integration/namespaced-integration-test.sh
@@ -1,0 +1,52 @@
+#! /bin/bash
+set -euo pipefail
+
+# This script is expected to be called from the root folder of Connaisseur
+
+echo 'Preparing Connaisseur config...'
+yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' helm/values.yaml connaisseur/tests/integration/namespaced-update.yaml
+echo 'Config set'
+
+echo 'Installing Connaisseur...'
+make install || { echo 'Failed to install Connaisseur'; exit 1; }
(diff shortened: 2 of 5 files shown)


phbelitz:
Doesn't it suffice to only do the namespaced integration test at this point? I feel like we are doing most of the testing here twice without any more value.

.github/workflows/release.yaml
```diff
@@ -54,3 +54,34 @@ jobs:
...
+          kind create cluster --wait 120s
+      - name: Check KinD cluster
+        run: kubectl get nodes
+      - name: Add images to KinD
+        run: |
+          kind load docker-image $(yq e '.deployment.image' helm/values.yaml)
+          kind load docker-image $(yq e '.deployment.helmHookImage' helm/values.yaml)
+
+      - name: Run actual integration test
+        run: bash connaisseur/tests/integration/namespaced-integration-test.sh
```
xopham, line 87: can we avoid the code duplication here for github actions?

connaisseur/tests/integration/namespaced-integration-test.sh
```diff
@@ -0,0 +1,63 @@
...
+
+echo 'Preparing Connaisseur config...'
+yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' helm/values.yaml connaisseur/tests/integration/namespaced-update.yaml
+echo 'Config set'
+
+echo 'Installing Connaisseur...'
+make install || { echo 'Failed to install Connaisseur'; exit 1; }
+echo 'Successfully installed Connaisseur'
+
+echo 'Testing unsigned image...'
```
xopham, line 14: description generic

connaisseur/tests/integration/namespaced-integration-test.sh
```diff
@@ -0,0 +1,63 @@
...
+
+if [[ "$(cat output.log)" != 'Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: could not find signed digest for image "docker.io/securesystemsengineering/testimage:unsigned" in trust data.' ]]; then
+  echo 'Failed to deny unsigned image or failed with unexpected error. Output:'
+  cat output.log
+  exit 1
+else
+  echo 'Successfully denied usage of unsigned image'
+fi
+
+echo 'Testing image signed under different key...'
```
xopham, line 25: would drop this case, as we check that images signed under a different key are rejected in the central integration tests

xopham (commented):
I like it. Would prefer if we briefly agreed on the test matrix we want to have here.
My proposal would be:

- signed image in specified NS
- unsigned image in specified NS
- signed image in other NS
- unsigned image in other NS

If we want this to be hollistic, we could also go for:
- signed under different key in NS
- signed under different key in other NS

I do not think that the actual testing of images is lengthy. The time consumption here is due to building and installing conny...

Starkteetje:
> Doesn't it suffice to only do the namespaced integration test at this point? I feel like we are doing most of the testing here twice without any more value.

I would say no, as we they're different setup cases and we hence test whether in the base case it just works and in the specified case it checks that checking is done in the respective namespaces. Would hence agree with @xopham that we e.g. do not need to test for different keys as that's a feature test done in the base case, but would check (un)signed in controlled and unsigned in uncontrolled as a very basic verification for "Connaisseur works as intended with this configuration"

> My proposal would be:
>  -  signed image in specified NS
>  -  unsigned image in specified NS
>  -  signed image in other NS
>  -  unsigned image in other NS

would leave out signed in other ns as if unsigned is allowed that should be sufficient to test that Connaisseur doesn't do validation in the namespace

.github/workflows/release.yaml
Starkteetje: There's [a workflow trigger thingy for Actions](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run), but that does not allow to let Workflow A spawn B as a child, but rather if A completes, B can be triggered. In practice I would assume that this leads to the release flow showing a checkmark once it triggered the integration tests and then not showing X even though they failed later-on, which I'd say is really bad for visibility of the results

xopham: I see, then let's make an issue about this and handle it when we got the time and think it is a good thing to pick up

phbelitz (dismissed):

phbelitz (approved):

PR #110, opened 2021-04-07
fix: fix error during build stage

annekebr:
In cases where there is no kubecontext set (automatic builds) all make commands fail as the cluster variable is defined globally. In any case, no make command should fail in case the cluster variable cannot be discovered. This PR fixes this issue.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,7 @@
 NAMESPACE = connaisseur
 IMAGE := $(shell yq e '.deployment.image' helm/values.yaml)
 HELM_HOOK_IMAGE := $(shell yq e '.deployment.helmHookImage' helm/values.yaml)
+CLUSTER := $(shell CONTEXT=`kubectl config current-context` && kubectl config view -ojson | jq --arg CONTEXT $$CONTEXT '.contexts[] | select(.name==$$CONTEXT) | .context.cluster')
 
 .PHONY: all docker certs install unistall upgrade annihilate
 
@@ -23,7 +24,7 @@ install: certs
 	#
 	#=============================================
 	#
-	helm install connaisseur helm --atomic
+	helm install connaisseur helm --atomic --set alerting.cluster=$(CLUSTER)
 
 uninstall:
 	kubectl config set-context --current --namespace $(NAMESPACE)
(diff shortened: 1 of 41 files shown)


Makefile
```diff
@@ -1,7 +1,7 @@
 NAMESPACE = connaisseur
 IMAGE := $(shell yq e '.deployment.image' helm/values.yaml)
 HELM_HOOK_IMAGE := $(shell yq e '.deployment.helmHookImage' helm/values.yaml)
-CLUSTER := $(shell CONTEXT=`kubectl config current-context` && kubectl config view -ojson | jq --arg CONTEXT $$CONTEXT '.contexts[] | select(.name==$$CONTEXT) | .context.cluster')
+CLUSTER := $(shell CONTEXT=`kubectl config current-context || true` && if [[ -z ${CONTEXT} ]];do kubectl config view -ojson | jq --arg CONTEXT $$CONTEXT '.contexts[] | select(.name==$$CONTEXT) | .context.cluster';done;else echo "cluster not discoverable";fi)
```
Starkteetje, line 4: Shouldnt this be `! -z` or `-n`?

Makefile
```diff
@@ -1,7 +1,7 @@
 NAMESPACE = connaisseur
 IMAGE := $(shell yq e '.deployment.image' helm/values.yaml)
 HELM_HOOK_IMAGE := $(shell yq e '.deployment.helmHookImage' helm/values.yaml)
-CLUSTER := $(shell CONTEXT=`kubectl config current-context` && kubectl config view -ojson | jq --arg CONTEXT $$CONTEXT '.contexts[] | select(.name==$$CONTEXT) | .context.cluster')
+CLUSTER := $(shell CONTEXT=`kubectl config current-context || true` && if [[ -z ${CONTEXT} ]];then echo "cluster not discoverable";else kubectl config view -ojson | jq --arg CONTEXT $$CONTEXT '.contexts[] | select(.name==$$CONTEXT) | .context.cluster';fi)
```
Starkteetje, line 4: From looking at it, this would result in `$CLUSTER = "cluster not discoverable"`, right? Is that intended or should it be echo "bla" >&2 (dunno if that works)

PR #111, opened 2021-04-07
fix: ensure backwards compatibility when upgrading to 1.5.0

annekebr:
Upgrading to 1.5.0 without pulling the newest version of the helm templates will result in a Configuration Error as there are configuration files missing that are needed for the alerting which came with 1.5.0. This fix ensures that one can update to the new image without failures even without repulling the connaisseur repo. The added feature will, however, not be functional without the new helm templates.

### DIFF
--- a/connaisseur/alert.py
+++ b/connaisseur/alert.py
@@ -110,7 +110,7 @@ class Alert:
 
 def load_config():
     try:
-        alert_config_dir = f'{os.getenv("ALERT_CONFIG_DIR")}'
+        alert_config_dir = f'{os.getenv("ALERT_CONFIG_DIR", "/app")}'
         with safe_path_func(
             open, alert_config_dir, f"{alert_config_dir}/alertconfig.json", "r"
         ) as configfile:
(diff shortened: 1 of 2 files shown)


connaisseur/alert.py
```diff
@@ -110,20 +110,23 @@ def _get_headers(receiver_config):
...
         json_validate(instance=alertconfig, schema=schema)
     except Exception as err:
-        raise ConfigurationError(
-            "Alerting configuration file either not present or not valid."
-            "Check in the 'helm/values.yml' whether everything is correctly configured. {}".format(
-                str(err)
-            )
-        ) from err
+        if isinstance(err, FileNotFoundError):
+            return {}
```
Starkteetje, line 122: I guess this should be logged, probably info/warning?

annekebr: yeah, I thought about it too - what would you log in that  case? That you have updated but without the new setup? Or just that it's missing - but that might be confusing bc this PR is to ensure everything works just as well with new new image.

Starkteetje: `Info: No alerting configuration file found. No alerts will be issued` shouldn't be that confusing I'd hope 🤷

annekebr: Hmya, but if you built it with the new templates there might be a alerting configuration file but still no alerts will be issued :D

annekebr: went for a mixture ;-)

phbelitz (approved):

PR #113, opened 2021-04-08
Experimental dockerfile and false in main image

Starkteetje:


Makefile
```diff
@@ -9,8 +9,9 @@ COSIGN_VERSION = 0.2.0
 all: docker install
 
 docker:
-	docker build --build-arg COSIGN_VERSION=$(COSIGN_VERSION) -f docker/Dockerfile -t $(IMAGE) .
```
xopham, line 12: you could use a `--build-arg` here to distinguish `experimental` and based on that skip integration of experimental features. Avoids code duplication, but makes the Dockerfile messy

Starkteetje:
Decided against experimental image for now

PR #115, opened 2021-04-08
hotfix: fixed image format validation

### LINKED ISSUE #114: kubernetes - not a valid image format
I have deployed notary-server and signer to my kubernetes cluster, made sure all TLS is setup properly, I then deployed connaisseur, and again, all green in kubernetes.

I have signed some images in my private registry also deployed on kubernetes, but when I try to use them in kube objects I got the error: 
```
Error from server: error when creating "failed.yml": admission webhook "connaisseur-svc.connaisseur.svc" denied the request: "private-registry:30005/kibana:5.6.8" is not a valid image format.

```

The relevant logs from connaisseur pod:

```
ERROR: {'message': '"private-registry:30005/kibana:5.6.8" is not a valid image format.', 'context': {'user': 'kubernetes-admin', 'operation': 'CREATE', 'kind': 'Job', 'namespace': 'default', 'name': 'hello', 'image': 'private-registry:30005/kibana:5.6.8'}, 'detection_mode': False}
```

Is this a known bug? Is there something specific I need?

phbelitz:
The image validation assumed that the registry/domain name of an image contains a dot (.) as domain usually do. However this doesn't work for images coming from private registries like 'private-registry:3000'. Making the dot optional also doesn't fix the problem since for image like 'path/to/image:tag', 'path' would be recognized as a domain name. Thus the validation has been change so that domains that do not contain a dot, necessarily need to specify a port in order for them to be recognized as domains.

fix #114


peterthomassen (approved):
lgtm.

Cherry on the cake: Test that when there's no dot, a port specification is required.

connaisseur/tests/test_image.py
```diff
@@ -57,6 +57,7 @@ def im():
             "path",
             "sub.registry.io",
         ),
+        ("registry:3000/image:tag", "image", "tag", None, "", "registry:3000"),
```
annekebr, line 60: maybe a test case for the error case (`registry/image:tag`)?

phbelitz: should be included here:
https://github.com/sse-secure-systems/connaisseur/blob/aee751bcd87232c2f92f4b6a1178799ccee357c9/connaisseur/tests/test_image.py#L42

peterthomassen:
> Cherry on the cake: Test that when there's no dot, a port specification is required.

I take that back; that's a different case for which there are already tests.

annekebr (approved):

PR #121, opened 2021-04-12
fix: Remove development files from image

Starkteetje:
Previously, we erronously built with code coverage files and the python cache if it was present on the developer's machine


xopham (approved):
lgtm

phbelitz (approved):
QL

Issue #118, opened 2021-04-09
Consider setting a codecov yaml to e.g. enforce minimum coverage...


xopham:
**Describe the feature**
A minimum code coverage via codecov yaml could be enforced: https://docs.codecov.io/docs/codecov-yaml

Starkteetje:
At least the base implementation of it needs the codecov app, which we don't want to use given the recent breach; we could also write our own interpreter of the test coverage file, but that seems to be too much work for not enough gain atm

PR #122, opened 2021-04-12
ci: Define test coverage targets

Starkteetje:


Starkteetje:
@xopham The problem of #118 to enforce coverage is not doable with just the github actions if we don't parse the coverage report, since checking the targets is not part of the action. We'd get a workflow job if we included the [Codecov Github App](https://github.com/apps/codecov) into the repository. Otherwise we only get the PR comments (haven't found out why they don't work sometimes for us), which worked for all my tests when I added the app the the repo and had OAuth with codecov for my own account enabled

xopham:
> @xopham The problem of #118 to enforce coverage is not doable with just the github actions if we don't parse the coverage report, since checking the targets is not part of the action. We'd get a workflow job if we included the [Codecov Github App](https://github.com/apps/codecov) into the repository. Otherwise we only get the PR comments (haven't found out why they don't work sometimes for us; they worked for all my tests when I added the app the the repo and had OAuth with codecov for my own account enabled)

Starkteetje:
Given current breach of codecov, we don't feel comfy including the app with permission on the repo

PR #136, opened 2021-04-12
fix: make error handling more specific

annekebr:
If the `cosign` binary indicates successful successful signature verification by exiting with status code `0` but Connaisseur is not able to retrieve at least one valid image digest from the output from the `cosign` binary, the error message should show that this is a parsing issue. Right now, we would have thrown an error indicating that signature verification has failed which is misleading because the actual verification has already happened on the `cosign` side at this point.

### DIFF
--- a/connaisseur/sigstore_validator.py
+++ b/connaisseur/sigstore_validator.py
@@ -36,19 +36,18 @@ def get_cosign_validated_digests(image: str, pubkey: str):
                     digest = sig_data["Critical"]["Image"].get(
                         "Docker-manifest-digest", ""
                     )
-                    if re.match(r"(sha256:)?[0-9A-Fa-f]{64}", digest) is None:
+                    if re.match(r"sha256:[0-9A-Fa-f]{64}", digest) is None:
                         raise Exception(
-                            "digest does not match expected digest pattern."
+                            f"digest '{digest}' does not match expected digest pattern."
                         )
                 except Exception as err:
                     raise UnexpectedCosignData(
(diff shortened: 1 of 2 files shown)


connaisseur/sigstore_validator.py
```diff
@@ -36,19 +36,28 @@ def get_cosign_validated_digests(image: str, pubkey: str):
                     digest = sig_data["Critical"]["Image"].get(
                         "Docker-manifest-digest", ""
                     )
+                    # check whether the digest matches "sha256:<hex representation of a sha256 hash>" or
+                    # just plain "<hex representation of a sha256 hash>"
                     if re.match(r"(sha256:)?[0-9A-Fa-f]{64}", digest) is None:
-                        raise Exception(
-                            "digest does not match expected digest pattern."
+                        logging.info(
```
Starkteetje, line 42: Why do we no longer raise here? If we got to this block, we have a valid json with the structure we expect. If then the digest is wrongly formatted, something is wrong and we should fail or is there a valid reason for the digest to be mis-formatted?

And if there is, then the subject of the PR and commit are rather "make parsing cosign more lenient and error handling more specific" or sth similar

connaisseur/tests/test_sigstore_validator.py
```diff
@@ -1,6 +1,7 @@
 import pytest
 import pytest_subprocess
 import subprocess
+import json
```
xopham, line 4: you are not using the 'json' module, are you?

xopham (approved):
looks good to me

PR #138, opened 2021-04-13
refactor: use init containers to wait for webhook

annekebr:
Currently, we have a Kubernetes job and lot of boilerplate to manage the webhook lifecycle. We can simplify this a lot by using an initContainer to wait for the webhook before deploying the actual connaisseur containers.


annekebr:
king @phbelitz found better solution

PR #144, opened 2021-04-15
docs/ci/test: delegations feature

Starkteetje:
- Document delegation feature
- Add test for delegation feature
- Add test for multiple defined digests due to delegations


connaisseur/tests/integration/integration-test.sh
```diff
@@ -51,6 +51,51 @@ else
   echo 'Successfully allowed usage of signed image'
 fi
 
+echo 'Testing edge case of tag defined in both targets and release json file...'
```
annekebr, line 54: This test case means essentially two digests for one tag, correct? Can that only happen with these two files?

annekebr (commented):
Nice PR!

connaisseur/tests/integration/integration-test.sh
Starkteetje: Tag can be e.g. signed by different people, but it will not show up in the current content. But if you push stuff before setting up delegations you can end up with multiplte "current" digests for a tag. See e.g. `docker trust inspect securesystemsengineering/testimage --pretty` and you'll see that there are two different digests for `signed`. For `double_sig` there's only one shown even though both exist (as is tested by the integration test introduced in one of the commits). @phbelitz is better at explaining how and why that happens. We just noticed again that it did happen and I decided this would be a good time to introduce the test that Connaisseur actually chooses the correct image in that case

phbelitz: @annekebr yea the explaination goes like this:

let's say you don't have delegations set up. then your signed digest will pop up in the `targets.json`, as it should. after you signed some images this way you come up with the great idea to introduce delegations. sure thing. now if you sign your images with the newly introduced delegation keys, your signed digests will pop up in the `targets/releases.json`, under the same tag which is already stored in the `targets.json`. now you have two different digests (an old one in the `targets.json` and a new one in the `targets/releases.json`) for the same tag. this test essentially should make sure the "right" one is used, so the new one in the `targets/releases.json`, since this is now the source of truth with delegations set up.

funnily enough from here there is yet another edge case! so when you already signed images without delegations, meaning their digest pop up in the `targets.json`, and then introduce delegation WITHOUT signing with them yet, you get a funny situation. since delegations are set up, you'd expect the "right" digest to be in the `targets/releases.json`, like in the case from above, but nope! you haven't used the delegation keys yet to sign any digests, so only the `targets.json` will actually contain any digests.

connaisseur/tests/integration/integration-test.sh
```diff
@@ -51,6 +51,51 @@ else
...
+if [[ "$(cat output.log)" != 'pod/pod2 created' ]]; then
+  echo 'Failed to allow image signed by designated signer. Output:'
+  cat output.log
+  exit 1
+else
+  echo 'Successfully allowed usage of image signed by designated signer'
+fi
+
+echo 'Testing image with missing designated signer...'
+kubectl run pod3 --image=securesystemsengineering/testimage:non_existant >output.log 2>&1 || true
```
phbelitz, line 76: Instead of testing whether a delegation hasn't signed a tag that doesn't exist in the first place, we should test if that delegation hasn't signed a tag that actually exists, right? so ideally you should add your signature to `:non_existant` so that a signature theoretically exists, but since i'm set up as required delegation, this will fail.

phbelitz (changes requested):

connaisseur/tests/integration/integration-test.sh
Starkteetje: Added it as `:wrong_signer` with a sig of mine

phbelitz (approved):
Bueno

PR #145, opened 2021-04-15
refactor: refactor workflows to reduce code duplication

### LINKED ISSUE #112: Reduce code repetition in Github Actions
**Describe the feature**
We have a lot of code duplication in Github Actions, especially with regards to the integration tests, that are run for both CICD and Release workflows and where both namespaced and "basic" integration test are very similar to each other

Github has the [workflow_run](https://docs.github.com/en/actions/reference/events-that-trigger-workflows#workflow_run) option that would allow triggering an integration test workflow from both CICD and Release, but probably the workflow result would not be shown as part of the workflow, i.e. one has CICD running due to a PR, CICD triggers integration tests, then shows green checkmark regardless of whether integration test workflow succeeded or failed. <- All that's only theory, I didn't test it

annekebr:
As we added more flavours of integration testing, we introduced some code duplication. This aims to remove redundance by using a private github action.

Fixes #112

### DIFF
--- a/.github/actions/integration-test/action.yaml
+++ b/.github/actions/integration-test/action.yaml
@@ -1,0 +1,49 @@
+name: integration-test
+description: 'action that prepares the integration test setup and runs the provided integration test script'
+inputs:
+  integration-test-script:
+    description: 'file name of integration test script'
+    required: true
+runs:
+  using: "composite"
+  steps:
+    - name: Load Docker images
+      run: |
+        docker load -i ${GITHUB_SHA}_image.tar
+        docker load -i ${GITHUB_SHA}_hook.tar
+      shell: bash
+    - name: Install yq and bash
+      run: |
+        sudo snap install yq
+        sudo apt update
+        sudo apt install bash -y
+      shell: bash
+    - name: Create KinD cluster
+      run: |
+        GO111MODULE="on" go get sigs.k8s.io/kind
+        kind create cluster --wait 120s
+      shell: bash
+    - name: Check KinD cluster
+      run: kubectl get nodes
+      shell: bash
+    - name: Add images to KinD
+      run: |
+        kind load docker-image $(yq e '.deployment.image' helm/values.yaml)
+        kind load docker-image $(yq e '.deployment.helmHookImage' helm/values.yaml)
+      shell: bash
+    - name: Set environment variables for alerting listener
+      run: |
+        SERVICE_CONTAINER_ID=\
+        $(docker container ls --no-trunc --format "{{json . }}" | \
+        jq ' . | select(.Image|match("alerting-endpoint"))' | jq -r .ID)
+        docker network connect kind ${SERVICE_CONTAINER_ID}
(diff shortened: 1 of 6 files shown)


annekebr:
As one cannot use `services` on an `if` condition, the alerting interface is spawned on all integration test cases even when it is not needed. One could prevent that by lowering the entries in the build matrix to the two scripts that don't need the service container and readd a separate code block for the job with service container.

However, the integration test script that needs the service container lasts the longest anyway, so time-wise I don't think it makes a difference. Don't feel strongly about it, but wanted to point out.

.github/actions/integration-test/action.yaml
```diff
@@ -0,0 +1,45 @@
+name: integration-test
+description: 'action that prepares the integration test setup and runs the provided integration test script'
+inputs:
+  integration-test-script:
+    description: 'file name of integration test script'
+    required: true
+    default: 'integration-test.sh'
```
Starkteetje, line 7: Does one need to set a default here? Otherwise I'd just leave the default out

Looks really nice, overall +1

.github/actions/integration-test/action.yaml
```diff
@@ -0,0 +1,45 @@
...
+      run: kubectl get nodes
+      shell: bash
+    - name: Add images to KinD
+      run: |
+        kind load docker-image $(yq e '.deployment.image' helm/values.yaml)
+        kind load docker-image $(yq e '.deployment.helmHookImage' helm/values.yaml)
+      shell: bash
+    - name: Set environment variables for alerting listener
+      run: |
+        SERVICE_CONTAINER_ID=$(docker container ls --no-trunc --format "{{json . }}" | jq ' . | select(.Image|match("alerting-endpoint"))' | jq -r .ID)
```
xopham, line 36: think it might make sense to linebreak here for better readability: https://stackoverflow.com/questions/59954185/github-action-split-long-command-into-multiple-lines
specifically if it is `bash` scripting

.github/actions/integration-test/action.yaml
```diff
@@ -0,0 +1,45 @@
...
+        docker network connect kind ${SERVICE_CONTAINER_ID}
+        ALERTING_ENDPOINT_IP=$(docker network inspect kind | jq -r --arg container_id ${SERVICE_CONTAINER_ID}  '.[].Containers[$container_id].IPv4Address' | sed s+/.*++g)
```
xopham, line 38: same as above

xopham (commented):
real awesome, modular refactoring. Please consider using `refactor` for the commit common type: https://github.com/sse-secure-systems/connaisseur/blob/master/CONTRIBUTING.md#semantic-and-conventional-commits

.github/actions/integration-test/action.yaml
annekebr: removed that line :-) Thanks!

xopham (approved):
awesome solution

PR #147, opened 2021-04-20
refactor: determine cluster name

annekebr:


helm/templates/env.yaml
```diff
@@ -38,9 +38,13 @@ data:
   IS_COSIGN: "1"
   {{- end}}
   ALERT_CONFIG_DIR: "/app/config"
+  {{- if .Values.alerting}}
```
xopham, line 41: I have seen this thread that seems to indicated that there is no way around nested if-clauses this: https://stackoverflow.com/questions/59795596/helm-optional-nested-variables

however, this seems to indicated, we could use a `default`: https://helm.sh/docs/chart_best_practices/values/#flat-or-nested-values
Something like the following seems to work or at least not fail in kube-linter:
```
{{- if .Values.alerting}}
CLUSTER_NAME: {{default "not specified" .Values.alerting.cluster }}
{{- end }}
```
Not absolutely sure why it fully covers the issue, since it still not set if there is no `alerting` key, but it seems to and from a practical PoV you might be catching it in the makefile...thoughts?

xopham (commented):
Really like it. Think it will tick the box.
I have only 1 minor comment and one up for discussion.

phbelitz (approved):
:+1:

PR #156, opened 2021-05-03
Feat/modular

phbelitz:
With the upcoming of more validation methods, connaisseur shouldn't be restricted to only support a single one. That's why the validation has been modularized to not only support multiple validation methods, but also to easily add new ones.


connaisseur/__main__.py
```diff
@@ -29,5 +29,5 @@
     # container
     APP.run(
         host="0.0.0.0",  # nosec
-        ssl_context=("/etc/certs/tls.crt", "/etc/certs/tls.key"),
+        ssl_context=("/app/certs/tls.crt", "/app/certs/tls.key"),
```
peterthomassen, line 32: I doubt this change is necessary for adding modular validation. Still, it is in this commit. I really *urge* everyone to keep changes minimal, and at least keep commit size minimal. If you want to change some directory structure, that should go into a separate commit.

Commits should not exceed let's say 200 lines.

peterthomassen: It's really hard to review commits like these, when you have to keep lots of different aspects in mind. It's unnecessarily hard.

peterthomassen: Ideally, this PR would start with a refactor commit that does *nothing* but modularizing validation under the hood. For this commit, you would, ideally, not have to touch any test cases.

Then, another commit that introduces the necessary syntax changes in the configuration. This of course would need to touch tests.

Then, another one that adds new validators.

helm/values.yaml
```diff
@@ -21,11 +21,18 @@ service:
   port: 443
 
 # configure access to the notary server
-notaries:
-# public notary from dockerhub
-- name: dockerhub
+validators:
+# public notary form dockerhub
```
peterthomassen, line 25: comment in wrong line?

connaisseur/config.py
```diff
@@ -27,55 +31,72 @@ def __init__(self):
...
+        config = self.__merge_configs(config_content, secrets_config_content)
+
+        self.__validate(config)
+
+        self.validators = [Validator(**validator) for validator in config]
+
+    def __merge_configs(self, config: dict, secrets_config: dict):
+        if not config:
             msg = "Error loading connaisseur config file."
             raise NotFoundException(message=msg)
```
peterthomassen, line 46: This error message belongs into line 32.

connaisseur/config.py
```diff
@@ -27,55 +31,72 @@ def __init__(self):
...
 
-        self.__validate(config_content)
-
-        self.notaries = [Notary(**notary) for notary in config_content]
+        for validator in config:
+            validator.update(secrets_config.get(validator.get("name"), {}))
+            auth_path = f'{self.__EXTERNAL_PATH}{validator["name"]}/auth.yaml'
```
peterthomassen, line 50: The validation hasn't run yet. It may be safe -- but are we sure this is safe?

connaisseur/config.py
```diff
@@ -27,55 +31,72 @@ def __init__(self):
...
             config,
             self.__SCHEMA_PATH,
             "Connaisseur configuration",
             InvalidConfigurationFormatError,
         )
-        notary_names = [notary.get("name") for notary in config]
-        if collections.Counter(notary_names)["default"] > 1:
-            msg = "Too many default notary configurations."
+        validator_names = [validator.get("name") for validator in config]
+        if collections.Counter(validator_names)["default"] > 1:
```
peterthomassen, line 67: nice

connaisseur/config.py
```diff
@@ -27,55 +31,72 @@ def __init__(self):
...
-                )
-
-        self.notaries = [Notary(**notary) for notary in config]
+        # for notary in config:
+        #     key_names = [key.get("name") for key in notary.get("pub_root_keys")]
+        #     if collections.Counter(key_names)["default"] > 1:
+        #         msg = "Too many default keys in notary configuration {notary_name}."
+        #         raise InvalidConfigurationFormatError(
+        #             message=msg, notary_name=notary.get("name")
+        #         )
```
peterthomassen, line 77: was this forgotten?

connaisseur/config.py
```diff
@@ -27,55 +31,72 @@ def __init__(self):
...
+        `validator_name` is None, the element with `name=default` is taken, or the only
+        existing element.
 
         Raises `NotFoundException` if no matching or default element can be found.
         """
         try:
-            if len(self.notaries) < 2:
-                return next(iter(self.notaries))
+            if len(self.validators) < 2:
+                return next(iter(self.validators))
```
peterthomassen, line 89: Did we not agree that removing a validator (or key) should not lead to another validator (or key) to turn effective implicitly? In other words, shouldn't we remove these two lines, and require the only validator's name to be `default` (or explicitly named in the policy)?

connaisseur/config.py
```diff
@@ -27,55 +31,72 @@ def __init__(self):
...
 
-            notary_name = notary_name or "default"
+            validator_name = validator_name or "default"
             return next(
-                notary for notary in self.notaries if notary.name == notary_name
+                validator
+                for validator in self.validators
+                if validator.name == validator_name
```
peterthomassen, line 95: It seems to be that validator name uniqueness is not enforced, right? If so, this would return the first one with the given name. I think we should require uniqueness at validation time.

Other approach:
```python
return list(filter(lambda v: v.name == (validator_name or "default"), self.validators))[0]
```

connaisseur/config.py
```diff
@@ -27,55 +31,72 @@ def __init__(self):
...
             )
         except StopIteration as err:
-            if len(self.notaries) < 2:
-                msg = "No notary configurations could be found."
+            if len(self.validators) < 2:
+                msg = "No validator configurations could be found."
```
peterthomassen, line 99: Why does `< 2` mean that none were found? It includes the case of one validator, doesn't it?

connaisseur/flask_server.py
```diff
@@ -138,8 +137,14 @@ def __create_logging_msg(msg: str, **kwargs):
     return str({"message": msg, "context": dict(**kwargs)})
 
 
-def __create_json_patch(path: str, index: int, image: Image):
-    return {"op": "replace", "path": path.format(str(index)), "value": str(image)}
+def __update_json_patch(
+    patches: dict, path: str, index: int, image: Image, digest: str
+):
+    if digest:
+        image.set_digest(digest)
```
peterthomassen, line 144: Why this side effect on `image`? (Previously, this was done outside the function.)

connaisseur/validators/__init__.py
```diff
@@ -0,0 +1 @@
+pass
```
peterthomassen, line 1: Is this necessary?

connaisseur/validators/cosign/__init__.py
```diff
@@ -0,0 +1 @@
+pass
```
peterthomassen, line 1: Is this necessary?

connaisseur/validators/interface.py
```diff
@@ -8,10 +9,17 @@ def __init__(self, name: str, **kwargs):
...
         """
         Validates an admission request, using the extra arguments from the image policy.
 
         Returns a list of trusted digests.
         """
         raise NotImplementedError
+
+    @property
+    def healthy(self):
+        return True
```
peterthomassen, line 22: This should not be default return `True`, but validator implementation should have to do this explicitly. Let's `raise NotImplementedError` here.

connaisseur/validators/notaryv2/__init__.py
```diff
@@ -0,0 +1 @@
+pass
```
peterthomassen, line 1: Is this necessary?

connaisseur/validators/notrayv1/__init__.py
```diff
@@ -0,0 +1 @@
+pass
```
peterthomassen, line 1: Is this necessary?

connaisseur/validators/notaryv2/notaryv2_validator.py
```diff
@@ -1,6 +1,10 @@
 from connaisseur.validators.interface import ValidatorInterface
+from connaisseur.image import Image
 
 
 class NotaryV2Validator(ValidatorInterface):
     def __init__(self, name: str, host: str, **kwargs):
+        raise NotImplementedError
+
+    def validate(self, image: Image, **kwargs):
         pass
```
peterthomassen, line 10: fail close?

connaisseur/validators/static/__init__.py
```diff
@@ -0,0 +1 @@
+pass
```
peterthomassen, line 1: Is this necessary?

connaisseur/validators/static/static_validator.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def __init__(self, name: str, approve: bool, **kwargs):
+        self.name = name
+        self.approve = approve
+
+    def validate(self, image: Image, **kwargs):
+        if not self.approve:
+            msg = "Static deny."
+            raise ValidationError(message=msg)
+        return None
```
peterthomassen, line 20: no-op

connaisseur/validators/notaryv2/notaryv2_validator.py
```diff
@@ -1,6 +1,10 @@
 from connaisseur.validators.interface import ValidatorInterface
+from connaisseur.image import Image
 
 
 class NotaryV2Validator(ValidatorInterface):
     def __init__(self, name: str, host: str, **kwargs):
+        raise NotImplementedError
```
peterthomassen, line 7: This could set `self.name = name`, instead of raising, and instead raise in `.validate()`. You can then call `super().__init__(name, host, **kwargs)` in the specific implementations.

Also, do *all* conceivable validators need a `host` argument?

connaisseur/validators/static/static_validator.py
```diff
@@ -0,0 +1,20 @@
...
+import logging
+
+
+class StaticValidator(ValidatorInterface):
+
+    name: str
+    approve: bool
+
+    def __init__(self, name: str, approve: bool, **kwargs):
+        self.name = name
```
peterthomassen, line 13: With the suggested change in the parent class, `super().__init__(name, **kwargs)`

helm/values.yaml
```diff
@@ -69,14 +76,11 @@ notaries:
...
 - pattern: "*:*"
-  verify: true
 - pattern: "k8s.gcr.io/*:*"
-  verify: false
-- pattern: "docker.io/securesystemsengineering/connaisseur:helm-*"
-  verify: false
+  validator: allow
 - pattern: "docker.io/securesystemsengineering/connaisseur:*"
-  verify: true
-  key: sse
```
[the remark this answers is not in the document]
phbelitz: hm. unsure how the new default configuration should look like...i'd go for:
```yaml
- pattern: "*:*"
  with:
    key: default
- pattern: "k8s.gcr.io/*:*"
  validator: allow
  validator: allow
```

showing explicitly, how to specify other keys within the `values.yaml` and allowing some key images (especially `connaisseur:helm-hook`)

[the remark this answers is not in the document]
phbelitz: would add the `with: key: default` to the cosign as well

connaisseur/config.py
phbelitz: I feel like we are doing all we can. Ppl can't break out of the directory through our `safe_path_func` and whatever funny business they put into the `secret_config` or `auth_file` will be validated against the schema in the end. They can overwrite existing values however they want, it still needs to be schema conform. I don't think doing the validation for each file on its own would change much here.

connaisseur/config.py
phbelitz: eh...that doesn't work for the static validators, who don't have any keys, so i temporarly took it out. I was think about putting the validation of validator-specific things into the respective validators, but haven't done so yet. then the validation of the config file will just check whether there is a `name` and `type` for a validator (+ only one is called default) and everything else is done by the validators themselves. but this will be probably done in a seperate PR :)

connaisseur/config.py
phbelitz: If there is only one validator and this one validator cannot be found, no validators can be found.

connaisseur/flask_server.py
phbelitz: The static validator return `None` as a trusted digest (if successful), indicating that nothing should change. In this case we don't wanna create any JSON patch and skip ahead. We could set the digest of the image to the trusted one and create patches only if the image has a digest, as we previously did, but then we might overwrite an existing digest with `None`. Connaisseur also checks image references that already have a digest.

connaisseur/validators/__init__.py
phbelitz: the `pass` or the file itself? `pytest` complains that it can't find the `connaisseur.validators` module, if there is no `__init__.py` defined. so i added them for all subdirectories :shrug:

connaisseur/config.py
phbelitz: :blush:

helm/values.yaml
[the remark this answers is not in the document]
peterthomassen: Why should we not validate the Connaisseur image with our key?

connaisseur/config.py
peterthomassen: I suggest to add a comment here that the schema validation hasn't run, and that we're only using a name parameter that is confined with `safe_path_func`.

Just to make sure that when this is later changed, no one will think that anything in `validator` is safe to use just because we're using `validator["name"]`. That is, let's be explicit about the "exceptional circumstance" that allows us to do this. (Exceptions confirm the rule! ;-) )

connaisseur/config.py
peterthomassen: That sounds reasonable. But shouldn't this orphaned comment be deleted, then? (If there's a to-do, let's make it an issue)

connaisseur/flask_server.py
peterthomassen: My question was not about the `if`, but about the side effect on `image`, namely that we're modifying the `image` object that is given as an argument to `__update_json_patch`. This function name doesn't sound like it would modify any of its arguments.

Why not do the `if` and `image.set_digest` outside of `__update_json_patch`, similar to what was done before?

connaisseur/validators/__init__.py
peterthomassen: The `pass`; it's fine for `__init__.py` to be empty. Is there any issue with it?

connaisseur/validators/notaryv2/notaryv2_validator.py
```diff
@@ -3,8 +3,8 @@
 
 
 class NotaryV2Validator(ValidatorInterface):
-    def __init__(self, name: str, host: str, **kwargs):
-        raise NotImplementedError
+    def __init__(self, name: str, **kwargs):
+        super().__init__(name, **kwargs)
```
peterthomassen, line 7: this is no-op and can go away

connaisseur/config.py
phbelitz: ah damn, i just realized, since validation hasn't run through we can't be sure a `name` field actually exists yet. so I better use `validator.get("name")` instead.

phbelitz:
The `connaisseur:helm-hook` image is needed for installing and uninstalling the webhook and currently has no signature. If `validator: allow` isn't set, we can't uninstall connaisseur anymore. This has always been the case.

peterthomassen:
> The `connaisseur:helm-hook` image is needed for installing and uninstalling the webhook and currently has no signature. If `validator: allow` isn't set, we can't uninstall connaisseur anymore. This has always been the case.

I see. What's the reason it has not been signed?

phbelitz:
> > The `connaisseur:helm-hook` image is needed for installing and uninstalling the webhook and currently has no signature. If `validator: allow` isn't set, we can't uninstall connaisseur anymore. This has always been the case.
> 
> I see. What's the reason it has not been signed?

Histroric reasons I guess. We didn't used to sign any of our images and only started with the main connaisseur image + this would make upgrading and deleting connaisseur from a cluster much slower, as the helm-hook image needs to be verified before it can delete the webhook.

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -0,0 +1,142 @@
...
+        self.keys = keys
+
+    def __get_key(self, key_name: str = None):
+        try:
+            if len(self.keys) < 2:
+                key = next(iter(self.keys))["key"]
+            else:
+                key_name = key_name or "default"
+                key = next(key["key"] for key in self.keys if key["name"] == key_name)
+            return "".join(key)
```
peterthomassen, line 33: Didn't we agree that there should be no special treatment if there's only one key? (I.e. it needs to be specified explicitly, or named "default", otherwise error out?)

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -0,0 +1,142 @@
...
+                    except Exception as err:
+                        msg = (
+                            "Could not retrieve valid and unambiguous digest from data "
+                            "received by cosign: {err_type}: {err}"
+                        )
+                        raise UnexpectedCosignData(
+                            message=msg, err_type=type(err).__name__, err=str(err)
+                        ) from err
+                    # remove prefix 'sha256'
+                    digests.append(digest[7:])
```
peterthomassen, line 82: `digest.removeprefix('sha256:')` (requires Python 3.9)

connaisseur/validators/notrayv1/notary.py
```diff
@@ -49,6 +50,28 @@ def __init__(self, name: str, host: str, pub_root_keys: list, **kwargs):
...
         self.is_acr = kwargs.get("is_acr", False)
         self.is_cosign = kwargs.get("is_cosign", False)
+        self.auth = kwargs.get("auth", {})
+        self.selfsigned_cert = self.__create_selfsigned_cert(
+            kwargs.get("selfsigned_cert")
+        )
+
+    def __create_selfsigned_cert(self, cert: str):
+        if not cert:
+            return None
```
peterthomassen, line 60: I think the caller should check whether it wants to call the function

connaisseur/validators/notrayv1/notary.py
```diff
@@ -49,6 +50,28 @@ def __init__(self, name: str, host: str, pub_root_keys: list, **kwargs):
         self.pub_root_keys = pub_root_keys
         self.is_acr = kwargs.get("is_acr", False)
         self.is_cosign = kwargs.get("is_cosign", False)
+        self.auth = kwargs.get("auth", {})
```
peterthomassen, line 53: better specify the keyword arguments explicitly in the function signature, with default:

```python
def __init__(self, name: str, host: str, pub_root_keys: list, is_acr=False, is_cosign=False, auth={})
```

connaisseur/validators/notrayv1/notary.py
```diff
@@ -49,6 +50,28 @@ def __init__(self, name: str, host: str, pub_root_keys: list, **kwargs):
...
+
+        if not safe_path_func(os.path.exists, "/app/connaisseur/certs", cert_path):
+            safe_path_func(
+                os.makedirs,
+                "/app/connaisseur/certs",
+                os.path.dirname(cert_path),
+                exist_ok=True,
+            )
+            with safe_path_func(open, "/app/connaisseur/certs", cert_path, "w") as file:
+                file.write(cert)
```
peterthomassen, line 72: This functionality did not previously exist. Why do we need it now if we didn't need it previously?

connaisseur/validators/notrayv1/notary.py
peterthomassen, line 53: Previously, this included [logic to extract credentials from environment variables](https://github.com/sse-secure-systems/connaisseur/blob/47b60abc46b4d84546ad2e48883cc8501660e47c/connaisseur/notary_api.py#L170-L174). Why did that go away?

peterthomassen (changes requested):
I reviewed that validators and added some comments.

There are some logic changes that do not appear to be motivated by the purpose of this PR (modularization). Not sure if all of it is intended or if something slipped through. I asked some questions about it; looking forward to settling these things.

Most questions would not have occurred, had the main commit of this PR been smaller. Next time such a larger PR/commit shows up, before I try to understand what's been done, I will add a review of type "request changes", and the requested change will be to break commits down into pieces that are actually digestable. :-)

Nevertheless, great work!

connaisseur/validators/notrayv1/notary.py
phbelitz: Previously the selfsigned certificates were extracted from the configuration during the helm chart creation. This has been changed, so that we don't really care how the configuration looks like and just put everything into a ConfigMap, except for some specifi fields which are put into a Secret. Then inside Connaisseur we merge these 2 together to get back the original configuration. 

Thing is, the python `requests` module expects the certificate to be inside a seperate `.crt` file and we can't pass it as a string. That's why the certificate is stored in a file inside the container.

connaisseur/validators/notrayv1/notary.py
phbelitz: Because I think putting things into a normal configfile and a "secret" configfile (one with sensitive data in it) and then merging them together, is a more modular approach then having a normal configfile and put in some specifi env variables in there, that depend on the type of configuration. We also don't really care what's in these config files, as long as the pass the schema validation.

connaisseur/validators/notrayv1/notary.py
peterthomassen: Understood. -- Why do we call the whole thing `selfsigned_cert`? `cert` would be easier. Also, we don't know whether it is self-signed.

connaisseur/validators/notrayv1/notary.py
peterthomassen: Ok. Such functional changes should ideally have their own commits, explaining in the commit message what's going on. :-)

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -0,0 +1,136 @@
...
+
+    def __init__(self, name: str, keys: list, **kwargs):
+        self.name = name
+        self.keys = keys
+
+    def __get_key(self, key_name: str = None):
+        try:
+            key_name = key_name or "default"
+            key = next(key["key"] for key in self.keys if key["name"] == key_name)
+            return "".join(key)
```
peterthomassen, line 30: The `key_name` assignment can go before the `try` block, and the `return` statement can go after the `except` block

connaisseur/validators/notrayv1/notary.py
```diff
@@ -60,68 +85,25 @@ def get_key(self, key_name: str = None):
...
-            else:
-                key_name = key_name or "default"
-                key = next(
-                    key["key"] for key in self.pub_root_keys if key["name"] == key_name
-                )
+            key_name = key_name or "default"
+            key = next(
+                key["key"] for key in self.pub_root_keys if key["name"] == key_name
+            )
             return "".join(key)
```
peterthomassen, line 92: The `key_name` assignment can go before the `try` block, and the `return` statement can go after the `except` block

connaisseur/validators/notrayv1/notary.py
```diff
@@ -47,8 +56,24 @@ def __init__(self, name: str, host: str, pub_root_keys: list, **kwargs):
         self.name = name
         self.host = host
         self.pub_root_keys = pub_root_keys
-        self.is_acr = kwargs.get("is_acr", False)
-        self.is_cosign = kwargs.get("is_cosign", False)
+        self.is_acr = is_acr
+        self.auth = auth
+        self.cert = self.__create_cert(cert) if cert else None
+
+    def __create_cert(self, cert: str):
```
peterthomassen, line 63: This sound like a certificate is generated, but it's actually only written. (I think this is where my initial confusion came from.) Renaming it to `__write_cert` would remove the ambiguity, but it's not too important.

peterthomassen (dismissed):
lgtm! Only cosmetic suggestions, but nothing to hold this up. :-)

peterthomassen:
What has changed since the last review / what do you want me to review?

helm/values.yaml
```diff
@@ -21,9 +21,16 @@ service:
...
-# public notary from dockerhub
-- name: dockerhub
+validators:
+- name: allow
+  type: static
+  approve: true
+- name: deny
+  type: static
+  approve: false
+- name: default
```
xopham, line 31: I am not sure if this is the best way to do this from a user perspective to get started fast.
Shouldn't we give this configuration a specific name and use it in the image policy below explicitly for `docker.io/library` (if I am not mistaken, that is the docker official images). We could call this `docker_basics` and the docker key `docker_official`.
In that case, we could add a default config below that would be applied to all images that are not official or sse. Something like:
```
- name: default
  type: notaryv1
  host: notary.docker.io
  pub_root_keys:
  - name: default
    key: |
      -----BEGIN PUBLIC KEY-----
      -----END PUBLIC KEY-----
```
if this key is setup as the key for all non-specific images in the image policy, it would mean that a new user only has to add the public key and it would work out of the box with minimal changes. One thing, I would consider important, is that just cloning the repo and installing connaisseur w/o changes also works and gives a meaningful error message. Essentially, we would be validating against an empty key (or we add a random key in the default config that would have to be changed). What do you think?

helm/values.yaml
```diff
@@ -21,9 +21,16 @@ service:
...
+# public notary form dockerhub
+  type: notaryv1
   host: notary.docker.io
   pub_root_keys:
```
xopham, line 35: as I see it, there is now `pub_root_keys`, `keys` (for cosign) and `root_keys` (in the comment) for setting keys. Do we need all of those or could we collapse this into 1 key value like `pubkeys` or something?

helm/values.yaml
peterthomassen: I like these ideas! However, I think we should not be putting random keys -- empty one (or placeholder) appeals much more to me. Otherwise, people don't know if they can change it or it's something hardcoded, and it looks suspicious anyway.

xopham: I agree with the concerns regarding a random key. Would you think adding a place holder like `<add your public key here>` and then checking for the place holder during verification to give a reasonable error message like "no valid key specified for image" or something be okay?
Would achieve a super simple start without further actions, a simple adjustment small use cases by adding only one key and still be open to get the full functionality...

peterthomassen: Sure, why not. -- I think we should not have special treatment in our application code to check for special values like placeholders.

helm/values.yaml
phbelitz: we can do that. for notary things might get confusing as there are so many keys flying around and pub_root_keys makes things more clear. I don't really have any stakes in this and can settle with just calling them `keys` or `pub_keys`.

xopham: Ok. Let's see. I would prefer to call them `pub` something to make sure people are aware that it is only public keys, otherwise we might have people adding a private key.
To the confusion about keys: cosign currently only has one type of keys and those are not "root" as there is no hierarchy. So that would mean, we need to add something like `pubkeys` or `pub_keys` (not sure which one is advised. Now for notary v1, there is different types of keys, but in connaisseur you can only specify the public root key and that seems unlikely to change, specifically as I do not see nv1 changing much any time soon or ever again. Consequently, it is only a matter of whether someone would think it might make sense to add a delegation key there (to my knowledge the only other keys in nv1). At that point, that someone is attempting a more complicated use case and would have had to take the extra steps to create a delegation key to begin with. I presume those uses will take the time to read our hopefully understandable docs.
As such, I would see no problem in calling it e.g. `pub_keys` in both cases to keep things simple and consistent. Would you see a case where this might become conflicting?

peterthomassen (approved):
lgtm!

peterthomassen (approved):

PR #157, opened 2021-05-05
Fix/wrong image path

phbelitz:
Fixed an error where acquiring image references and patching them pointed to different locations. The acquisition of image references previsouly just stored all in a list, whether they were from the initContainers or not. During validation we iterated over this list, keeping each items index in mind so we can then patch the correct location in the received request. Problem is, that the initContainers and regular containers have differnt indizes in the request, while we stored all in the same list. The consequence of this was, should someone deploy a resource with valid initContainer and valid regular container, connaisseur will attempt to patch container indizes, that don't exist. Now each image will keep track of its own index and container_type ("containers" or "initContainers"), so when patching the request after validation, the correct container gets patched.

### DIFF
--- a/connaisseur/flask_server.py
+++ b/connaisseur/flask_server.py
@@ -14,7 +14,6 @@ from connaisseur.alert import send_alerts
 from connaisseur.config import Config
 from connaisseur.admission_request import AdmissionRequest
 from connaisseur.policy import ImagePolicy
-from connaisseur.image import Image
 
 
 DETECTION_MODE = os.environ.get("DETECTION_MODE", "0") == "1"
@@ -130,35 +129,32 @@ def __create_logging_msg(msg: str, **kwargs):
     return str({"message": msg, "context": dict(**kwargs)})
 
 
-def __create_json_patch(path: str, index: int, image: Image):
-    return {"op": "replace", "path": path.format(str(index)), "value": str(image)}
-
-
 def __admit(admission_request: AdmissionRequest):
+    logging_context = dict(admission_request.context)
     policy = ImagePolicy()
-    parent_images = admission_request.wl_object.parent_images
     patches = []
 
-    for index, c_image in enumerate(admission_request.wl_object.container_images):
+    for type_index, image in admission_request.wl_object.containers.items():
+        original_image = str(image)
+        type_, index = type_index
+        logging_context.update(image=original_image)
+
+        # child resources have mutated image names, as their parents got mutated
+        # before their creation. this may result in mismatch of rules or duplicate
+        # lookups for already approved images. so child resources are automatically
+        # approved without further check ups, when their parents were approved
+        # earlier.
+        if image in admission_request.wl_object.parent_containers.values():
+            msg = f'automatic child approval for "{original_image}".'
+            logging.info(__create_logging_msg(msg, **logging_context))
+            continue
+
         try:
-            logging_context = dict(admission_request.context, image=c_image)
-
-            # child resources have mutated image names, as their parents got mutated
-            # before their creation. this may result in mismatch of rules or duplicate
-            # lookups for already approved images. so child resources are automatically
-            # approved without further check ups, when their parents were approved
-            # earlier.
-            if c_image in parent_images:
-                msg = f'automatic child approval for "{c_image}".'
-                logging.info(__create_logging_msg(msg, **logging_context))
-                continue
-
-            image = Image(c_image)
             policy_rule = policy.get_matching_rule(image)
             validator = CONFIG.get_validator(policy_rule.validator)
 
             msg = (
-                f'starting verification of image "{c_image}" using rule '
+                f'starting verification of image "{original_image}" using rule '
                 f'"{str(policy_rule)}" with arguments {str(policy_rule.arguments)}'
                 f' and validator "{str(validator)}".'
             )
@@ -169,17 +165,15 @@ def __admit(admission_request: AdmissionRequest):
             )
 
             trusted_digest = validator.validate(image, **policy_rule.arguments)
-            if trusted_digest:
-                image.set_digest(trusted_digest)
-                patches += [
-                    __create_json_patch(
-                        admission_request.wl_object.container_path, index, image
-                    )
-                ]
-
-            msg = f'successful verification of image "{c_image}"'
-            logging.info(__create_logging_msg(msg, **logging_context))
         except BaseConnaisseurException as err:
             err.update_context(**logging_context)
             raise err
+
+        if trusted_digest:
+            image.set_digest(trusted_digest)
+            patches.append(
+                admission_request.wl_object.get_json_patch(image, type_, index)
+            )
+        msg = f'successful verification of image "{original_image}"'
+        logging.info(__create_logging_msg(msg, **logging_context))
     return get_admission_review(admission_request.uid, True, patch=patches)
--- a/connaisseur/image.py
+++ b/connaisseur/image.py
@@ -86,3 +86,6 @@ class Image:
         )
         tag = f":{self.tag}" if not self.digest else f"@sha256:{self.digest}"
         return f"{repo_reg}{self.name}{tag}"
+
+    def __eq__(self, other):
+        return str(self) == str(other)
--- a/connaisseur/workload_object.py
+++ b/connaisseur/workload_object.py
@@ -1,4 +1,5 @@
 import connaisseur.kube_api as k_api
+from connaisseur.image import Image
 from connaisseur.exceptions import UnknownAPIVersionError, ParentNotFoundError
 
 
@@ -15,7 +16,7 @@ SUPPORTED_API_VERSIONS = {
 
 
 class WorkloadObject:
-    container_path = "/spec/template/spec/containers/{}/image"
+    container_path = "/spec/template/spec/{container_type}/{index}/image"
 
     def __new__(
         cls, request_object: dict, namespace: str
@@ -49,8 +50,8 @@ class WorkloadObject:
             )
 
     @property
-    def parent_images(self):
-        parent_images = []
+    def parent_containers(self):
+        parent_containers = {}
         for owner in self._owner:
             api_version = owner["apiVersion"]
             kind = owner["kind"].lower() + "s"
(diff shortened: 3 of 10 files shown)


connaisseur/flask_server.py
```diff
@@ -138,41 +137,37 @@ def __create_logging_msg(msg: str, **kwargs):
...
-            logging_context = dict(admission_request.context, image=c_image)
+        for image in admission_request.wl_object.container_images:
+            logging_context.update(image=str(image))
 
             # child resources have mutated image names, as their parents got mutated
             # before their creation. this may result in mismatch of rules or duplicate
             # lookups for already approved images. so child resources are automatically
             # approved without further check ups, when their parents were approved
             # earlier.
-            if c_image in parent_images:
```
Starkteetje, line 159: Does this not work due to the image being an object? If so, can you not overwrite the equality for the object instead of transmuting the list of images to a list of strings and the image to a string?

connaisseur/tests/integration/integration-test.sh
```diff
@@ -108,6 +108,18 @@ else
...
+echo 'Testing deployment of valid init container along with a valid container...'
+kubectl apply -f connaisseur/tests/integration/valid_init_container.yaml >output.log 2>&1 || true
+NUMBER_OF_VALID_DEPLOYMENTS+=1
+
+if [[ "$(cat output.log)" != 'pod/connaisseur-integration-test-pod-valid-init created' ]]; then
+  echo 'Failed to deploy a valid initContainer along with a valid container. Output:'
+  cat output.log
+  exit 1
+else
+  echo 'Successfully allowed image in init container'
```
Starkteetje, line 120:
```suggestion
  echo 'Successfully allowed valid image in init container and container'
```

connaisseur/image.py
```diff
@@ -25,7 +25,13 @@ class Image:
     tag: str
     digest: str
 
-    def __init__(self, image: str):
+    index: int
+    container_type: str
+
+    def __init__(self, image: str, index: int, container_type: str):
```
Starkteetje, line 31: This doesn't really define an image though. The position of an image in a deployment should not be part of an image; maybe it would be better to define a wrapper class ContainerImage/DeploymentImage/AdmissionImageInfo(?) that holds `(image: Image, index: int, container_type: str)` instead of pushing it onto the image class. Then my comment above about `c_image in parent_images` also works again if you compare Images by their full-name, which would be what one expects

connaisseur/flask_server.py
phbelitz: did as you asked

connaisseur/image.py
phbelitz: I added the `Container` class as a wrapper.

phbelitz:
i think it's best to wait for #156 before doing anymore work here

connaisseur/container.py
```diff
@@ -0,0 +1,15 @@
...
+    image: Image
+    index: int
+    container_type: str
+
+    def __init__(self, image: str, index: int, container_type: str):
+        self.image = Image(image)
+        self.index = index
+        self.container_type = container_type
+
+    def __eq__(self, other):
```
Starkteetje, line 14: Is there a reason for this equality? Given that this is explicitily encapsulates the index in addition to the Image, it seems weird that it would define equality the same way. If there's a reason, I'd add a comment

phbelitz: https://github.com/sse-secure-systems/connaisseur/pull/157#discussion_r627288860

twas your idea to overwrite the equality function. this also makes the code for checking whether the current container already has been validated previously really **sexy**.

```python
 if container in admission_request.wl_object.parent_containers:
     msg = f'automatic child approval for "{str(container.image)}".'
     logging.info(__create_logging_msg(msg, **logging_context))
     continue
```

Starkteetje: Then I'd still add a comment to the equality telling that equality is also used for containers to check whether their image is the same and there would be no other apparent reason to compare containers with the intention of having the index play a part
As I said above, I'm not opposed, but it looks weird since its a data structure that explicitly has an additional parameter that is then not used. Hence, I would say it warrants a comment

peterthomassen: Thinking about it, I'd say the correct way of doing this would be to keep the index outside of the `Container` object. `Container` objects are members of [lists like this](https://github.com/sse-secure-systems/connaisseur/pull/157/commits/c85234c351640efcac62b20487821ba969327e7a#diff-6346c82f50278989f0e85329f1488c325a9bc84afdc883e961bd9ba096384affR80), and that list could instead be a dictionary mapping `index`es on `Container` objects (and the object would not have an `index` attribute). The same would then apply to the `container_type` argument.

In other words, what you need is something like
```python
container_list = {
    index: Container(container["image"])
    for index, container in enumerate(spec["containers"])
]
init_container_list = [
    index: Container(container["image"])
    for index, container in enumerate(spec.get("initContainers", []))
]
```
and then merge them into something with a combined look-up key `(container_type, index)`. Or directly:
```python
container_list = {
    (container_type, index): Container(container["image"])
    for container_type in ['containers', 'initContainers']
    for index, container in enumerate(spec.get(container_type, []))
]
```

You can then do your comparison with `if container in ...parent_containers.values()`.

Once you have this structure, I'm not sure whether the `Container` class is still needed.

connaisseur/flask_server.py
```diff
@@ -135,35 +134,30 @@ def __create_logging_msg(msg: str, **kwargs):
...
+            logging_context.update(image=str(container.image))
 
             # child resources have mutated image names, as their parents got mutated
             # before their creation. this may result in mismatch of rules or duplicate
             # lookups for already approved images. so child resources are automatically
             # approved without further check ups, when their parents were approved
             # earlier.
-            if c_image in parent_images:
-                msg = f'automatic child approval for "{c_image}".'
+            if container in admission_request.wl_object.parent_containers:
```
annekebr, line 151: correct me if I'm wrong, but if you use sth like
`if container.image in [container['image'] for container in admission_request.wl_object.parent_containers]:` here instead you could get rid of the controversial `__eq__` function for containers, couldn't you?

connaisseur/container.py
phbelitz: i implemented @peterthomassen's suggestion and removed the `Container` wrapper.

connaisseur/flask_server.py
phbelitz: did go for this implementaion: https://github.com/sse-secure-systems/connaisseur/pull/157#discussion_r666315210

connaisseur/flask_server.py
```diff
@@ -130,35 +129,34 @@ def __create_logging_msg(msg: str, **kwargs):
...
 
-def __create_json_patch(path: str, index: int, image: Image):
-    return {"op": "replace", "path": path.format(str(index)), "value": str(image)}
-
-
 def __admit(admission_request: AdmissionRequest):
-    policy = ImagePolicy()
-    parent_images = admission_request.wl_object.parent_images
-    patches = []
+    try:
```
peterthomassen, line 133: That `try` block is *very* long. Can we narrow it down to closer where the relevant exception could actually occur?

connaisseur/workload_object.py
```diff
@@ -70,38 +71,48 @@ def parent_images(self):
...
         spec = self._spec["template"]["spec"]
-        return [
-            container["image"]
-            for container in (spec["containers"] + spec.get("initContainers", []))
-        ]
+        return {
+            (container_type, index): Image(container["image"])
+            for container_type in ["containers", "initContainers"]
+            for index, container in enumerate(spec.get(container_type, []))
+        }
```
peterthomassen, line 84: This code is often duplicated in child classes, with the only difference being the value of `spec`.

I would change line 83 so that it uses `self.spec`, and then remove all of the `spec =` assignments (including in child classes), and instead make a property called `spec` which returns the right-hand side value that was previously assigned. This removes the code duplication, but you still have the flexibility and on-demand evaluation.

peterthomassen (changes requested):

connaisseur/flask_server.py
phbelitz: This is intended. Whatever error may occur, we wanna catch it here and add some needed logging information to the error object, which we wouldn't have durring the initial raising of the error. This is then catched by yet another `try except` where the actual logging happens.

connaisseur/workload_object.py
phbelitz: did as you asked

connaisseur/flask_server.py
peterthomassen: That argument does not convince me, because you catch `BaseConnaisseurException` exceptions only, but the most part of the code does not raise that. -- Worse, `logging_context` depends on the loop iteration, but the `return` statement *after* the loop is also part of the `try` block. That means if the `return` statement somehow raises a `BaseConnaisseurException` exception, you will have a misleading logging context.

I looked which parts can throw that exception, and it seems like it's only the methods which are called on the policy and the validator. I would thus suggest to rewrite this as follows:

```python
def __admit(admission_request: AdmissionRequest):
    logging_context = dict(admission_request.context)
    policy = ImagePolicy()
    patches = []

    for type_index, image in admission_request.wl_object.containers.items():
        original_image = str(image)
        type_, index = type_index
        logging_context.update(image=original_image)

        # child resources have mutated image names, as their parents got mutated
        # before their creation. this may result in mismatch of rules or duplicate
        # lookups for already approved images. so child resources are automatically
        # approved without further check ups, when their parents were approved
        # earlier.
        if image in admission_request.wl_object.parent_containers.values():
            msg = f'automatic child approval for "{original_image}".'
            logging.info(__create_logging_msg(msg, **logging_context))
            continue

        try:
            policy_rule = policy.get_matching_rule(image)
            validator = CONFIG.get_validator(policy_rule.validator)

            msg = (
                f'starting verification of image "{original_image}" using rule '
                f'"{str(policy_rule)}" with arguments {str(policy_rule.arguments)}'
                f' and validator "{str(validator)}".'
            )
            logging.debug(
                __create_logging_msg(
                    msg, **logging_context, policy_rule=policy_rule, validator=validator
                )
            )

            trusted_digest = validator.validate(image, **policy_rule.arguments)
            err.update_context(**logging_context)
            raise err

        if trusted_digest:
            image.set_digest(trusted_digest)
            patches.append(admission_request.wl_object.get_json_patch(image, type_, index))
        msg = f'successful verification of image "{original_image}"'
        logging.info(__create_logging_msg(msg, **logging_context))
        return get_admission_review(admission_request.uid, True, patch=patches)
```

(Note that I also exchanged `+=` with `.append()`.)

peterthomassen (changes requested):

connaisseur/flask_server.py
phbelitz: done.

peterthomassen (approved):
nice! lgtm.

phbelitz:
did the rebase

peterthomassen (approved):
lgtm (did not inspect; rebase only)

PR #161, opened 2021-05-14
fix: alert configuration error on hook image

annekebr:
Do not check alert config on hook image to ensure `make uninstall` is working

### DIFF
--- a/connaisseur/alert.py
+++ b/connaisseur/alert.py
@@ -151,11 +151,11 @@ class Alert:
 def send_alerts(
     admission_request: AdmissionRequest, admit_event: bool, reason: str = None
 ):
+    if __is_hook_image(admission_request):
+        return
     al_config = AlertingConfiguration()
     event_category = "admit_request" if admit_event else "reject_request"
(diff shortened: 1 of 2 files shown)


connaisseur/alert.py
```diff
@@ -151,11 +151,11 @@ def __get_headers(receiver_config):
 def send_alerts(
     admission_request: AdmissionRequest, admit_event: bool, reason: str = None
 ):
+    if not __is_not_hook_image(admission_request):
```
phbelitz, line 154: maybe we wanna change the not(not(not(not()))) .... so just `if __is_hook_image(): ..` and switch all return values of `__is_not_hook_image`.

Starkteetje, line 154: This is a double negative and since `__is_not_hook_image` is not used elsewhere (as far as i can see), you should change that to `__is_hook_image`

phbelitz (approved):
\*happy noises\*

phbelitz (approved):

PR #163, opened 2021-05-28
Fix/nv1 parallel

phbelitz:
A major amount of connaisseur's runtime (for nv1 validation) is wasted with receiving the trust data from the notary server. To reduce this time, the requests have been parallelized.

### DIFF
--- a/connaisseur/validators/notaryv1/notary.py
+++ b/connaisseur/validators/notaryv1/notary.py
@@ -5,7 +5,6 @@ import requests
 import yaml
 from connaisseur.image import Image
 from connaisseur.validators.notaryv1.tuf_role import TUFRole
-from connaisseur.validators.notaryv1.trust_data import TrustData
 from connaisseur.exceptions import (
     UnreachableError,
     NotFoundException,
@@ -150,7 +149,7 @@ class Notary:
 
         response.raise_for_status()
 
-        return TrustData(response.json(), str(role))
+        return response.json()
 
     def get_delegation_trust_data(self, image: Image, role: TUFRole, token: str = None):
         try:
--- a/connaisseur/validators/notaryv1/notaryv1_validator.py
+++ b/connaisseur/validators/notaryv1/notaryv1_validator.py
@@ -1,5 +1,6 @@
 import os
 import base64
+from multiprocessing import Pool
 from connaisseur.validators.interface import ValidatorInterface
 from connaisseur.image import Image
 from connaisseur.admission_request import AdmissionRequest
@@ -7,6 +8,7 @@ from connaisseur.validators.notaryv1.notary import Notary
 from connaisseur.util import safe_path_func
 from connaisseur.validators.notaryv1.tuf_role import TUFRole
 from connaisseur.validators.notaryv1.key_store import KeyStore
+from connaisseur.validators.notaryv1.trust_data import TrustData
 from connaisseur.exceptions import (
     UnreachableError,
     NotFoundException,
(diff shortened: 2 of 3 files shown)


connaisseur/validators/notaryv1/notary.py
```diff
@@ -150,7 +149,7 @@ def get_trust_data(self, image: Image, role: TUFRole, token: str = None):
 
         response.raise_for_status()
 
-        return TrustData(response.json(), str(role))
+        return response.json()
```
Starkteetje, line 152: Why the change to not return the TrustData object? So a potential error gets exposed during the validator?

connaisseur/validators/notaryv1/notaryv1_validator.py
```diff
@@ -240,16 +256,31 @@ def __search_image_targets_for_tag(trust_data: dict, image: Image):
...
+                    [(image, TUFRole(delegation)) for delegation in delegations],
+                )
+                delegation_trust_data_list = result.get(timeout=30)
+            except Exception as err:
+                msg = "Error retrieving delegation trust data form notary."
+                raise NotFoundException(
+                    message=msg, notary=str(self.notary.name)
+                ) from err
+
+        # when delegations are added to the repository, but weren't yet used for signing,
```
Starkteetje, line 272: This comment doesn't make sense anymore since TrustData() cannot be None if I see this correctly. However, this was already that way before this commit, so you probably changed that during your refactoring/modularization commit. Not having looked into it, it looks like we lost a case here? Can you investigate?

connaisseur/validators/notaryv1/notary.py
phbelitz: If i use `TrustData` there, python throws an error that some instance can't be _pickled_ ...? My guess is that if you want to run something in parallel, the return value of the function being run in parallel must of a primitiv type, or at least be _picklable_, which `TrustData` aparently isn't, but `str` is. 

tldr; parallelization doesn't work with `TrustData` but with `str`.

maybe @peterthomassen has more insight ...?

peterthomassen: Let's please have a discussion about this. Pickle is a vulnerability-ridden feature of Python (in particular, RCE vulnerabilities), and one should never use it, unless you can more or less prove that it's _not_ an issue in the specific use case.

peterthomassen: Also, this change breaks the tests (which you're fixing in another commit). It's not a problem, but it means that this is not a refactor (because refactors don't break tests, by definition). It is actually a (not very important) API change.

I would suggest to include the test fixes in the same commit, so that there is no commit in the history which fails the tests.

connaisseur/validators/notaryv1/notaryv1_validator.py
```diff
@@ -111,8 +113,22 @@ def __process_chain_of_trust(
...
+                trust_data_list = result.get(timeout=30)
+            except Exception as err:
+                msg = "Error retrieving trust data form notary."
+                raise NotFoundException(
+                    message=msg, notary=str(self.notary.name)
+                ) from err
+        trust_data = {
+            tuf_roles[i]: TrustData(trust_data_list[i], tuf_roles[i])
+            for i in range(len(tuf_roles))
+        }
```
peterthomassen, line 131:
```python
trust_data = {k: TrustData(v, k) for k, v in zip(tuf_roles, trust_data_list)}
```

or similar (did not test)

peterthomassen (changes requested):

phbelitz:
This PR will be closed in favor of the `asyncio` approach.

PR #164, opened 2021-06-03
Fix/fix preconfig

xopham:
this PR fixes the pre-configuration in `helm/values.yaml` and improves the docu around it


helm/values.yaml
```diff
@@ -20,20 +20,24 @@ service:
...
 
-# configure access to the notary server
+### VALIDATORS ###
+# validators are a set of configurations (hosts, types, public keys) that can be
+# used for validating one or multiple images (or image signatures). They are tied
+# to their respective image(s) via the image policy below. There are a few handy
+# validators pre-configured.
 validators:
-- name: allow
+- name: allow     # always allow image
```
peterthomassen, line 29: The functionality does not follow from the name, so the comment is misleading.

Also, it's common to use two spaces before inline comments

helm/values.yaml
```diff
@@ -20,20 +20,24 @@ service:
...
   type: static
   approve: true
-- name: deny
+- name: deny      # always deny image
   type: static
   approve: false
-- name: default  # the default validator which is used, if no validators are specified
+- name: default   # the default validator which is always used, if no validator is specified in image policy
```
peterthomassen, line 35:
```suggestion
- name: default   # the default validator which is used if no validator is specified in image policy
```

helm/values.yaml
```diff
@@ -20,20 +20,24 @@ service:
...
+# used for validating one or multiple images (or image signatures). They are tied
+# to their respective image(s) via the image policy below. There are a few handy
+# validators pre-configured.
 validators:
-- name: allow
+- name: allow     # always allow image
   type: static
   approve: true
-- name: deny
+- name: deny      # always deny image
```
peterthomassen, line 32: ditto

helm/values.yaml
```diff
@@ -20,20 +20,24 @@ service:
...
   type: static
   approve: false
-- name: default  # the default validator which is used, if no validators are specified
+- name: default   # the default validator which is always used, if no validator is specified in image policy
   type: notaryv1  # cosign is also experimetally supported
   host: notary.docker.io
   pub_keys:
-  - name: default  # the default key which is used, if no keys are specified
-    key: |
+  - name: default   # the default key which is always used, if no key is specified in image policy
```
peterthomassen, line 39:
```suggestion
  - name: default   # the default key which is used if no key is specified in image policy
```

helm/values.yaml
```diff
@@ -47,55 +51,58 @@ validators:
   #   password: notarypass
   #   # or
   #   secret_name: notarysecret  # define a k8s secret with `username` and `password` fields
-- name: docker_essentials  # public notary form dockerhub
+- name: dockerhub_essentials   # pre-configured validator for public notary from dockerhub
```
peterthomassen, line 54:
```suggestion
- name: dockerhub_essentials   # pre-configured validator for public notary from Docker Hub
```

helm/values.yaml
```diff
@@ -47,55 +51,58 @@ validators:
...
   type: notaryv1
   host: notary.docker.io
   pub_keys:
-  # this is the public key for all official dockerhub images (e.g. redis)
-  - name: docker_official
+  - name: docker_official   # public key for official docker images (e.g. redis)
```
peterthomassen, line 58:
```suggestion
  - name: docker_official   # public key for official docker images
```

helm/values.yaml
```diff
@@ -47,55 +51,58 @@ validators:
...
     key: |
       -----BEGIN PUBLIC KEY-----
       MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEOXYta5TgdCwXTCnLU09W5T4M4r9f
       QQrqJuADP6U7g5r9ICgPSmZuRHP/1AYUfOQW3baveKsT969EfELKj1lfCA==
       -----END PUBLIC KEY-----
-  # this is the public key to the `securesystemsengineering` repo on dockerhub
-  - name: sse_official
+  - name: sse_official      # public key for connaisseur images
```
peterthomassen, line 64:
```suggestion
  - name: sse_official      # public key for Connaisseur images
```

helm/values.yaml
```diff
@@ -47,55 +51,58 @@ validators:
...
       -----BEGIN PUBLIC KEY-----
       MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEsx28WV7BsQfnHF1kZmpdCTTLJaWe
       d0CA+JOi8H4REuBaWSZ5zPDe468WuOJ6f71E7WFg3CVEVYHuoZt2UYbN/Q==
       -----END PUBLIC KEY-----
-- name: cosign_demo  # for verifying cosign signatures
+- name: cosign_essentials   # pre-configured validator for public cosign signatures
   type: cosign
   pub_keys:
-  - name: sse
+  - name: sse_official      # public key for connaisseur images (currently only testimages)
```
peterthomassen, line 73: What test images?

Also, time dependent statements (like "currently") are prone to become outdated.

helm/values.yaml
```diff
@@ -47,55 +51,58 @@ validators:
...
       MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEvtc/qpHtx7iUUj+rRHR99a8mnGni
       qiGkmUb9YpWWTS4YwlvwdmMDiGzcsHiDOYz6f88u2hCRF5GUCvyiZAKrsA==
       -----END PUBLIC KEY-----
 
-# the image policy, which defines all repositories that need to be
-# verified. more detail in the git repo README.md
+### IMAGE POLICY ###
+# the image policy ties validators and images together whereby always only the most
+# specific rule (pattern) is applied. Specify if and how images should be validated
+# by which validator.
```
peterthomassen, line 83:
```suggestion
# The image policy ties validators and images together. Only the most specific rule (pattern)
# is applied. Specify whether (and how) images should be validated by which validator.
```

peterthomassen (changes requested):
Nice! Only minor suggestions.

I did not review tests/integration/preconfig-integration-test.sh because I'm not familiar with what it's testing. Appreciate if someone else could do it.

helm/values.yaml
xopham: we maintain a few signed testimages on dockerhub that we for example use for integration tests. We could use the official key also to sign connaisseur in the future, however that is not being done for now, as everything is experimental

phbelitz (approved):

helm/values.yaml
peterthomassen: Why is testing stuff in the configuration that we ship by default?

peterthomassen (approved):
lgtm

helm/values.yaml
xopham: we can use the shipped cosign key to also sign connaisseur itself in the future. otherwise, it may make getting started easier. I also added the testimages in the image policy below. That is sort of nice to demo, but might be conflicting with the default configuration. Do you think I should remove it?

peterthomassen: I think @Starkteetje can weigh in here.

helm/values.yaml
```diff
@@ -15,25 +15,33 @@ deployment:
...
-# configure connaisseur service
+# configure Connaisseur service
 service:
   type: ClusterIP
   port: 443
 
-# configure access to the notary server
+### VALIDATORS ###
+# validators are a set of configurations (hosts, types, public keys) that can be
+# used for validating one or multiple images (or image signatures). They are tied
```
Starkteetje, line 25: Just bc I noticed: If you start a sentence in a comment capitalized, then the first sentence should also be capitalized. That kind of (but not strictly) breaks with all the comment starting lower-case. I also think in other places senteces are not capitalized, not sure though

helm/values.yaml
```diff
@@ -15,25 +15,33 @@ deployment:
...
+- name: default
   type: notaryv1  # cosign is also experimetally supported
-  host: notary.docker.io
+  host: notary.docker.io # configure the notary server to be used
   pub_keys:
-  - name: default  # the default key which is used, if no keys are specified
-    key: |
+  # the `default` key is used if no key is specified in image policy
+  - name: default
+    key: |  # enter your key below
```
Starkteetje, line 44: This way the default validator is defined but not operable. Would either comment it out completly or set to docker_official

helm/values.yaml
```diff
@@ -15,25 +15,33 @@ deployment:
...
   type: static
   approve: true
+# static validator that denies each image
 - name: deny
   type: static
   approve: false
-- name: default  # the default validator which is used, if no validators are specified
+# the `default` validator is used if no validator is specified in image policy
+- name: default
   type: notaryv1  # cosign is also experimetally supported
```
Starkteetje, line 39: If `cosign` is the correct type, I'd highlight it with `"cosign"` if you want to have this comment here

helm/values.yaml
```diff
@@ -47,56 +55,43 @@ validators:
...
   validator: allow
-- pattern: "docker.io/securesystemsengineering/connaisseur:*"
-  validator: dockerhub
+- pattern: "docker.io/securesystemsengineering/*:*"
+  validator: dockerhub_essentials
   with:
-    key: sse
+    key: sse_official
 - pattern: "docker.io/securesystemsengineering/connaisseur:helm-hook-*"
   validator: allow
```
Starkteetje, line 94: That this stays at allow is the chicken-egg problem during bootstrapping?

helm/values.yaml
Starkteetje: Afaik, and as documented in the [RELEASING.md](https://github.com/sse-secure-systems/connaisseur/blob/master/RELEASING.md) we do sign the production Connaisseur images. Then adding our pubkey as config with the image policy prescribing said key for our images is worthwhile I'd say. If it were only for testing, I'd remove it instead, but as it actually has sec benefits. I'd keep it as is

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -19,23 +19,25 @@ class CosignValidator(ValidatorInterface):
     name: str
     keys: list
 
-    def __init__(self, name: str, pub_keys: list, **kwargs):
+    def __init__(self, name: str, trust_roots: list, **kwargs):
         self.name = name
-        self.keys = pub_keys
+        self.keys = trust_roots
```
annekebr, line 24:
```suggestion
        self.trust_roots = trust_roots
```

and then adopt in the following lines

helm/values.yaml
```diff
@@ -47,72 +55,59 @@ validators:
...
   - name: docker_official
     key: |
-      -----BEGIN PUBLIC KEY-----
+      -----BEGIN PUBLIC key-----
       MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEOXYta5TgdCwXTCnLU09W5T4M4r9f
       QQrqJuADP6U7g5r9ICgPSmZuRHP/1AYUfOQW3baveKsT969EfELKj1lfCA==
-      -----END PUBLIC KEY-----
-  # this is the public key to the `securesystemsengineering` repo on dockerhub
+      -----END PUBLIC key-----
+  # public key for Connaisseur images
```
annekebr, line 70: That is not true is it? If we deploy other images in the `securesystemsengineering` repo they'd be signed by the same key right? (Which is ok as we discussed but I'd change the comment back to the repo wording)

helm/values.yaml
annekebr: I think this comment doesn't really fit here. Maybe add a list of supported validators to the comment in the beginning?

helm/values.yaml
```diff
@@ -47,72 +55,59 @@ validators:
...
   type: notaryv1
   host: notary.docker.io
-  pub_keys:
-  # this is the public key for all official dockerhub images (e.g. redis)
+  trust_roots:
+    # public key for official docker images (https://hub.docker.com/search?q=&type=image&image_filter=official)
   - name: docker_official
     key: |
-      -----BEGIN PUBLIC KEY-----
+      -----BEGIN PUBLIC key-----
```
annekebr, line 66: That seems weird and happened in a few places. Replace all mistake maybe?

helm/values.yaml
```diff
@@ -47,72 +55,59 @@ validators:
...
+    trust_root: docker_official
 - pattern: "k8s.gcr.io/*:*"
   validator: allow
-- pattern: "docker.io/securesystemsengineering/connaisseur:*"
-  validator: dockerhub
+- pattern: "docker.io/securesystemsengineering/*:*"
+  validator: dockerhub_essentials
   with:
-    key: sse
+    trust_root: sse_official
```
annekebr, line 92: Here I'd strongly vote for renaming to `securesystemsengineering_official`

helm/values.yaml
annekebr: Ah I think I know what you wanted to say: I'd change the comment to something like "this is an example only; you can choose from the list of available validators"

helm/values.yaml
```diff
@@ -47,72 +55,59 @@ validators:
...
-# verified. more detail in the git repo README.md
+### IMAGE POLICY ###
+# the image policy ties validators and images together whereby always only the most specific rule (pattern)
+# is applied. Specify if and how images should be validated by which validator.
 policy:
 - pattern: "*:*"
-  validator: dockerhub
 - pattern: "docker.io/library/*:*"
-  validator: dockerhub
+  validator: dockerhub_essentials
```
annekebr, line 84: I don't like the renaming to `dockerhub_essentials` because I don't think allowing all official dockerhub image is _essential_, but the naming could discourage people who want to do stricter validation.

helm/values.yaml
```diff
@@ -47,72 +55,59 @@ validators:
...
+# namespaced validation allows to restrict the namespaces that will be subject to Connaisseur verification.
+# When enabled, all namespaces with label "de.securesystems.connaisseur/webhook=ignore" will be ignored.
+namespacedValidation: false
 
 # debug: true
 
 # alerting is implemented in form of simple POST requests with json payload
-# you can use and/or adapt the predefined Slack/OpsGenie/Keybase templates and the examples below
-# to channel alert notifications to Slack/OpsGenie/Keybase or create a custom template for a customized alert
+# you can use and/or adapt the predefined Slack/OpsGenie/keybase templates and the examples below
```
annekebr, line 109: replace all?

helm/values.yaml
```diff
@@ -128,7 +123,7 @@ targetNamespaces: ['*']
 #      - template: opsgenie #REQUIRED!
 #        receiver_url: https://api.eu.opsgenie.com/v2/alerts #REQUIRED!
 #        priority: 4 #(defaults to 3)
-#        custom_headers: ["Authorization: GenieKey <Your-Genie-Key>"]
```
annekebr, line 131: replace all?

helm/values.yaml
```diff
@@ -146,5 +141,5 @@ targetNamespaces: ['*']
 #  reject_request:
 #    templates:
 #      - template: keybase  #REQUIRED!
-#        receiver_url: https://bots.keybase.io/webhookbot/<Your-Keybase-Hook-Token>
```
annekebr, line 149: replace all?

helm/values.yaml
xopham: as discussed, with @peterthomassen and @phbelitz , I switched to commenting the whole key, but key the validator

helm/values.yaml
xopham: fixed it

helm/values.yaml
```diff
@@ -15,28 +15,36 @@ deployment:
...
+# configure Connaisseur service
 service:
   type: ClusterIP
   port: 443
 
-# configure access to the notary server
+### VALIDATORS ###
+# validators are a set of configurations (hosts, types, public keys) that can be
+# used for validating one or multiple images (or image signatures). they are tied
+# to their respective image(s) via the image policy below. There are a few handy
```
Starkteetje, line 26:
```suggestion
# to their respective image(s) via the image policy below. there are a few handy
```

helm/values.yaml
```diff
@@ -47,66 +55,56 @@ validators:
...
   - name: docker_official
     key: |
       -----BEGIN PUBLIC KEY-----
       MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEOXYta5TgdCwXTCnLU09W5T4M4r9f
       QQrqJuADP6U7g5r9ICgPSmZuRHP/1AYUfOQW3baveKsT969EfELKj1lfCA==
       -----END PUBLIC KEY-----
-  # this is the public key to the `securesystemsengineering` repo on dockerhub
-  - name: sse_official
+  # public key securesystemsengineering repo including Connaisseur images
+  # !this key is critical for connaisseur!
```
Starkteetje, line 72:
```suggestion
  # !this key is critical for Connaisseur!
```

helm/values.yaml
```diff
@@ -47,66 +55,56 @@ validators:
...
+# and logged. this allows testing the functionality without
 # interrupting operation.
-detection_mode: false
+detectionMode: false
 
-# A list of namespaces that will be subject to Connaisseur verification.
-# If the list contains '*' - all namespaces will be monitored.
-targetNamespaces: ['*']
+# namespaced validation allows to restrict the namespaces that will be subject to Connaisseur verification.
+# When enabled, based on namespaced validation mode ('ignore' or 'validate')
```
Starkteetje, line 102:
```suggestion
# when enabled, based on namespaced validation mode ('ignore' or 'validate')
```

image_ls
```diff
@@ -0,0 +1,58 @@
+REF
```
annekebr, line 1: What's this? Is that intended?

tests/integration/preconfig-integration-test.sh
```diff
@@ -0,0 +1,54 @@
+#! /bin/bash
+set -euo pipefail
+
+# This script is expected to be called from the root folder of Connaisseur
+
+echo 'Preparing Connaisseur config...'
+yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' helm/values.yaml tests/integration/preconfig-update.yaml
+yq e '.policy +={"pattern": "docker.io/securesystemsengineering/connaisseur:helm-hook-*"} | .policy[4].pattern style="double"' -i helm/values.yaml
```
annekebr, line 8: What does this do? Would it be possible to integrate it into the `preconfig-update.yaml` file?

xopham: this excludes the helm-hook from validation, since a new built is not signed and thus uninstallation would fail. If you write that into the `preconfig-update.yaml` we overwrite all other policies or would have to add all and at that point it is likely that test and actual preconfig diverge...

tests/integration/namespaced-integration-test.sh
```diff
@@ -3,16 +3,69 @@ set -euo pipefail
...
+
+if [[ ! "$(cat output.log)" =~ 'Unable to find signed digest for image docker.io/securesystemsengineering/testimage:unsigned.' ]]; then
+  echo 'Failed to deny unsigned image or failed with unexpected error. Output:'
+  cat output.log
+  exit 1
+else
+  echo 'Successfully denied usage of unsigned image'
+fi
+
+echo 'Testing signed image in unlabelled namespace...'
```
annekebr, line 33: I might be wrong trying to wrap my head around what needs to be tested here, but doesn't it make sense to test whether an unsigned image would be denied in an unlabelled namespace here?

annekebr (approved):
Nice one! Good work!

phbelitz (approved):
mission accomplished

PR #168, opened 2021-06-05
Fix/switch to helm install

xopham:
PR is rebased on #164 and should only be merged after rebasing and merging #164 

also once this is merged, we need to set the integration test to 'required'


xopham:
old and busted. moved changes to #172

PR #169, opened 2021-06-11
refactor(nv1): parallel trust data request

phbelitz:
A major amount of connaisseur's runtime (for nv1 validation) is wasted with receiving the trust data from the notary server. To reduce this time, the requests have been parallelized.

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -57,7 +57,8 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install packages
-        run: pip3 install -r requirements_dev.txt
+        # Since we run inside an alpine based container, we cannot compile yarl and multidic
+        run: YARL_NO_EXTENSIONS=1 MULTIDICT_NO_EXTENSIONS=1 pip3 install -r requirements_dev.txt
       - name: Lint
         run: cd connaisseur && pylint --ignore-patterns=tests,coverage *.*
 
@@ -102,6 +103,9 @@ jobs:
     runs-on: ubuntu-latest
     container:
       image: python:alpine
+    env:
+      YARL_NO_EXTENSIONS: 1
+      MULTIDICT_NO_EXTENSIONS: 1
     steps:
       - uses: actions/checkout@v2
       - uses: ./.github/actions/safety
--- a/connaisseur/validators/notaryv1/notary.py
+++ b/connaisseur/validators/notaryv1/notary.py
@@ -1,11 +1,14 @@
+import json
 import os
 import re
+import ssl
 from urllib.parse import quote, urlencode
 import requests
 import yaml
+import aiohttp
 from connaisseur.image import Image
-from connaisseur.validators.notaryv1.tuf_role import TUFRole
 from connaisseur.validators.notaryv1.trust_data import TrustData
+from connaisseur.validators.notaryv1.tuf_role import TUFRole
 from connaisseur.exceptions import (
     UnreachableError,
     NotFoundException,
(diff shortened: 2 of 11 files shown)


connaisseur/validators/notaryv1/notary.py
```diff
@@ -1,11 +1,14 @@
+import json
 import os
 import re
+import ssl
```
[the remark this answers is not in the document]
Starkteetje: Same for the other files

connaisseur/validators/notaryv1/notaryv1_validator.py
```diff
@@ -111,8 +115,13 @@ def __process_chain_of_trust(
...
 
         # load all trust data
-        for role in tuf_roles:
-            trust_data[role] = self.notary.get_trust_data(image, TUFRole(role))
+        t1 = dt.datetime.now()
+        trust_data_list = await asyncio.gather(
+            *[self.notary.get_trust_data(image, TUFRole(role)) for role in tuf_roles]
+        )
+        t2 = (dt.datetime.now() - t1).total_seconds()
+        logging.debug(f"TIMER: {t2}")
```
Starkteetje, line 123: If you want to leave the debug stmt in here, I'd make it more specific: `Pulled trust data for image {image} in {t2} seconds` or the like

connaisseur/validators/notaryv1/notaryv1_validator.py
```diff
@@ -237,19 +246,28 @@ def __search_image_targets_for_tag(trust_data: dict, image: Image):
...
+            *[
+                self.notary.get_delegation_trust_data(image, TUFRole(delegation))
+                for delegation in delegations
+            ]
+        )
+
+        # when delegations are added to the repository, but weren't yet used for signing,
+        # the delegation files don't exist yet and are `None`. in this case validation
+        # must be skipped
+        delegation_trust_data = {
```
Starkteetje, line 262: This is subtly different from the previous version. Previous `delegation_trust_data` that were `None` were included in `trust_data` while not being validated. Now they are not included and not validated

Haven't really thought this through and maybe you see this more easily. Does that make a difference?

Starkteetje: Also the comment is now wrong

.github/workflows/cicd.yaml
```diff
@@ -57,7 +57,7 @@ jobs:
     steps:
       - uses: actions/checkout@v2
       - name: Install packages
-        run: pip3 install -r requirements_dev.txt
+        run: YARL_NO_EXTENSIONS=1 MULTIDICT_NO_EXTENSIONS=1 pip3 install -r requirements_dev.txt
```
Starkteetje, line 60:
```suggestion
        # Since we run inside an alpine based container, we cannot compile yarl and multidict
        # See https://pypi.org/project/yarl/ and https://github.com/aio-libs/multidict
        run: YARL_NO_EXTENSIONS=1 MULTIDICT_NO_EXTENSIONS=1 pip3 install -r requirements_dev.txt
```
would add this as a comment to explain; they both mention performance losses of 20-50x (not % !), so maybe we'd want to benchmark that and if grave, switch to non-alpine?

Starkteetje: Probably same for Dockerfile

connaisseur/validators/notaryv1/notary.py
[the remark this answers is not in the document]
phbelitz: I think we should do a "tidy up" PR where we delete any unused imports and take a closer look at `pylint` as it currently only scans the files in `./connaisseur` but not its subdirectories :shrug:

connaisseur/validators/notaryv1/notaryv1_validator.py
phbelitz: now that i think about it. currently we are downloading ALL delegation trust data, whether we need them or not. i think we should change and only download those we actually need. then wen can also raise an error here, should an required delegation trust data be `None`

.github/workflows/cicd.yaml
phbelitz: @annekebr tested the performance with a `python:slim` image and there is no difference, so i think we can keep the alpine

connaisseur/validators/notaryv1/notaryv1_validator.py
Starkteetje: Sounds reasonable and would also prevent DoS-like attack of spamming trust data. Can't see anything wrong with the idea 👍

phbelitz: #171 will be done in a seperate PR

Starkteetje (approved):

PR #170, opened 2021-06-25
Set side effect to None so server-side dry run is possible

pcallewaert:
👋 

I had troubles deploying a terraform workflow, that executed a server side dry-run:

```txt
...: patch failed 'application/strategic-merge-patch+json': admission webhook "connaisseur-svc.connaisseur.svc" does not support dry run
```
Reading through these [docs](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#side-effects), I've updated the webhook config to have `sideEffects: None` (defaults to Unknown), and this solved the problem.

I'm not sure if `None` is the correct value, but I don't think in the case of Connaisseur it is a problem.

### DIFF
--- a/helm/templates/webhook/webhook.yaml
+++ b/helm/templates/webhook/webhook.yaml
@@ -19,9 +19,10 @@ webhooks:
         apiGroups: ["*"]
         apiVersions: ["*"]
         resources: ["pods", "deployments", "replicationcontrollers", "replicasets", "daemonsets", "statefulsets", "jobs", "cronjobs"]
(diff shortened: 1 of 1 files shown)


xopham:
Thanks @pcallewaert ! Good catch.
Connaisseur should not cause any out-of-band changes to the cluster and thus `sideEffects: None` should be correct.

Had to change the base branch for merging and sign the commit.

xopham (approved):
lgtm! Connaisseur should have no `side-effects`.

PR #172, opened 2021-07-16
Refactor/switch to helm install

xopham:
using `helm install` is just more modern than make...and while handy during dev, users benefit if we test with their tools


phbelitz (approved):
:amphora:

PR #174, opened 2021-07-19
fix: auth config schema

xopham:
config schema for auth required "user" attribute while the actual attribute is "username"


phbelitz (approved):
:+1:

PR #183, opened 2021-07-23
Fix get root utility

Starkteetje:
Previously, building the get-root utility failed due to a new requirement of the cryptography package, that now requires a rust compiler on alpine-based distributions. This commit adds the rust compiler to fix it. Also, the utility was out of sync with the changes to the notary interface, which is also rectified by this commit. Lastly, it fixes a faulty reference in the docs

### DIFF
--- a/.github/workflows/nightly-scans.yaml
+++ b/.github/workflows/nightly-scans.yaml
@@ -61,3 +61,16 @@ jobs:
         with:
            name: trivy-reports
            path: trivy-reports
+
+  get-root:
+    runs-on: ubuntu-latest
+    container:
+      image: docker:stable
+    steps:
+      - uses: actions/checkout@v2
+      - name: Build and test get-root utility
+        run: |
+          docker build -t get-root-key -f docker/Dockerfile.getRoot .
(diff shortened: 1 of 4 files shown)


peterthomassen:
In my experience, adding the Rust compiler to an Alpine image blows up its size enormously. It may be justified to switch to a Debian-based base image until pre-compiled cryptography for Alpine is available. (It is not, because Alpine's glibc variant, musl, requires a build chain that pypi.org currently doesn't support, but may be will in the future.)

Starkteetje:
> 
> 
> In my experience, adding the Rust compiler to an Alpine image blows up its size enormously. It may be justified to switch to a Debian-based base image until pre-compiled cryptography for Alpine is available. (It is not, because Alpine's glibc variant, musl, requires a build chain that pypi.org currently doesn't support, but may be will in the future.)

You're right. Size is > 700MB by now. Since it's not main image that's being deployed, it wouldn't be critical, but I also had reservations due to build time, so maybe I switch over immediately instead of leaving that as another task

phbelitz (approved):

PR #185, opened 2021-07-23
refactor(Makefile): make annihilate deletes everything

phbelitz:
Previously `make annihilate` didn't delete the image policy CRD and the connaisseur namespace. Now it does, which allows to freshly reinstall connaisseur eventhough changes to the image policies format were made.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -31,4 +31,5 @@ upgrade:
 	helm upgrade connaisseur helm --wait
 
 annihilate:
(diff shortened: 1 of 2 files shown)


xopham (approved):
lgtm

Starkteetje:
It's not a refactor as it changes behaviour. Otherwise, lgtm

xopham:
Could you rename it, since it is not a refactor

PR #193, opened 2021-07-26
fix: image loading for release workflow (#192)

phbelitz:
Fixed the release workflow, so it actually can load the Connaisseur images from DockerHub.


annekebr (approved):
top

PR #196, opened 2021-07-27
fix: nightly scans (#195)

phbelitz:
With the new dependencies from asyncio, the nightly scans fail as they run on an alpine based container which can't install the dependencies without special enviornment variables. This was fix by switching to a debian based container.


xopham (approved):
lgtm

PR #208, opened 2021-07-30
ci: add dependabot config

annekebr:
Dependabot will open MRs as soon as one of our pip dependencies is outdated.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,33 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "update"
+    insecure-external-code-execution: "deny"
+    target-branch: "develop"
+  - package-ecosystem: "pip"
+    directory: "/docs"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "update"
+    insecure-external-code-execution: "deny"
+    target-branch: "develop"
+  - package-ecosystem: "docker"
+    directory: "/docker"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "update"
+    target-branch: "develop"
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "update"
+    target-branch: "develop"
+


.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
+    directory: "/"
+    schedule:
+      interval: "daily"
+    commit-message:
+      prefix: "update"
```
xopham, line 8:
```suggestion
      prefix: "update:"
```

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
```
xopham, line 3: what about `docker`, `git submodules` (?), `github actions`?

.github/dependabot.yml
```diff
@@ -0,0 +1,11 @@
...
+    directory: "/"
```
xopham, line 4: will this check subfolders and find `/docs/requirements_docs.txt`? does it find `requirements.txt` at all?

xopham (commented):
neat! like the idea. Did you test it in another repo adding outdated dependencies?
wondering how dependabot handles e.g. our version pinning

annekebr:
image
Tested it in my own repo - looks like that :)

.github/dependabot.yml
annekebr: For `docs` we need an additional package manager entry

.github/dependabot.yml
annekebr: can add `docker` and `github-actions`, why `git submodules`?

.github/dependabot.yml
annekebr: The colon is added automatically :)

.github/dependabot.yml
xopham: didn't check that. the first 2 should suffice

.github/dependabot.yml
xopham: because of the path? that could also be changed...

annekebr: and yes, it finds `requirements.txt`, but not requirements_dev.txt`. I think we can accept that, what do you think? 

(Other people are having the same question though: https://github.community/t/multiple-requirements-file-in-dependabot/190134)

annekebr: added additional package manager

xopham (dismissed):
lgtm

phbelitz (approved):

PR #209, opened 2021-07-30
fix: oci image spec

### LINKED ISSUE #194: Image format validation seems to reject valid images
Actually, I think you have a bug in your 1.5 code because even with this setting in values.yml:
_policy:
  \- pattern: "*"
    verify: false_
it still denies the request:
_denied the request: "master-node:5000/k8s.gcr.io/kube-apiserver:v1.18.6" is not a valid image format._

I think the double ":" throws your code off.

I see 2.0 is out today, so this bug might be unimportant now.

_Originally posted by @apopaa in https://github.com/sse-secure-systems/connaisseur/discussions/178#discussioncomment-1058490_

phbelitz:
The image reference validation now follows the OCI image specification, so that valid image do not get denied by connaisseur.

fix #194


annekebr (approved):
Nice!

PR #235, opened 2021-08-06
performance: Fix order of webhook and sentinel probes during readiness probe

### LINKED ISSUE #233: Connaisseur could respond ready twice as fast and with half traffic
**Describe the bug**
Readiness probe always checks sentinel first and then webhook, so even way after installation it will query k8s api twice instead of once (could succeed early after webhook check)

Starkteetje:
This commit changes the order of the readiness endpoint checking for webhook and sentinel. During normal operations the webhook will be accessible in which case the pod can immediately respond ready. Prior to this commit the sentinel was asked first every time.

Fix #233


xopham (approved):
Lgtm

PR #146, opened 2021-04-19
ci:add kube-linter

xopham:
KubeLinter is a SAST tool to scan Kubernetes YAML and Helm Charts: https://github.com/stackrox/kube-linter
The documentation can be found here: https://docs.kubelinter.io/

The following questions arise:
* KL is currently in early development and we have to see whether that is okay for us
* Check the findings
* Does it provide value? Yes. It actually found bugs in our charts.

Open Questions if we want to proceed:
- [x] reasonable resource limits for **connaisseur**
- [x] reasonable resource limits for **hook**
- [x] reasonable resource limits for **sentinel**
- [x] solid security settings for **connaisseur**
- [x] solid security settings for **hook**
- [x] solid security settings for **sentinel**
- [x] connaisseur pod affinity for **connaisseur**
  - [x] test pod affinity
- [x] fix service account failure for **hook**

To see the results, check the GH Action.

**For Review:**
- Please closely review and extend security settings. We should configure this very rigorously.
- Should we bump *MINOR* Version?

PR #255, opened 2021-08-11
fix: make upgrade

### LINKED ISSUE #165: use 'startup probes' instead of sentinel
**Describe the feature**
A `startup probe` could be used to manage liveliness and healthiness during Connaisseur startup: https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/#define-startup-probes

The sentinel was introduced to handle the startup process of Connaisseur in order to manage liveliness and readiness before webhook and others are setup. A startup probe may be the more native solution.

### LINKED ISSUE #181: Fix helm upgrade / Connaisseur Update
**Describe the bug**
When modifying the configuration via `helm/values.yaml` and attempting to update Connaisseur via:

```bash
helm upgrade connaisseur helm --wait
```

Connaisseur performs the update and restarts properly. However, when attempting to run an image, the webhook fails to call the service:

```bash
Error: warning: Hook pre-delete connaisseur/templates/webhook/job.yaml failed: Internal error occurred: failed calling webhook "connaisseur-svc.connaisseur.svc": Post "https://connaisseur-svc
.connaisseur.svc:443/mutate?timeout=30s": x509: certificate signed by unknown authority (possibly because of "x509: invalid signature: parent certificate cannot sign this kind of certificate"
 while trying to verify candidate authority certificate "connaisseur-svc.connaisseur.svc")
```

This should either be fixed or a suitable way to update connaisseur configuration should be devised.

**Expected behavior**
Configuration should be updateable via an update mechanism and not via `uninstall` -> `install`.

Also see: #178

phbelitz:
In ADR 1 we discussed the problem of Connaisseurs installation order and why we needed a bootstraping pod and the helm-hook image. This isn't necessary anymore. Now the webhook is part of the normal Connaisseur chart and won't be installed via the helm-hook image. Instead, an empty webhook configuration is applied in the normal installation phase next to the actual Connaisseur pods, which will be filled with the actual data in the post-install stage of the helm installation. This in it's own causes the problem, that everytime Connaisseur is upgraded, the webhook will have a new certificate, while the pods won't unless they are restarted. This is also fixed by taking an already existing certificate by using the helm `lookup` function and thus not changing the actual certificate.

fix #181, fix #165


xopham (changes requested):
Doesn't work for me. I get into a loop where new pods get created and instantly terminated again.

Some further notes:
* we should include the functionality in one of the integration tests (maybe one that already uninstalls and re-installs connaisseur)
* please update the docs
* after #146 , we should be able to get rid of the `kubectl config set-context --current --namespace $(NAMESPACE)` and only reference the namespace in the helm command

xopham (changes requested):
Really awesome changes!! :partying_face: 
A few notes:
- The 5min notice in Makefile could be removed
- switch from admissionregistration v1beta1 to v1: https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#webhook-request-and-response
- don't see why the certificate_webhook-conf.yaml is one file and the name is really off

helm/templates/certificate_webhook-conf.yaml
```diff
@@ -13,19 +13,98 @@ metadata:
...
+      service:
+        name: {{ .Chart.Name }}-svc
+        namespace: {{ .Release.Namespace }}
+        path: /mutate
+      caBundle: {{ default ($certs.Cert | b64enc) (include "getInstalledTLSCert" .) }}
+    rules: []
+    sideEffects: None
+    admissionReviewVersions: ["v1beta1"]
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
```
xopham, line 42: version

helm/templates/certificate_webhook-conf.yaml
```diff
@@ -13,19 +13,98 @@ metadata:
...
+    failurePolicy: Fail
+    clientConfig:
+      service:
+        name: {{ .Chart.Name }}-svc
+        namespace: {{ .Release.Namespace }}
+        path: /mutate
+      caBundle: {{ default ($certs.Cert | b64enc) (include "getInstalledTLSCert" .) }}
+    rules: []
+    sideEffects: None
+    admissionReviewVersions: ["v1beta1"]
```
xopham, line 61: version

helm/templates/certificate_webhook-conf.yaml
```diff
@@ -13,19 +13,98 @@ metadata:
...
+---
+apiVersion: admissionregistration.k8s.io/v1beta1
```
xopham, line 63: version

helm/templates/certificate_webhook-conf.yaml
```diff
@@ -13,19 +13,98 @@ metadata:
...
   annotations:
-    helm.sh/hook: post-install, post-upgrade
-    helm.sh/hook-delete-policy: hook-succeeded, hook-failed
-data:
-  webhook.yaml: |
-    {{- include "webhooks.yaml.tpl" (dict "context" . "cert" $certs.Cert) | indent 4}}
+    "helm.sh/hook": pre-upgrade,pre-rollback
+webhooks:
+  - name: {{ .Chart.Name }}-svc.{{ .Release.Namespace }}.svc
+    failurePolicy: Fail
```
xopham, line 52: is it common to leave this at fail for upgrade? seems risky though no rule would make any call necessary. One could at this point hook in connaisseur with a `RollingUpdate` Policy and keep it functional throughout the upgrade :shrug: 
if we do with turning it off, we'd be interrupting operation

xopham (changes requested):
Further required changes:
- Need to update the administration section in the `docs/basics.md`. Check for further docs that need changing.

connaisseur/flask_server.py
```diff
@@ -86,43 +84,7 @@ def healthz():
 # readiness probe
 @APP.route("/ready", methods=["GET", "POST"])
 def readyz():
```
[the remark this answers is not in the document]
phbelitz: the `helm install --wait` will wait indefinitely as long as the conny pods are not ready. so no.

helm/templates/_helpers.tpl
```diff
@@ -134,3 +110,17 @@ limits:
     {{- end -}}
 {{- end -}}
 {{- end -}}
+
+{{- define "getInstalledTLSCert" -}}
```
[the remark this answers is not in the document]
phbelitz: yes.

helm/templates/certificate_webhook-conf.yaml
```diff
@@ -13,19 +13,98 @@ metadata:
     {{- include "helm.labels" . | nindent 4 }}
 type: Opaque
 data:
-  tls.crt: {{ $certs.Cert | b64enc }}
-  tls.key: {{ $certs.Key | b64enc }}
+  tls.crt: {{ default ($certs.Cert | b64enc) (include "getInstalledTLSCert" .) }}
+  tls.key: {{ default ($certs.Key | b64enc) (include "getInstalledTLSKey" .) }}
 ---
```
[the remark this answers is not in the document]
phbelitz: the tls-certificate and the webhook need the same certificate. otherwise conny will not work. if we want the use helm for generating said certificate, both of these (tls and webhook) need to be defined in the same file so they can access the same certificate. otherwise helm would generate 2 different certificates and thus break conny. the name comes from the combination of the two resources that are defined in there. the tls certificate and the webhook.

xopham (changes requested):
you should mention how to upgrade now in `docs/basics.md`

Starkteetje (dismissed):

docs/basics.md
```diff
@@ -93,6 +93,14 @@ kubectl run hello-world --image=docker.io/hello-world
 > pod/hello-world created
 ```
 
+### Upgrade
+
+Should you want to make some adjustments to the Connaisseur configuration while Connaisseur is already running, Helm can be used to upgrade the current release.
```
xopham, line 98:
```suggestion
The configuration of a running Connaisseur instance can be  updated by a Helm upgrade of the current release:
```
the proposed sentence is a bit  backwards and unnecessarily lengthy

Makefile
```diff
@@ -29,5 +27,5 @@ upgrade:
 	helm upgrade connaisseur helm  -n $(NAMESPACE) --wait
 
 annihilate:
-	kubectl delete all,mutatingwebhookconfigurations,clusterroles,clusterrolebindings,configmaps,imagepolicies,secrets,serviceaccounts,crds -lapp.kubernetes.io/instance=connaisseur
+	kubectl delete all,mutatingwebhookconfigurations,clusterroles,clusterrolebindings,configmaps,secrets,serviceaccounts -lapp.kubernetes.io/instance=connaisseur
```
xopham, line 30: I presume it doesn't hurt to keep it in and on the other hand allows to remove the remaining resources of earlier releases. If it does not cause an error, I'd just keep it.

helm/templates/deployment.yaml
```diff
@@ -11,6 +11,10 @@ spec:
     matchLabels:
       app.kubernetes.io/name: {{ include "helm.name" . }}
       app.kubernetes.io/instance: {{ .Chart.Name }}
+  strategy:
+    type: RollingUpdate
+    rollingUpdate:
+      maxUnavailable: 1
```
xopham, line 17: Why do we need to allow `1` unavailable? If we do so, what happens if there is only 1 replica?
Couldn't we set this to `0` and have it spinned up before anything is spinned down?

helm/values.yaml
```diff
@@ -1,7 +1,7 @@
 # configure Connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v2.1.2
+  image: securesystemsengineering/connaisseur:v2.2.0
   helmHookImage: securesystemsengineering/connaisseur:helm-hook-v1.1
```
xopham, line 5:
```suggestion
```
not needed anymore

helm/values.yaml
```diff
@@ -1,7 +1,7 @@
...
   imagePullPolicy: Always
```
xopham, line 6:
```suggestion
  imagePullPolicy: IfNotPresent
```
Dow we actually need `Allways`?

tests/integration/update.yaml
```diff
@@ -37,10 +37,6 @@ policy:
       trust_root: sse
       delegations: ["belitzphilipp", "starkteetje"]
   - pattern: "docker.io/securesystemsengineering/connaisseur:*"
-    validator: dockerhub
-    with:
-      trust_root: sse
-  - pattern: "docker.io/securesystemsengineering/connaisseur:helm-hook-*"
     validator: allow
```
xopham, line 40: I see why this is necessary here. But I would strongly recommend that we run a test that the update to the new connaisseur image works upon release in the release workflow.
Can be minimal just like:
- grep last and new release
- install last with last preconfig
- get new charts
- upgrade

xopham (changes requested):
Really like the improvement!
I am not sure if it works. Fake upgrade to a same version v2.2.1 worked. Downgrade to v2.1.2 did not work bc readiness probe failed. Will we have to make notes for upgrading for the release?

On top of my comments, please fix:
- there is several instances of the `helm-hook` remaining, e.g. in `helm/values.yaml` and some of the tests
- I suspect there will also still be some notes on it in the docs
- I would recommend to simply do a grep on the whole repo and check all instances
- please do not rebase and only fixup, as I can otherwise not track my changes

Should we wait with merging this for implementation of #280 bc currently it will be unintuitive that the config is not updated

docs/adr/ADR-5_no-more-bootstrap.md
```diff
@@ -0,0 +1,69 @@
...
+### Solution 2.3 - External TLS
+
+Go back to using an external TLS certificate which is not being generated within the Helm chart, but by pre-configuring it or using OpenSSL.
+
+**Pros:** Fastest solution.
+**Cons:** More configurational effort and/or not pipeline friendly (may need OpenSSL).
+
+### Decision outcome (2)
+
+Solution 2.1 is being implemented, as it is important that Connaisseur works with as little configuration effort from the get go as possible. Nonetheless an external configuration of TLS certificates is still considered for later development.
```
Starkteetje, line 67:
```suggestion
Solution 2.1 is being implemented, as it is important that Connaisseur works with as little configuration effort as possible from the get-go. Nonetheless an external configuration of TLS certificates is still considered for later development.
```

docs/basics.md
```diff
@@ -93,6 +93,14 @@ kubectl run hello-world --image=docker.io/hello-world
 > pod/hello-world created
 ```
 
+### Upgrade
+
+The configuration of a running Connaisseur instance can be  updated by a Helm upgrade of the current release:
```
Starkteetje, line 98:
```suggestion
The configuration of a running Connaisseur instance can be updated by a Helm upgrade of the current release:
```

helm/templates/certificate_webhook-conf.yaml
```diff
@@ -13,19 +13,76 @@ metadata:
...
+        operator: NotIn
+        values:
+          - ignore
+    {{- else if eq .Values.namespacedValidation.mode "validate"}}
+        operator: In
+        values:
+          - validate
+   {{- end }}
+   {{- end }}
+   {{- end }}
```
Starkteetje, line 88: Do we have a convention on whether there should be linebreaks at file end?

helm/values.yaml
```diff
@@ -1,9 +1,8 @@
 # configure Connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v2.1.2
-  helmHookImage: securesystemsengineering/connaisseur:helm-hook-v1.1
-  imagePullPolicy: Always
+  image: securesystemsengineering/connaisseur:v2.2.0
+  imagePullPolicy: IfNotPresent
```
Starkteetje, line 5: Why this change?

.github/workflows/release.yaml
```diff
@@ -28,10 +28,9 @@ jobs:
     strategy:
       matrix:
         integration-test-script: [
-          'integration-test.sh',
```
Starkteetje, line 31: Why delete the integration test in pipeline and still change it?

tests/integration/release-integration-test.sh
```diff
@@ -0,0 +1,238 @@
...
+declare -i NUMBER_OF_INVALID_DEPLOYMENTS=0
+
+echo 'Preparing Connaisseur config...'
+envsubst < tests/integration/release-update.yaml > update
+yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' helm/values.yaml update
+rm update
+echo 'Config set'
+
+echo 'Installing Connaisseur...'
+# 'make' is chosen deliberately below to test the Makefile, while other tests use 'helm' directly
```
Starkteetje, line 15: Now there's two tests with this disclaimer, so it's incorrect

tests/integration/release-integration-test.sh
```diff
@@ -0,0 +1,238 @@
+#! /bin/bash
```
Starkteetje, line 1: Is this test to make sure some upgrade works like xoph suggested? If so, do we need all the tests or can we reduce the amount of tests run?
Also, what do you think of writing a comment at the start of this/each integration test highlighting why it is needed in addition to the base case?

docs/adr/ADR-1_bootstrap-sentinel.md
```diff
@@ -3,6 +3,7 @@
 ## Status
 
 Amended in [ADR-3](ADR-3_multi_notary_config.md).
+Amended in [ADR-5](ADR-5_no-more-bootstrap.md).
```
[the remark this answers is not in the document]
Starkteetje: @phbelitz you missed this?

connaisseur/util.py
```diff
@@ -71,7 +71,7 @@ def get_admission_review(
         }
     """
     review = {
-        "apiVersion": "admission.k8s.io/v1beta1",
+        "apiVersion": "admission.k8s.io/v1",
```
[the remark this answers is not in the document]
xopham: I think it is possible to specify that either `v1` or `v1beta1` is fine. @phbelitz what do you think?

helm/values.yaml
xopham: why would we always pull? I think this makes sense when using the latest tag or overwriting tags, but we should not be doing that and this may improve and speedup installation of connaisseur, right?
I believe this was introduced as it is easier in our local testing (depending on how you do that).

xopham: just checked: even when using `Always` it only repulls when the digest is different. However, that will still require all setups to have access to the internet (dockerhub) and it is not enough to just push the image to the kubelet...
@Starkteetje what do you think would be preferred?

xopham: could also keep it as is, and people who need it differently can change it manually anyways

connaisseur/util.py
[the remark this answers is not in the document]
phbelitz: welp, depends on how many versions we want to support. if we say `latest - 5`, then we can savely get rid of all `v1beta`s, since it's deprecated since v1.16

xopham (commented):
made some comments

Makefile
```diff
@@ -29,5 +27,5 @@ upgrade:
 	helm upgrade connaisseur helm  -n $(NAMESPACE) --wait
 
 annihilate:
-	kubectl delete all,mutatingwebhookconfigurations,clusterroles,clusterrolebindings,configmaps,imagepolicies,secrets,serviceaccounts,crds -lapp.kubernetes.io/instance=connaisseur
+	kubectl delete all,mutatingwebhookconfigurations,clusterroles,clusterrolebindings,configmaps,secrets,serviceaccounts,crds -lapp.kubernetes.io/instance=connaisseur
```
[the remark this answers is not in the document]
phbelitz: this will create an error when the `imagepolicies` are already gone -> we can't use `make annihilate` anymore

phbelitz:
changed the pipeline, so the normal integration test is always used (for both release and develop pipeline), but when running in the develop pipeline, the Connaisseur image will be whitelisted, though a manual change in the pipeline

tests/integration/update.yaml
```diff
@@ -36,12 +36,6 @@ policy:
     with:
       trust_root: sse
       delegations: ["belitzphilipp", "starkteetje"]
-  - pattern: "docker.io/securesystemsengineering/connaisseur:*"
-    validator: dockerhub
-    with:
-      trust_root: sse
-  - pattern: "docker.io/securesystemsengineering/connaisseur:helm-hook-*"
-    validator: allow
```
xopham, line 44: this configuration seems to be bound to fail. You removed the connaisseur image pattern and only add an allow for the `cicd.yaml` which means that there is no pattern for the `release.yaml`. I would propose to keep the entry with `allow` entry for cicd and overwrite the entry using `yq` to the dockerhub sse validator for release or the other way round. Only challenge is you will have to pick the correct entry by number as it is an array which will fail at some point when we change the pre-configuration...
So maybe choose option three and add an entry during the `release.yaml` as well...

xopham (commented):
I think we are almost there.
Remaining changes are:
- silently failing when trying to delete the old imagepolicy CRD
- fixing the release.yaml: most beautiful solution would be to only overwrite that specific entry for the release case, but we would have to pick the entry from an array by number...whatever the case, it needs to be fixed to work

tests/integration/update.yaml
phbelitz: i don't quite understand, the `cicd.yaml` pipeline and the `release.yaml` pipeline run independently from eachother, so changes to `cicd.yaml` do not affect `release.yaml`. and for `release.yaml`, there is still a `*:*` rule, that the connaisseur image applies to.

xopham: you are correct. did not see that the default policy is sse pub key

xopham (approved):
lgtm!

xopham (approved):
lgtm

Issue #258, opened 2021-08-12
Add PodSecurityPolicy to the Helm manifests


youssefazrak:
**Describe the bug**
Installing Connaisseur on a Kubernetes cluster that has the PodSecurityPolicy feature enabled is not working.

**Expected behavior**
It should work.

**Optional: To reproduce**
Simply deploy a K8s cluster with the psp feature enabled.

I'm deploying a fix for this issue. It simply add psp feature to the values file and both ClusterRoles modified to include the psp if needed.

youssefazrak:
Hi @xopham, indeed that is the case but as you said we are currently using it and would like to have the backward support.
Much appreciated.

PR #259, opened 2021-08-12
Add PodSecurityPolicy to the Helm manifests

youssefazrak:
Installing Connaisseur on a Kubernetes cluster that has the
PodSecurityPolicy feature enabled is not working.

This PR fixes this issue by adding the option on Helm chart to take into
acccount the PSP feature.

Fixes: #258

Signed-off-by: Youssef Azrak <yazrak.tech@gmail.com>

### DIFF
--- a/helm/templates/podsecuritypolicy.yaml
+++ b/helm/templates/podsecuritypolicy.yaml
@@ -1,0 +1,36 @@
+{{- if .Values.deployment.podSecurityPolicy }}
+{{- if .Values.deployment.podSecurityPolicy.enabled }}
+{{- if has "connaisseur-psp" .Values.deployment.podSecurityPolicy.name }}
+apiVersion: policy/v1beta1
+kind: PodSecurityPolicy
+metadata:
+  name: {{ .Chart.Name }}-psp
+  annotations:
+    seccomp.security.alpha.kubernetes.io/allowedProfileNames: runtime/default,docker/default
+    seccomp.security.alpha.kubernetes.io/defaultProfileName: docker/default
+spec:
+  privileged: false
+  allowPrivilegeEscalation: false
+  requiredDropCapabilities:
+  - ALL
+  runAsUser:
+    rule: 'MustRunAsNonRoot'
+  fsGroup:
+    rule: 'MustRunAs'
+    ranges:
+      - min: 1
+        max: 65535
+  seLinux:
+    rule: 'RunAsAny'
+  supplementalGroups:
+    rule: 'MustRunAs'
+    ranges:
+      - min: 1
+        max: 65535
+  readOnlyRootFilesystem: true
+  volumes:
+  - configMap
+  - secret
+{{- end }}
+{{- end }}
+{{- end }}
--- a/helm/templates/role.yaml
+++ b/helm/templates/role.yaml
@@ -9,3 +9,14 @@ rules:
 - apiGroups: ["*"]
   resources: ["deployments", "pods", "replicacontrollers", "replicasets", "daemonsets", "statefulsets", "jobs", "cronjobs", "imagepolicies", "mutatingwebhookconfigurations"]
   verbs: ["get"]
+{{- if .Values.deployment.podSecurityPolicy }}
+{{- if .Values.deployment.podSecurityPolicy.enabled }}
+- apiGroups: ['policy']
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+    {{- range  .Values.deployment.podSecurityPolicy.name }}
+    - {{ . | quote }}
+    {{- end }}
+{{- end}}
+{{- end}}
--- a/helm/templates/webhook/role.yaml
+++ b/helm/templates/webhook/role.yaml
@@ -18,3 +18,14 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["delete", "get", "list"]
+{{- if .Values.deployment.podSecurityPolicy }}
+{{- if .Values.deployment.podSecurityPolicy.enabled }}
+- apiGroups: ['policy']
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+    {{- range  .Values.deployment.podSecurityPolicy.name }}
+    - {{ . | quote }}
+    {{- end }}
+{{- end}}
+{{- end}}
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -26,6 +26,10 @@ deployment:
                     - connaisseur
             topologyKey: kubernetes.io/hostname
           weight: 100
+  # PodSecurityPolicy is deprecated as of Kubernetes v1.21, and will be removed in v1.25
+  podSecurityPolicy:
+    enabled: false
+    name: ["connaisseur-psp"]  # list of PSPs to use, "connaisseur-psp" is the project-provided default
 
 # configure Connaisseur service
 service:


youssefazrak:
@xopham would be great to get a review on this one.
Also, we were thinking of adding a PSP manifest to be deployed directly to the cluster. Sometimes the existing PSPs are not tailored for a specific workload. If you are ok with this, I will add the PSP manifest too.

xopham:
@youssefazrak thanks for the PR. We'll look into this and review.

helm/values.yaml
```diff
@@ -108,6 +108,11 @@ namespacedValidation:
   enabled: false
   mode: ignore  # 'ignore' or 'validate'
 
+# PodSecurityPolicy for Kubernetes cluster using this feature
+psp:
+  enabled: false
+  name:  # name of the psp
```
Starkteetje, line 114: Not deep into PSPs, but two caveats:
- PSPs have been deprecated. I think we don't mind making backwards compatible changes, but we need to make sure that they won't break the future, so we'd probably want to add a k8s version check to `{{- if .Values.psp.enabled }}` and leave it out for >=1.25 (and maybe warn for >=1.21 [@phbelitz does allow helm that?])
- Afaik, one can specify multiple PSPs, so the `name` field here should be an array, which needs changes to `- {{ .Values.psp.name }}`.

Also, would we want to check for empty names? With a single empty name, installation breaks if enabled=true, which I'd guess is what we want. But if it becomes an array, do we ignore empty lines (if valid ones exist) or still fail due to invalid k8s manifest? Would guess we choose the latter, but I guess @phbelitz is more into that

Starkteetje (dismissed):

helm/templates/role.yaml
```diff
@@ -9,3 +9,10 @@ rules:
 - apiGroups: ["*"]
   resources: ["deployments", "pods", "replicacontrollers", "replicasets", "daemonsets", "statefulsets", "jobs", "cronjobs", "imagepolicies", "mutatingwebhookconfigurations"]
   verbs: ["get"]
+{{- if .Values.psp.enabled }} 
+- apiGroups: ['policy']
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+  - {{ .Values.psp.name }}
+{{- end}}
```
xopham, line 18:
```suggestion
{{- if .Values.psp }}
{{- if .Values.psp.enabled }} 
- apiGroups: ['policy']
  resources: ['podsecuritypolicies']
  verbs:     ['use']
  resourceNames:
  - {{ .Values.psp.name }}
{{- end}}
{{- end}}
```
when implemented like this, not providing any `psp` entry in `helm/values.yaml` does not cause an error which is somewhat more fail safe.

helm/templates/webhook/role.yaml
```diff
@@ -18,3 +18,10 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["delete", "get", "list"]
+{{- if .Values.psp.enabled }}
+- apiGroups: ['policy']
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+  - {{ .Values.psp.name }}
+{{- end}}
```
xopham, line 27:
```suggestion
{{- if .Values.psp }}
{{- if .Values.psp.enabled }}
- apiGroups: ['policy']
  resources: ['podsecuritypolicies']
  verbs:     ['use']
  resourceNames:
  - {{ .Values.psp.name }}
{{- end}}
{{- end}}
```
same as in comment above

helm/values.yaml
```diff
@@ -108,6 +108,11 @@ namespacedValidation:
   enabled: false
   mode: ignore  # 'ignore' or 'validate'
 
+# PodSecurityPolicy for Kubernetes cluster using this feature
+psp:
```
xopham, line 112: I think the name should be explicit and self-explanatory as `podSecurityPolicy`.
Furthermore, I would consider PodSecurityPolicy a part of the `deployment` section at the beginning of `helm/values.yaml`, as it only really affects the actual deployment of Connaisseur itself and does not hold any value for its features. I would propose to append to the end of `deployment` as `deployment.podSecurityPolicy`.
Note that the templates above will require adjustment.

helm/values.yaml
xopham: I think I would disagree on the first point. In case someone sets a PSP in v1.25 and above, Connaisseur should not silently drop it, but actually fail. As for the thoughts on issuing a warning, I would keep that to Kubernetes itself and not inherit Kubernetes deprecation.

On the second point, I agree that it should be a list and handled as such in case a list is possible (which I think is correct). I would think that an empty list needs to be handled such that no policy is written in that case or an error is thrown. In case the list contains empty elements, I would suggest to fail if it fails.

xopham (changes requested):
@youssefazrak a very nice PR. I added some comments that would need to be resolved before merging, but see no major obstacles.

If I understand your earlier comment correctly, you would like to provide a PSP tailored to Connaisseur specifically, right? If that is the case, I would consider it a nice idea. We would just need to make sure that the policy is handled such that Connaisseur won't break by default for Kubernetes v1.25 and above. However, it could probably simply be the default value for the `PSP.name` which would make it easily applicable.

Many thanks for contributing :pray:

helm/values.yaml
```diff
@@ -14,6 +14,11 @@ deployment:
   nodeSelector: {}
   tolerations: []
   affinity: {}
+  # podSecurityPolicy for Kubernetes cluster using this feature
```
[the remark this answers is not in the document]
youssefazrak: Good idea.

helm/templates/podsecuritypolicy.yaml
```diff
@@ -0,0 +1,26 @@
+{{- if and (.Values.deployment.podSecurityPolicy) (.Values.deployment.podSecurityPolicy.enabled) (.Values.deployment.podSecurityPolicy.customManifest) }}
```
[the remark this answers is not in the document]
youssefazrak: @xopham I'm still having issues with the `{{ .Chart.Name }}-psp` part. Tried a bit of everything to get it working inside the curly brackets but still not able to get it done. For the moment I'm using a simple "connaisseur-psp" string.

helm/templates/webhook/role.yaml
```diff
@@ -18,3 +18,12 @@ rules:
 - apiGroups: [""]
   resources: ["pods"]
   verbs: ["delete", "get", "list"]
+{{- if .Values.deployment.podSecurityPolicy }}
+{{- if .Values.deployment.podSecurityPolicy.enabled }}
+- apiGroups: ['policy']
+  resources: ['podsecuritypolicies']
+  verbs:     ['use']
+  resourceNames:
+  - {{ .Values.deployment.podSecurityPolicy.name }}
```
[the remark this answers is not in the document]
youssefazrak: @xopham indeed, I pushed the changed for the list.

youssefazrak:
@xopham thanks again for the review!
Just pushed the fixes minor the comment about the `{{ }}` inside the `{{ if statement }}`.
Also rebased and followed the commit description convention.
Cheers!

xopham (changes requested):
Lgtm! Played around with it a bit, but did not do a full test. You probably know more about PSPs and tested that it works in your setup.

Made few minor last comments. After that, I think we can merge.
Will you rebase and sign it or should I?

Looking really neat. Thanks :pray:

youssefazrak:
@xopham cheers!
IIRC restricting to `MustRunAsNonRoot` caused an issue for the job, hence my less restrictive approach. I will play with it a bit and try to optimize as much as possible.
Regarding the rebasing, I will take care of it no worries. From my understanding, we are rebasing against the `develop` branch right?
Thanks again xopham, much appreciated!

xopham (commented):
@youssefazrak correct. We rebase against `develop`.
Concerning the rules: if it doesn't work, feel free to drop it. Are you using Openshift or OKD? I think Openshift set its own user and group and therefore doesn't like specifying those. We had another PR recently that tackled this topic: https://github.com/sse-secure-systems/connaisseur/pull/288/files

youssefazrak:
@xopham well, working fine now. Seems like [this ](https://github.com/sse-secure-systems/connaisseur/commit/7db4d2d7eb8663187c316b3206cfc030259a1883#diff-51370d6a2e956f2584c34242a516829ded5f1768a3fb4ce60f588ffe99d15679 )fixed the issue regarding the webhook job requirements for root privileges.
I pushed the fixes and rebased against the `develop` branch. Let me know if anything else is needed. Thanks!

xopham (approved):
lgtm!
Really nice work and collaboration. Thanks for contributing :pray:

youssefazrak:
@xopham pleasure mate :)

PR #264, opened 2021-08-13
tests: Add labels to integration test resources

Starkteetje:
This commit adds the connaisseur label to all resources created during a successful integration test, such that they can be deleted with make annihilate regardless of namespace


xopham (approved):
lgtm

xopham (approved):
lgtm

PR #266, opened 2021-08-13
build: add namespaces to makefile commands

xopham:


Makefile
```diff
@@ -22,13 +22,11 @@ install:
...
 uninstall:
-	kubectl config set-context --current --namespace $(NAMESPACE)
-	helm uninstall connaisseur
+	helm uninstall connaisseur -n $(NAMESPACE)
 	kubectl delete ns $(NAMESPACE)
 
 upgrade:
-	kubectl config set-context --current --namespace $(NAMESPACE)
-	helm upgrade connaisseur helm --wait
+	helm upgrade connaisseur helm  -n $(NAMESPACE)--wait
```
Starkteetje, line 29: Do we also want to fix the NS in the integrations tests (ie https://github.com/sse-secure-systems/connaisseur/blob/master/tests/integration/integration-test.sh#L15 ) while you're at it?

xopham: even introducing make as well to test both

tests/integration/integration-test.sh
```diff
@@ -12,7 +12,7 @@ rm update
 echo 'Config set'
 
 echo 'Installing Connaisseur...'
-helm install connaisseur helm --atomic || { echo 'Failed to install Connaisseur'; exit 1; }
+make install || { echo 'Failed to install Connaisseur'; exit 1; }
```
Starkteetje, line 15: If you conciously choose a different installation method here that differs from the other ITs, I'd add a comment stating this explicitly

Starkteetje: But I like using both make and helm 👍

Starkteetje (approved):
Nice

PR #269, opened 2021-08-13
ci: fix pylint to scan connaisseur dir

### LINKED ISSUE #253: Pylint on subdirectories
**Describe the bug**
Pylint doesn't check the subdirectories for linting purposes.

**Expected behavior**
It should.

xopham:
fixes #253

### DIFF
--- a/connaisseur/validators/cosign/cosign_validator.py
+++ b/connaisseur/validators/cosign/cosign_validator.py
@@ -20,7 +20,7 @@ class CosignValidator(ValidatorInterface):
     trust_roots: list
 
     def __init__(self, name: str, trust_roots: list, **kwargs):
-        self.name = name
+        super().__init__(name, **kwargs)
         self.trust_roots = trust_roots
 
     def __get_key(self, key_name: str = None):
@@ -30,15 +30,15 @@ class CosignValidator(ValidatorInterface):
                 key["key"] for key in self.trust_roots if key["name"] == key_name
             )
         except StopIteration as err:
-            msg = (
-                'Trust root "{key_name}" not configured for validator "{notary_name}".'
-            )
+            msg = 'Trust root "{key_name}" not configured for validator "{validator_name}".'
             raise NotFoundException(
-                message=msg, key_name=key_name, notary_name=self.name
+                message=msg, key_name=key_name, validator_name=self.name
             ) from err
         return "".join(key)
 
-    def validate(self, image: Image, trust_root: str = None, **kwargs):
+    def validate(
+        self, image: Image, trust_root: str = None, **kwargs
+    ):  # pylint: disable=arguments-differ
         pub_key = self.__get_key(trust_root)
         return self.__get_cosign_validated_digests(str(image), pub_key).pop()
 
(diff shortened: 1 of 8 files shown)


connaisseur/config.py
```diff
@@ -40,6 +40,11 @@ def __init__(self):
 
         config = self.__merge_configs(config_content, secrets_config_content)
 
+        # rename 'type' to 'vtype' as 'type' is an inbuilt function
+        for validator in config:
+            if validator.get("type", None):
```
peterthomassen, line 45: `if 'type' in validator`

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -20,6 +20,7 @@ class CosignValidator(ValidatorInterface):
     trust_roots: list
 
     def __init__(self, name: str, trust_roots: list, **kwargs):
+        super().__init__(name, **kwargs)
         self.name = name
```
peterthomassen, line 24: The second is already done by the first

peterthomassen (changes requested):

peterthomassen (dismissed):
lgtm!

connaisseur/config.py
```diff
@@ -40,6 +40,11 @@ def __init__(self):
 
         config = self.__merge_configs(config_content, secrets_config_content)
 
+        # rename 'type' to 'vtype' as 'type' is a builtin function
```
peterthomassen, line 43: This appears to have consequences on the JSON schema. Is this correct?

I don't think that implementation details (like what keywords exist in the programming language used by Connaisseur) should be exposed to the JSON schema. Have you considered working around this by using `**kwargs` instead of a static `type` argument, and mimicking the required-ness of the `type` (`vtype`) argument in below methods by inspecting `kwargs`? If you did consider it, which were the reasons to not pursue it?

xopham: It does have consequences because I rename before the validation method below. Could rename afterwards and the schema itself would not be affected, but we'd still be working with `vtype` later. Might be better from a validate-before-parsing PoV and leav3 the schema intact. What do you think?
Have not considered using kwargs, but think that may feel somewhat obscure if the validator type comes out of the dark. Also not entirely sure if and how that would work. To be fair though, changing to `vtype` required supprisingly little change...

peterthomassen: I find it problematic that the schema (which requires `vtype`) does not reflect the input format because the input is modified before the schema is applied. This is going to be very confusing for anyone looking at the schema and configuration files, without looking at the code.

So, I'd definitely prefer validating the schema before making the modification.

My favorite solution would be to not modify the configuration. That's easily achieved by replacing `Validator.__new__` with
```python
    def __new__(cls, **kwargs):
        validator_type = kwargs.pop("type")
        try:
            return cls.class_map[validator_type](**kwargs)
        except KeyError:
            msg = f"{validator_type} is not a supported validator."
            raise NoSuchClassError(message=msg)  # pylint: disable=raise-missing-from
```

This way, you don't need several places to be aware of the matter.

xopham: True. Like the idea!

peterthomassen (dismissed):
lgtm!

Starkteetje (approved):

PR #273, opened 2021-08-17
ci: add trivy IaC scan

xopham:


.github/workflows/cicd.yaml
```diff
@@ -129,6 +129,37 @@ jobs:
           name: safety-report
           path: safety-report.txt
 
+  trivy-misconfiguration:
```
Starkteetje, line 132: Why `mis`-config?

Starkteetje: I see, but might not be intuitive. Maybe add a `-scanner`?

.github/workflows/cicd.yaml
```diff
@@ -129,6 +129,37 @@ jobs:
...
+          helm template helm > deployment/deployment.yaml
+      - name: Run Trivy on deployment.yaml
+        uses: aquasecurity/trivy-action@master
+        with:
+          scan-type: 'config'
+          scan-ref: 'deployment'
+          hide-progress: false
+          format: 'table'
+          exit-code: '1'
+          ignore-unfixed: true
```
Starkteetje, line 150: Think for config, there shouldn't be unfixed ones and if there were, I'd probably wanna know

Starkteetje: Not sure whether same for docker, but might well be the same

helm/templates/_helpers.tpl
```diff
@@ -46,9 +46,9 @@ capabilities:
     - ALL
 privileged: false
 readOnlyRootFilesystem: true
-runAsGroup: 2000
```
Starkteetje, line 49: Why 1000:2000 in Dockerfiles, but not in config?

xopham: trivy recommends >10000. Could try to adjust the Dockerfile :shrug:

Starkteetje: k8s sec page does 1k:2k:3k https://kubernetes.io/docs/tasks/configure-pod-container/security-context/
Would be interested in why trivy wants it higher

Starkteetje: Guess it's collision blabla, but our image isn't even clone to using the uids and guids; don't like changing existing code for "tool said pls change" reasons, but not enough to not approve

Starkteetje (dismissed):

Starkteetje (approved):

Issue #251, opened 2021-08-10
Allow dynamic adjustment of policy and validator configuration


xopham:
**Describe the feature**
Currently, a change of validators requires a restart of all pods. Ideally, this should be possible by dynamically fetching updated configuration as it does not actually change Connaisseur itself.

**Optional: Implementation ideas**
Could be done via CRDs. However, it needs a mechanism to mount updated configuration into Connaisseur.

xopham:
We started an Architecture Decision Record in #280 on how to update the configuration (`ConfigMap` vs. `CRD`). Feel free to share your opinions. We'd be happy to get some external input on this.

FYI @jackalsh @ressetkk

PR #280, opened 2021-08-20
fix: pod restart on config change

phbelitz:
Added an ADR discussing how Connaisseur should access its configuration files going forward.


docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
+# ADR 6 Dynamic Configuration
```
Starkteetje, line 1:
```suggestion
# ADR 6: Dynamic Configuration
```

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+The configuration of validators are mounted into Connaisseur as a configmap, as it it common practice in the Kubernetes ecosystem. When this configmap is upgraded, say with a `helm upgrade`, the resource itself in Kubernetes is updated accordingly, but that doesn't mean it's automatically updated inside the pods which mounted it. That only occurs once the pods are restarted and util they aren't the pods still have an old version of the configuration lingering around. This is a fairly unintuitive behavior and the reason why Connaisseur doesn't mount the image policy into its pods. Instead, the pods have access to the kube API and get the image policy dynamically from there. The same could be done for the validator configuration, but there is also another more kubernetes native solution.
```
Starkteetje, line 9:
```suggestion
The configuration of validators are mounted into Connaisseur as a configmap, as it is common practice in the Kubernetes ecosystem. When this configmap is upgraded, say with a `helm upgrade`, the resource itself in Kubernetes is updated accordingly, but that doesn't mean it's automatically updated inside the pods which mounted it. That only occurs once the pods are restarted and util they aren't the pods still have an old version of the configuration lingering around. This is a fairly unintuitive behavior and the reason why Connaisseur doesn't mount the image policy into its pods. Instead, the pods have access to the kube API and get the image policy dynamically from there. The same could be done for the validator configuration, but there is also another more kubernetes native solution.
```

Starkteetje, line 9:
```suggestion
The configuration of validators are mounted into Connaisseur as a configmap, as it it common practice in the Kubernetes ecosystem. When this configmap is upgraded, say with a `helm upgrade`, the resource itself in Kubernetes is updated accordingly, but that doesn't mean it's automatically updated inside the pods which mounted it. That only occurs once the pods are restarted and util they are the pods still have an old version of the configuration lingering around. This is a fairly unintuitive behavior and the reason why Connaisseur doesn't mount the image policy into its pods. Instead, the pods have access to the kube API and get the image policy dynamically from there. The same could be done for the validator configuration, but there is also another more kubernetes native solution.
```

Starkteetje, line 9:
```suggestion
The configuration of validators are mounted into Connaisseur as a configmap, as it it common practice in the Kubernetes ecosystem. When this configmap is upgraded, say with a `helm upgrade`, the resource itself in Kubernetes is updated accordingly, but that doesn't mean it's automatically updated inside the pods which mounted it. That only occurs once the pods are restarted and until they aren't the pods still have an old version of the configuration lingering around. This is a fairly unintuitive behavior and the reason why Connaisseur doesn't mount the image policy into its pods. Instead, the pods have access to the kube API and get the image policy dynamically from there. The same could be done for the validator configuration, but there is also another more kubernetes native solution.
```

Starkteetje, line 9: `but there is also another more kubernetes native solution.` already sounds like an evaluation and therefore shouldn't be in the intro, I'd argue

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Problem 1 - Access to Configuraiton
+
+How should Connaisseur get access to its configuration files...?
+
+### Solution 1.1 - Dynamic Access
+
+This is the same solution as currently employed for the image policy configuration. The validators will get its own CustomResourceDefinition and Connaisseur gets via RBAC access to this resource so it can use the kube API to read the configuration.
```
Starkteetje, line 17:
```suggestion
This is the same solution as currently employed for the image policy configuration. The validators will get their own CustomResourceDefinition and Connaisseur gets access to this resource via RBAC so it can use the kube API to read the configuration.
```

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+
+**Pros:** Pods don't need to be restarted and the configuration can be changed "on the fly", without using helm.
```
Starkteetje, line 19:
```suggestion
**Pros:** Pods don't need to be restarted and the configuration can be changed "on the fly", without using Helm.
```

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+**Cons:** Not a very Kubernetes native approach and Connaisseur must always do some network requests to access its config.
+
+### Solution 1.2 - Restart Pods
+
+The other solution would be to use ConfigMaps for validators and image policy, as it is supposed to be and then restart the pods, once there were changes in the configurations. This can be achieved by setting the hash of the config files as annotations into the deployment. If there are changes in the configuration, the hash will change and thus a new deployment will be rolled out as it has a new annotation.
```
Starkteetje, line 24: "as it is supposed to be" is an evaluation and not description of the solution possibility

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+
+**Pros:** Kubernetes native and no more CustomResourceDefinitions!
+**Cons:** No more "on the fly" changes.
+
+## Problem 2 - How many configmaps are to many?
```
Starkteetje, line 29:
```suggestion
## Problem 2 - How many configmaps are too many?
```

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+
+When both the image policy and validator configurations are either CustomResourceDefinitions or ConfigMaps, is there still a need to separate them or can they be merged into one file?
+
+### Solution 2.1 - 2 Concerns, 2 Resources
+
+There will be 2 Resources, one for the image policy and one for the validators.
```
Starkteetje, line 35:
```suggestion
There will be 2 resources, one for the image policy and one for the validators.
```

annekebr:
1.2 and 2.2. :rocket:

docs/adr/ADR-6_dynamic-config.md
phbelitz: https://github.com/sse-secure-systems/connaisseur/pull/255 already has adr 5

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+The configuration of validators are mounted into Connaisseur as a configmap, as it it common practice in the Kubernetes ecosystem. When this configmap is upgraded, say with a `helm upgrade`, the resource itself in Kubernetes is updated accordingly, but that doesn't mean it's automatically updated inside the pods which mounted it. That only occurs once the pods are restarted and util they aren't the pods still have an old version of the configuration lingering around. This is a fairly unintuitive behavior and the reason why Connaisseur doesn't mount the image policy into its pods. Instead, the pods have access to the kube API and get the image policy dynamically from there. The same could be done for the validator configuration, but there is also another more kubernetes native solution.
+
+## Problem 1 - Access to Configuraiton
```
Starkteetje, line 11:
```suggestion
## Problem 1 - Access to configuration
```

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+
+How should Connaisseur get access to its configuration files...?
+
+### Solution 1.1 - Dynamic Access
```
Starkteetje, line 15:
```suggestion
### Solution 1.1 - Dynamic access
```

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+
+This is the same solution as currently employed for the image policy configuration. The validators will get its own CustomResourceDefinition and Connaisseur gets via RBAC access to this resource so it can use the kube API to read the configuration.
+
+**Pros:** Pods don't need to be restarted and the configuration can be changed "on the fly", without using helm.
+**Cons:** Not a very Kubernetes native approach and Connaisseur must always do some network requests to access its config.
+
+### Solution 1.2 - Restart Pods
```
Starkteetje, line 22:
```suggestion
### Solution 1.2 - Restart pods
```

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+The other solution would be to use ConfigMaps for validators and image policy, as it is supposed to be and then restart the pods, once there were changes in the configurations. This can be achieved by setting the hash of the config files as annotations into the deployment. If there are changes in the configuration, the hash will change and thus a new deployment will be rolled out as it has a new annotation.
+
+**Pros:** Kubernetes native and no more CustomResourceDefinitions!
+**Cons:** No more "on the fly" changes.
+
+## Problem 2 - How many configmaps are to many?
+
+When both the image policy and validator configurations are either CustomResourceDefinitions or ConfigMaps, is there still a need to separate them or can they be merged into one file?
+
+### Solution 2.1 - 2 Concerns, 2 Resources
```
Starkteetje, line 33:
```suggestion
### Solution 2.1 - 2 concerns, 2 resources
```

phbelitz:
1.2 and 2.2 👺

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+
+## Status
+
+Proposed
+
+## Context
+
+The configuration of validators are mounted into Connaisseur as a configmap, as it is common practice in the Kubernetes ecosystem. When this configmap is upgraded, say with a `helm upgrade`, the resource itself in Kubernetes is updated accordingly, but that doesn't mean it's automatically updated inside the pods which mounted it. That only occurs once the pods are restarted and until they are the pods still have an old version of the configuration lingering around. This is a fairly unintuitive behavior and the reason why Connaisseur doesn't mount the image policy into its pods. Instead, the pods have access to the kube API and get the image policy dynamically from there. The same could be done for the validator configuration, but there is also another solution.
+
+## Problem 1 - Access to configuraiton
```
Starkteetje, line 11:
```suggestion
## Problem 1 - Access to configuration
```

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
...
+### Solution 1.1 - Dynamic access
+
+This is the same solution as currently employed for the image policy configuration. The validators will get their own CustomResourceDefinition and Connaisseur gets access to this resource via RBAC so it can use the kube API to read the configuration.
+
+**Pros:** Pods don't need to be restarted and the configuration can be changed "on the fly", without using Helm.
+**Cons:** Not a very Kubernetes native approach and Connaisseur must always do some network requests to access its config.
+
+### Solution 1.2 - Restart pods
+
+The other solution would be to use ConfigMaps for validators and image policy and then restart the pods, once there were changes in the configurations. This can be achieved by setting the hash of the config files as annotations into the deployment. If there are changes in the configuration, the hash will change and thus a new deployment will be rolled out as it has a new annotation. This corresponds to the [suggestion](https://helm.sh/docs/howto/charts_tips_and_tricks/#automatically-roll-deployments) made by Helm.
```
xopham, line 24: I like this approach, but would want to combine it with `rollingupdate`. Seems sort of disadvantageous that every time the config changes, we disarm the webhook, turn the validation off for a moment and then turn back on.
If we could find a more continuous strategy for updating at least: like use a [rolling update](https://kubernetes.io/docs/concepts/workloads/controllers/statefulset/#update-strategies) strategy, don't disarm the webhook and run down the old pods waiting for the new ones to be up...

Also: if I was to change it manually, that would not update the annotation and consequently no restart, right?

xopham (commented):
I like 1.2 and 2.2 as well because they feel clean and native.

I dislike about 1.2 that we get service interruptions and those are quite significant if that happens every time connaisseur is patched with anything during a deployment time.
It seems we have external votes in favor of CRDs:
* #251 
* #278 

Would give this a week to see if we get further opinions on it.

Two more notes:
- adjust the title to a meaningful input like `docs: ADR 6 - Dynamic Configuration`
- add the ADR to `mkdocs.yml`

xopham (commented):
lgtm! Since `RollingUpdate` is introduced in #255 , I think this can be merged and implemented, at least as an initial solution. In case, pod restart is to be avoided in the future that can be done separately. For now, this will make `helm upgrade` functional and we should move forward with it.

Please, resolve my two notes in [my previous comment](https://github.com/sse-secure-systems/connaisseur/pull/280#pullrequestreview-735189150) and we can approve and merge it.

docs/adr/ADR-6_dynamic-config.md
```diff
@@ -0,0 +1,39 @@
+# ADR 6: Dynamic Configuration
+
+## Status
+
+Proposed
```
Starkteetje, line 5: If this already is fixed with the PR, change status and add outcome decisions

Starkteetje:
PR name is misleading as this does way more than add ADR6

Makefile
```diff
@@ -29,5 +29,5 @@ upgrade:
 	helm upgrade connaisseur helm  -n $(NAMESPACE) --wait
 
 annihilate:
-	kubectl delete all,mutatingwebhookconfigurations,clusterroles,clusterrolebindings,configmaps,imagepolicies,secrets,serviceaccounts,crds -lapp.kubernetes.io/instance=connaisseur
+	kubectl delete all,mutatingwebhookconfigurations,clusterroles,clusterrolebindings,configmaps,secrets,serviceaccounts,crds -lapp.kubernetes.io/instance=connaisseur
```
xopham, line 32: this change was already made in #255 and I think we should not do this for some time to ensure smooth upgrading to the latest version. It should not fail in case there is no imagepolicy, right?

xopham (commented):
minor comment

Makefile
phbelitz: it does fail :crying_cat_face:

xopham: as discussed...might be fixed by deleting separately with silent failing

xopham: and it is still a duplicate and will cause a rebase issue

xopham (approved):
lgtm

xopham (approved):
lgtm

PR #288, opened 2021-08-26
Openshift fix

pflaeging:
After this small fix connaisseur runs as a normal helm deployment inside OKD / OpenShift 4 (tested with OKd 4.7).

Only set the value: `openshift: true` in your value.yaml file.

## Reason

Openshift and OKD are using secure container by default (randomized user and group execution). So we have to disable the fixed user and group setting. This let OpenShift decide ;-)

### DIFF
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -18,10 +18,9 @@ spec:
       labels:
         app.kubernetes.io/name: {{ include "helm.name" . }}
         app.kubernetes.io/instance: {{ .Chart.Name }}
-      {{- /* This will be set automatically by Kubernetes 1.19 when using the securityContext.seccompProfile  */ -}}
-      {{- if lt (.Capabilities.KubeVersion.Minor | int) 19 }}
+      {{- if .Values.deployment.annotations }}
       annotations:
-        seccomp.security.alpha.kubernetes.io/pod: runtime/default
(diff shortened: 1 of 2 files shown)


pflaeging:
> Thanks for contributing @pflaeging ! You are pointing out an important compatibility issue with the secure defaults.
> 
> I think your last sentence almost nails it: we should just let the user decide ;-)
> I'd propose to simply expose the whole security context and annotation (comment the annotation and make a note for k8s <v1.18) via `helm/values.yaml` under `deployment.` before `resources`. Would that work for you?

Jepp, that would be OK for me.

> 
> Also:
> 
> * changed the base branch to `develop` (merged to `master` for releases)
> * we should rename the commit for [semantic commits](https://sse-secure-systems.github.io/connaisseur/v2.1.0/CONTRIBUTING/#semantic-and-conventional-commits), maybe `feat: expose security context and annotations in values.yaml`
> * have you tested on Openshift / OKD? We should add a note in the docs then.

The patched version is running in an OpenShift 4.6 cluster and in an OKD 4.7 cluster without problems (I'm only checking specific namespaces!).

> 
> Let me know if I should support!

Thx ;-)

:peter

xopham (dismissed):

phbelitz (dismissed):

phbelitz (approved):

PR #298, opened 2021-08-29
ci: connaisseur upgrade integration test

xopham:
Fixes no issue

## Description

- implements an Connaisseur upgrade integration test that will help highlight breaking changes (for upgrade)
- Connaisseur is installed in the `master` version and then upgraded to the `branch`
- integration tests confirm the proper working of Connaisseur

## Checklist

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -271,4 +271,57 @@ jobs:
         run: |
           bash tests/integration/workload-objects-integration-test.sh
         shell: bash
+  
+  upgrade-test:
+    runs-on: ubuntu-latest
+    needs: [build]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Install yq and bash
+        run: |
+          sudo snap install yq
+          sudo apt update
+          sudo apt install bash -y
+        shell: bash
+      - uses: actions/download-artifact@v2
+        with:
+          name: images
+      - name: Load Docker images
+        run: |
+          docker load -i ${GITHUB_SHA}_image.tar
+        shell: bash
+      - uses: ./.github/actions/k8s-version-config
+        name: Setup k8s cluster
+        with:
+          k8s-version: v1.22
+      - uses: actions/checkout@v2
+        with:
+          ref: 'master'
+      - name: Configure Connaisseur (master)
+        run: |
+          yq e '.detectionMode=true' -i helm/values.yaml
+      - name: Install Connaisseur (master)
+        run: |
+          make install
+      - name: Get image name & version (master)
+        run: |
+          kubectl get pods -n connaisseur -o jsonpath="{.items[*].spec.containers[*].image}"
+      - name: Run integration tests (master)
+        run: |
+          bash tests/integration/upgrade-integration-test.sh
+        shell: bash
+      - uses: actions/checkout@v2
+      - name: Upgrade Connaisseur to current branch
+        run: |
+          make upgrade
+      - name: Get image name & version (branch)
+        run: |
+          kubectl get pods -n connaisseur -o jsonpath="{.items[*].spec.containers[*].image}"
+      - name: Run integration tests (branch)
+        run: |
+          bash tests/integration/upgrade-integration-test.sh
+        shell: bash
+      - name: Uninstall Connaisseur
+        run: |
+          make uninstall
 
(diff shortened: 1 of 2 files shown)


.github/workflows/cicd.yaml
```diff
@@ -267,4 +267,69 @@ jobs:
...
+      - uses: ./.github/actions/k8s-version-config
+        name: Setup k8s cluster
+        with:
+          k8s-version: v1.22
+      - uses: actions/checkout@v2
+        with:
+          ref: 'master'
+      - name: Configure Connaisseur (master)
+        run: |
+          yq e '.detectionMode = "true"' -i helm/values.yaml
```
phbelitz, line 298: this will set the `detectionMode` to a string with value `"true"`, NOT a boolean. internally this will still evaluate as `true`, since the string isn't empty, but might cause trouble when changing back.

```suggestion
          yq e '.detectionMode=true' -i helm/values.yaml
```

annekebr, line 298: Why do we need that?

.github/workflows/cicd.yaml
```diff
@@ -267,4 +267,69 @@ jobs:
...
+      - name: Install Connaisseur (master)
+        run: |
+          make install
+      - name: Get image name & version
+        run: |
+          kubectl get pods -n connaisseur -o jsonpath="{.items[*].spec.containers[*].image}"
+      - name: Test denial of unsigned nv1 image
+        run: |
+          kubectl run pod --image=securesystemsengineering/testimage:unsigned >output.log 2>&1 || true
```
annekebr, line 307: No positive test needed here?

helm/values.yaml
```diff
@@ -1,7 +1,7 @@
 # configure Connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v2.2.0
+  image: securesystemsengineering/connaisseur:v2.2.1
```
annekebr, line 4: Why is there a version bump? This only touches tests, doesn't it?

helm/values.yaml
```diff
@@ -1,7 +1,7 @@
 # configure Connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v2.2.0
```
Starkteetje, line 4: This PR should not increase the version

.github/workflows/cicd.yaml
```diff
@@ -267,4 +267,69 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Upgrade Connaisseur to current branch
+        run: |
+          make upgrade
+      - name: Get image name & version
+        run: |
+          kubectl get pods -n connaisseur -o jsonpath="{.items[*].spec.containers[*].image}"
+      - name: Test denial of unsigned nv1 image
+        run: |
+          kubectl run pod --image=securesystemsengineering/testimage:unsigned >output.log 2>&1 || true
```
annekebr, line 324: No positive test?

phbelitz (dismissed):

.github/workflows/cicd.yaml
xopham: the develop version will not be signed yet and thus `upgrade` would fail

.github/workflows/cicd.yaml
xopham: unsure, but can add positive test, might make a separate bash script

.github/workflows/cicd.yaml
xopham: as above

xopham:
NOTE: The integration test fails right now bc the integration test script won't be there on master. To make sure it does work, I simply provided the branch script to the master branch. I would however prefer if this was not done further down the road, as the test script would better be branch specific as it should work for the version present on that branch. To see the results: https://github.com/sse-secure-systems/connaisseur/runs/4116082859?check_suite_focus=true

.github/workflows/cicd.yaml
xopham: fixed

helm/values.yaml
xopham: fixed

.github/workflows/cicd.yaml
xopham: fixed

annekebr (approved):

PR #299, opened 2021-08-31
Production WSGI server

Starkteetje:
Use a production WSGI server to host the WSGI Connaisseur application. Load and stress testing led us to choose cheroot by a slim margin.

PR also contains some loadtests for the pipeline integration tests


docs/adr/ADR-7_wsgi-server.md
```diff
@@ -0,0 +1,298 @@
...
+
+There's plenty of WSGI server around and the question poses itself, which one to pick. Flask itself has a [list of servers](https://flask.palletsprojects.com/en/1.1.x/deploying/), there's comparisons around, for example [here](https://medium.com/django-deployment/which-wsgi-server-should-i-use-a70548da6a83) and [here](https://www.appdynamics.com/blog/engineering/a-performance-analysis-of-python-wsgi-servers-part-2/). The choice, which WSGI servers to test was somewhat arbitrary among better performing ones in the posts.
+
+Contenders were Bjoern, Cheroot, Flask, Gunicorn and uWSGI. Bjoern was immediately dropped, since it works only with Python2. Gunicorn was tested for a bit, but since it delivered worse results than the others and it requires a writable log directory, it was also dropped from contention.
+
+The remaining three were tested over a rather long of development, i.e. from before the first bit of validation parallelization to after the 2.0 release. All tests were run on local minikube/kind clusters with rather constrained resources in the expectation that this will still provide reasonable insight into the servers' behaviour on regular production clusters.
```
xopham, line 27:
```suggestion
The remaining three were tested over a rather long time of development, i.e. from before the first bit of validation parallelization to after the 2.0 release. All tests were run on local minikube/kind clusters with rather constrained resources in the expectation that this will still provide reasonable insight into the servers' behavior on regular production clusters.
```

xopham (commented):
wow...awesome review!

This seems to give a tendency towards cheroot which seems a good choice though less known.

However, when checking these out I stumbled across the following blog post: https://fgimian.github.io/blog/2018/05/17/choosing-a-fast-python-api-framework/

In there, it is mentioned that CherryPy (Cheroot) is ruled out bc it was slow, but more importantly that flask was ruled out as a WSGI framework since it was slow and that [falcon](https://github.com/falconry/falcon) is apparently blazingly fast and "The no-nonsense REST API and microservices framework for Python developers," which sounds pretty good for our use-case. Now, this will cause some additional effort to test whether the framework makes a significant difference, as that would indeed require some rewriting, but I am wondering whether it makes sense to take a look.
Also, Bjoern is claimed to work for python 2 and 3: https://github.com/jonashaag/bjoern . Apparently, it is a strong choice for containerized environments:
- https://github.com/kubeup/python-wsgi-benchmark
- https://www.appdynamics.com/blog/engineering/a-performance-analysis-of-python-wsgi-servers-part-2/

Beyond that I noticed that uwsgi is supposedly very complex and overloaded.

In summary, it seems cheroot is a decent choice, Bjoern may be better and apparently using Falcon instead of Flask could strongly affect speed, but would require a bit of rewriting...

tests/integration/load-test.sh
```diff
@@ -0,0 +1,57 @@
...
+NUMBER_CREATED=$(cat output.log | grep "deployment[.]apps/redis-[0-9]* created" | wc -l || echo "0")
+if [[ ${NUMBER_CREATED} != "${NUMBER_OF_INSTANCES}" ]]; then
+  echo "Failed load test. Only ${NUMBER_CREATED}/${NUMBER_OF_INSTANCES} pods were created"
+  exit 1
+else
+  echo "Successfully passed load test"
+fi
+
+# Cleanup can be skipped in GitHub Actions, since uninstall is already tested in other integration test
+if [[ "${GITHUB_ACTIONS-}" == "true" ]]; then
```
phbelitz, line 45: Is the `-` at the end of `GITHUB_ACTIONS` intentional?

Starkteetje: Yes. If running locally ${GITHUB_ACTIONS} will not be set. Since we run with -u this would throw. This is bash for "use GITHUB_ACTIONS if set, else use the string after the dash, i.e. empty string -> hence not throwing due to -u locally. Plan was to add this to all but one integration test later to speed up the pipeline

helm/values.yaml
```diff
@@ -1,7 +1,7 @@
 # configure Connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v2.2.1
+  image: securesystemsengineering/connaisseur:v2.3.0
```
xopham, line 4: changing the webserver does not qualify as minor change, does it? it is not a feature, does not extend or remove functionality. Why would you go for `v2.3.0`?

phbelitz (approved):

xopham (commented):
one small version comment, otherwise lgtm!

helm/values.yaml
Starkteetje: I think it was a gut decision at the time of "being bigger" but you are correct. I shall revert this

annekebr (commented):
Looks very good! Can we delete flask from the requirements?

Starkteetje:
> 
> 
> Looks very good! Can we delete flask from the requirements?

No, since we're still using the flask WSGI application. Just no longer the server. But that's a great point, gotta rename the flask_server file

docs/adr/ADR-7_wsgi-server.md
```diff
@@ -0,0 +1,297 @@
...
+    image: redis
+  - name: init3
+    image: node
+</pre>
+</details>
+
+None of the servers regularly managed to pass this particular loadtest. However, the pods powered by the Flask server regularly died and had to be restarted, whereas both Cheroot and uWSGI had nearly no restarts and never on all instances. uWSGI seldomly even manged to pass the test.
+
+##### Less complex requests with some load
+Since in the above the most complex request was the bottleneck, we tried an instance of the test with less complexity in the individual requests but more. However, that led to no real distinguishing behaviour across the servers.
```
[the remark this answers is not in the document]
Starkteetje: Yes

phbelitz (approved):

PR #318, opened 2021-09-24
fix: only load required delegantions

### LINKED ISSUE #171: Only download required delegations
**Describe the bug**
Connaisseur downloads all delegation trust data.

**Expected behavior**
Connaisseur should only download the trust data that is actually needed, ignoring the rest. Should one of the trust data be empty, raise an error.

phbelitz:
When using delegation for nv1, Connaisseur now only loads those delegation files it actually needs for verification and not all available.

fixes #171


Starkteetje (changes requested):
There should be a test to highlight and ensure the functional diff

Starkteetje (changes requested):
Commit has `delegantions`, other than that lgtm

phbelitz:
OKOKOK. I had to change the code a bit after adding additional tests and relizing that there were some bugs... so the last commit adds 3 tests: 
1. checking bahavior when a delegation is required, that doesn't exist (didn't had to change the code for that). 
2. checking behavior when a delegation is required (e.g. `del1`) + the `targets.json` also has said delegation listed, but the actual file (e.g. `targets/del1.json`) is missing (had to change the code).
3. checking bahavior when no delegations are required, the `targets.json` says it theoretically has delegation (even though not required), but the delegation files themselves are missing. this corresponds to a very rare edge case, where delegations are added to a repo, but no signatures were yet created using these delegations. normally we would use the `targets/releases.json` file, but since this file doesn't exists yet, we have to use the `targest.json` instead (had to change the code again).

Starkteetje (approved):
Nice one catching the regression

PR #381, opened 2021-10-22
ci: Rework integration tests

phbelitz:
Reworked the integration tests, so that all logic is bundled into a single shell script file and all test cases in a yaml file. The upgrade test has been excluded.


tests/integration/values.yaml
```diff
@@ -0,0 +1,149 @@
+# configure Connaisseur deployment
```
Starkteetje, line 1: Do you really want to maintain a copy of the values.yml? I think using a copy/the default one and make targeted changes is a) easier to maintain and b) makes sure we actually test the current state and not some old things

tests/integration/new-integration-test.sh
```diff
@@ -0,0 +1,197 @@
+set -eo pipefail
```
Starkteetje, line 1: Why no -u?

tests/integration/new-integration-test.sh
```diff
@@ -0,0 +1,197 @@
...
+upgrade() {
+  echo -n 'Upgrading Connaisseur...'
+  make upgrade > /dev/null || { echo -e ${FAILED}; exit 1; }
+  echo -e "${SUCCESS}"
+}
+
+### UNINSTALING CONNAISSEUR ####################################
+uninstall() {
+  echo -n 'Uninstalling Connaisseur...'
+  # 'make' is chosen deliberately below to test the Makefile, while other tests use 'helm' directly
```
Starkteetje, line 70: This probably makes no sense here. Also if this is intended to replace all, make sure we still test both make and helm

tests/integration/new-integration-test.sh
```diff
@@ -0,0 +1,197 @@
...
+    namespace_val_int_test
+    uninstall
+    ;;
+  "deployment")
+    update_values '.policy += {"pattern": "docker.io/library/*:*", "validator": "dockerhub-basics", "with": {"trust_root": "docker-official"}}' 
+    install
+    deployment_int_test
+    uninstall
+    ;;
+  "all")
```
Starkteetje, line 178: This gonna run for 2 days

tests/integration/deployment_test_cases.txt
```diff
@@ -0,0 +1,10 @@
+Testing 1 signed image deployment...;apply;tests/integration/deployments/deployment_i1.yaml;default;deployment.apps/i1-deployment created;
```
xopham, line 1: would strongly recommend to use a better machine readable format (YAML, JSON) that will also be less fragile against strings with `;` and easy to edit via `yq` or `jq`. This is hard to read and worse to edit, I would think

xopham: btw. would prefer yaml ;-)

xopham (commented):
made an initial comment on the test case data format

tests/integration/values.yaml
phbelitz: okokok

tests/integration/new-integration-test.sh
phbelitz: then the color coding for `SUCCESS` and `FAILURE` don't work

tests/integration/new-integration-test.sh
phbelitz: added functions for helm as well

tests/integration/new-integration-test.sh
phbelitz: let's have some options, shall we?

tests/integration/deployment_test_cases.txt
phbelitz: changed to yaml, have a look

tests/integration/cases.yaml
```diff
@@ -0,0 +1,210 @@
+test_cases:
+  regular:
+  - id: ru
```
xopham, line 3: I would prefer explicit names, also these are not IDs

phbelitz: the purpose of these are a quick look up in the yaml file, should one of the tests fail + unique names for the pods. they are not suposed to describe anything and could be completely arbitrary, as long as they are unique.

.github/workflows/release.yaml
```diff
@@ -27,65 +27,77 @@ jobs:
     needs: [version-match]
     strategy:
       matrix:
-        integration-test-script: [
-          'integration-test.sh',
-          'preconfig-integration-test.sh',
-          'cosign-integration-test.sh',
-          'deployment-integration-test.sh']
+        integration-test-arg: ["regular", "cosign", "deployment", "pre-config"]
```
Starkteetje, line 30: In comment in cicd.yaml it was named `preconfig`, here `pre-config`. Consistency

.github/workflows/release.yaml
```diff
@@ -27,65 +27,77 @@ jobs:
...
     services:
       alerting-endpoint:
         image: securesystemsengineering/alerting-endpoint
         ports:
           - 56243:56243
     steps:
       - uses: actions/checkout@v2
-      - name: Install yq and bash
```
Starkteetje, line 42: it still installs bash, why remove?

tests/integration/cases.yaml
```diff
@@ -0,0 +1,210 @@
...
+    txt: Testing unsigned image...
+    type: deploy
+    ref: securesystemsengineering/testimage:unsigned
+    namespace: default
+    msg: Unable to find signed digest for image docker.io/securesystemsengineering/testimage:unsigned.
+    validity: INVALID
```
Starkteetje, line 9: Maybe expexted_result instead of validity?

tests/integration/cases.yaml
```diff
@@ -0,0 +1,210 @@
+test_cases:
+  regular:
+  - id: ru
+    txt: Testing unsigned image...
+    type: deploy
+    ref: securesystemsengineering/testimage:unsigned
+    namespace: default
+    msg: Unable to find signed digest for image docker.io/securesystemsengineering/testimage:unsigned.
```
Starkteetje, line 8: Maybe expected_msg?

.github/workflows/cicd.yaml
```diff
@@ -191,61 +191,84 @@ jobs:
...
-          'load-test.sh'
-        ]
+        integration-test-arg:
+          [
+            "regular",
+            "cosign",
+            "namespace-val",
+            "deployment",
+            "pre-config",
+            "stress-test",
```
Starkteetje, line 210: We still need to fix that, but since the stress-test runs way longer than the rest, can we move it up in the matrix such that it at least gets executed first?

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+  fi
+
+  if [[ $7 != "null" ]]; then
+    DEPLOYMENT_VALIDITY[$7]=$((${DEPLOYMENT_VALIDITY[$7]} + 1))
+  fi
+}
+
+### MULTI TEST CASE FROM FILE ####################################
+multi_test() {
+  test_cases=$(yq e -o=json ".test_cases.$1" tests/integration/cases.yaml)
```
Starkteetje, line 41: If the initial set is in yaml, why convert to JSON and use jq instead of yq on initial set?

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+    echo -e "${SUCCESS}"
+  fi
+
+  if [[ $7 != "null" ]]; then
+    DEPLOYMENT_VALIDITY[$7]=$((${DEPLOYMENT_VALIDITY[$7]} + 1))
+  fi
+}
+
+### MULTI TEST CASE FROM FILE ####################################
+multi_test() {
```
Starkteetje, line 40: Would be nice to have the expected args here similar to above or comment

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+    REF=$(echo ${test_case} | jq -r ".ref")
+    NAMESPACE=$(echo ${test_case} | jq -r ".namespace")
+    MSG=$(echo ${test_case} | jq -r ".msg")
+    VALIDITY=$(echo ${test_case} | jq -r ".validity")
+    single_test "${ID}" "${TEST_CASE_TXT}" "${TYPE}" "${REF}" "${NAMESPACE}" "${MSG}" "${VALIDITY}"
+  done
+}
+
+### WORKLOAD TEST ####################################
+workload_test() {
```
Starkteetje, line 57: Would be nice to have the expected args here similar to above or comment

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+    cat output.log
+    echo "::endgroup::"
+    EXIT="1"
+  else
+    echo -e "${SUCCESS}"
+  fi
+}
+
+### INSTALLING CONNAISSEUR ####################################
+install() {
```
Starkteetje, line 111: Maybe install_make to be explicit about the difference to helm?

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+  helm install connaisseur helm --atomic --create-namespace \
+    --namespace connaisseur >/dev/null || {
+    echo -e "${FAILED}"
+    exit 1
+  }
+  echo -e "${SUCCESS}"
+}
+
+### UPGRADING CONNAISSEUR ####################################
+upgrade() {
```
Starkteetje, line 131: Maybe upgrade_make?

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+upgrade_helm() {
+  echo -n 'Upgrading Connaisseur...'
+  helm upgrade connaisseur helm -n connaisseur --wait >/dev/null || {
+    echo -e ${FAILED}
+    exit 1
+  }
+  echo -e "${SUCCESS}"
+}
+
+### UNINSTALING CONNAISSEUR ####################################
```
Starkteetje, line 149:
```suggestion
### UNINSTALLING CONNAISSEUR ####################################
```

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+uninstall() {
```
Starkteetje, line 150: Maybe uninstall_make?

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+  echo -n 'Uninstalling Connaisseur...'
+  helm uninstall connaisseur -n connaisseur &&
+    kubectl delete ns connaisseur >/dev/null || {
+    echo -e "${FAILED}"
+    exit 1
+  }
+  echo -e "${SUCCESS}"
+}
+
+update_values() {
```
Starkteetje, line 169: What are expected args?

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+### COSIGN TEST ####################################
+cosign_int_test() {
+  multi_test "cosign"
+}
+
+### NAMESPACE VALIDATION TEST ####################################
+namespace_val_int_test() {
+  echo -n "Creating namespaces..."
+  kubectl create namespace ignoredns >/dev/null
+  kubectl label ns ignoredns securesystemsengineering.connaisseur/webhook=ignore >/dev/null
```
Starkteetje, line 229: Should this be ignore NS instead of DNS? Same with validatens below

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
...
+  kubectl create namespace validatedns >/dev/null
+  kubectl label ns validatedns securesystemsengineering.connaisseur/webhook=validate >/dev/null
+  echo -e "${SUCCESS}"
+
+  multi_test "ignore-namespace-val"
+  update_values '.namespacedValidation.mode="validate"'
+  upgrade
```
Starkteetje, line 236: Why does namespace test need an upgrade?

tests/integration/integration-test.sh
phbelitz: processing all in `yq` was very slow. `jq` was faster so i decided to output as json and work with `jq` instead.

tests/integration/integration-test.sh
phbelitz: ingnored ns, even though i as well read DNS everytime.

tests/integration/integration-test.sh
phbelitz: because the `namespacedValidation.mode` is switched from `ignore` to `validate`, which needs to be changed in the webhookconfiguration. that can only happen through an `helm upgrade`.

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,305 @@
-#! /bin/bash
-set -euo pipefail
+set -eo pipefail
```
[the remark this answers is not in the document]
phbelitz: ok dunno, didn't work before, now it does. :shrug:

tests/integration/integration-test.sh
Starkteetje: 😆 lalalala

tests/integration/integration-test.sh
Starkteetje: That's pretty unintuitive. Please add a comment to explain

tests/integration/integration-test.sh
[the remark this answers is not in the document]
Starkteetje: 🎉

tests/integration/integration-test.sh
Starkteetje: As I always say, I'm not firm with the helm part. If that's obvious, feel free to ignore, otherwise add a comment

.github/workflows/release.yaml
```diff
@@ -44,48 +40,64 @@ jobs:
...
-        mkdir -p charts
-        mv connaisseur*.tgz ./charts
-        git checkout gh-pages
-        cd charts
-        helm repo index . --url https://sse-secure-systems.github.io/connaisseur/charts
-        cd ..
-        git add ./charts
-        git commit -m "Publish helm chart ${CHART_VERSION}"
-        git push https://${{ secrets.GITHUB_TOKEN }}@github.com/sse-secure-systems/connaisseur.git
+      - name: Install helm and git
```
Starkteetje, line 75: Since we "Configure Connaisseur" this should be "Install Helm and Git", same for "Lint Helm chart" and potentially more. Or we go for no names in yml chart and go for "Configure connaisseur"
Idc, but consistency

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,304 @@
...
+    cat output.log
+    echo "::endgroup::"
+    EXIT="1"
+  else
+    echo -e "${SUCCESS}"
+  fi
+}
+
+### INSTALLING CONNAISSEUR ####################################
+install_make() {
```
Starkteetje, line 111: Sorry, probably my suggestion was bad: This reads like "please install make on my system". Can we do "action_via_tool" or "tool_action" instead? 😅

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,304 @@
...
+  pre_config_int_test
+  for wo in "${WOLIST[@]}"; do
+    workload_test "${wo}"
+  done
+  ;;
+"stress-test")
+  install_make
+  stress_test
+  ;;
+"all")
```
Starkteetje, line 285: Why does `all` not contain stress and pre-and-workload?

tests/integration/integration-test.sh
```diff
@@ -1,238 +1,304 @@
...
+  multi_test "pre-config"
+}
+
+case $1 in
+"regular")
+  install_make
+  regular_int_test
+  uninstall_make
+  ;;
+"cosign")
```
Starkteetje, line 255: For running locally, all tests that run successfully, should also uninstall resources they created, in particular Conny. You can hide that behind a flag, as e.g. in previous stress-test

Starkteetje: But also spawned pods/deployments...

Starkteetje:
```
# Cleanup can be skipped in GitHub Actions, since uninstall is already tested in other integration test
if [[ "${GITHUB_ACTIONS-}" == "true" ]]; then
  exit 0
fi
```

.github/workflows/release.yaml
phbelitz: hey that's unfair. i just did the formatting here ... will apply changes though ....

Starkteetje (approved):

PR #393, opened 2021-10-28
Allow CAPS in image tag

hsuchan:
Only tags can have uppercase characters.
Excluding tags from lowercase comparison.

### DIFF
--- a/connaisseur/image.py
+++ b/connaisseur/image.py
@@ -58,7 +58,7 @@ class Image:
             image.removesuffix(name_tag).removeprefix(self.registry)
         ).strip("/") or ("library" if self.registry == "docker.io" else "")
 
-        if (self.repository + name_tag).lower() != self.repository + name_tag:
+        if (self.repository + self.name).lower() != self.repository + self.name:
             msg = "{image} is not a valid image reference."
             raise InvalidImageFormatError(message=msg, image=image)
 
--- a/tests/test_image.py
+++ b/tests/test_image.py
@@ -104,6 +104,15 @@ import connaisseur.exceptions as exc
             "docker.io",
             pytest.raises(exc.InvalidImageFormatError),
         ),
+        (
+            "docker.io/library/image:Tag",
+            "image",
+            "Tag",
+            None,
+            "library",
+            "docker.io",
+            fix.no_exc(),
+        ),
     ],
 )
 def test_image(


phbelitz (dismissed):
please add another test case in `tests/test_image.py`, or tell me if I should do the changes :) thank you very much for your support!

hsuchan:
> please add another test case in `tests/test_image.py`, or tell me if I should do the changes :) thank you very much for your support!

I added a test case for tag using uppercase letter.

connaisseur/image.py
```diff
@@ -58,7 +58,8 @@ def __init__(self, image: str):
             image.removesuffix(name_tag).removeprefix(self.registry)
         ).strip("/") or ("library" if self.registry == "docker.io" else "")
 
-        if (self.repository + name_tag).lower() != self.repository + name_tag:
+        first_name_tag = name_tag.split(":")[0]
```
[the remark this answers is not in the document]
hsuchan: I updated the if test to use `self.name`

xopham (commented):
@hsuchan lgtm!

Could you adjust the commit message to follow the semantic and conventional commits as described in the [contributing guide](https://sse-secure-systems.github.io/connaisseur/v2.2.0/CONTRIBUTING/#semantic-and-conventional-commits). Should be something along the lines:
```
fix: allow CAPS in image tag

fixes #392 
```
We would then need rebase to `develop` again, sign it and could probably merge it.

Thanks for contributing :pray:

hsuchan:
I did rebase to develop. Let me know if I did it correctly.

tests/test_image.py
```diff
@@ -104,6 +104,15 @@
             "docker.io",
             pytest.raises(exc.InvalidImageFormatError),
         ),
+        (
+            "docker.io/library/image:Tag",
+            "image",
+            "Tag",
+            None,
+            None,
```
xopham, line 112:
```suggestion
            "library",
```
This needs to be the repository value which is "library" in your case, as given in line 8 though I admit the description is a bit confusing

hsuchan:
I also signed my commit.

xopham (approved):
@hsuchan great work :pray: 
I briefly fixed the minor test issue. looks perfect now :slightly_smiling_face:

tests/test_image.py
hsuchan: I was just copying the previous test ("docker.io/Library/image:tag") and changed the test case.
I also noticed that the repository was set to "None" but reasoned that any Image under docker.io/library can be called by it's short name: 
```
docker pull docker.io/library/nginx
docker pull library/nginx
docker pull nginx
```
Let me know if you want me to change the repository to "library"

xopham (dismissed):
forgot to sign the commit previously :sweat_smile: 
will merge now :rocket:

tests/test_image.py
hsuchan: I made the suggested change.

xopham (approved):

tests/test_image.py
xopham: I see what you mean. That is, however, only a unit test that does not take the docker.io specifics into account. I agree that specifically choosing `library` as a repo name in the previous example is confusing though.

thanks, looks good now.

PR #396, opened 2021-10-29
fix: support api versions for k8s workloadobjects, add tests

### LINKED ISSUE #382: Update supported k8s resources and api version
**Describe the bug**
with latest releases API versions and available resources have changed. Connaisseur should support the latest changes.

xopham:
fixes #382 

## description
- adds integration tests for workload objects and api versions for each Kubernetes version
- fixes currently unsupported api versions

## checklist
- [x] according to Contributing guide
- [x] tests added where necessary

### DIFF
--- a/tests/integration/workload-objects-integration-test.sh
+++ b/tests/integration/workload-objects-integration-test.sh
@@ -1,0 +1,69 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+# This script is expected to be called from the root folder of Connaisseur
+
+# List of Workload Objects
+woList=("CronJob" "DaemonSet" "Deployment" "Job" "Pod" "ReplicaSet" "ReplicationController" "StatefulSet")
+
+echo '** Prepare Connaisseur config'
+yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' helm/values.yaml tests/integration/preconfig-update.yaml
+echo '>> Config set'
+
+echo -e '\n** Install Connaisseur'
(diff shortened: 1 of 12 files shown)


annekebr:
This PR is not only related to `tests` as it changes the supported API versions of Connaisseur as well, which should be a separate PR I think

xopham:
@annekebr but then this pr fails. I renamed the pr to make it: fixing the workload object api versions, added tests accordingly. Would that work?

tests/integration/workload-objects-integration-test.sh
```diff
@@ -0,0 +1,69 @@
...
+    # UNSIGNED
+    export TAG=unsigned
+
+    echo -e "\n** Render '${KIND}' using '${APIVERSION}' and '${TAG}' image"
+    envsubst < tests/integration/workload-objects/${KIND}.yaml | cat
+
+    echo -e "\n** Deploy '${KIND}' using '${APIVERSION}' and '${TAG}' image"
+    envsubst < tests/integration/workload-objects/${KIND}.yaml | kubectl apply -f - >output.log 2>&1 || true
+
+    if [[ ! "$(cat output.log)" =~ 'Unable to find signed digest for image docker.io/securesystemsengineering/testimage:unsigned.' ]]; then
```
annekebr, line 33: It would probably be sufficient to deploy a valid signed image without error, wouldn't it? But as long as we don't run into timing issues, I don't mind testing more :)

annekebr (dismissed):

tests/integration/workload-objects-integration-test.sh
xopham: that is what I thought and it seems that the k3s never fails with timing issues

annekebr (approved):

PR #405, opened 2021-11-05
ci: speedup upgrade integration test

xopham:
Fixes no issue

## Description

- randomizes the pod names to avoid having to delete pods between tests on `master` and test branch

## Checklist

### DIFF
--- a/tests/integration/upgrade-integration-test.sh
+++ b/tests/integration/upgrade-integration-test.sh
@@ -3,8 +3,11 @@ set -euo pipefail
 
 # This script is expected to be called from the root folder of Connaisseur
 
+# Creating a random index to label the pods and avoid name collision for repeated runs
+RAND=$(head -c 5 /dev/urandom | hexdump -ve '1/1 "%.2x"')
+
 echo 'Testing nv1 unsigned image...'
-kubectl run pod --image=securesystemsengineering/testimage:unsigned >output.log 2>&1 || true
+kubectl run pod-${RAND} --image=securesystemsengineering/testimage:unsigned >output.log 2>&1 || true
 
 if [[ ! "$(cat output.log)" =~ 'Unable to find signed digest for image docker.io/securesystemsengineering/testimage:unsigned.' ]]; then
   echo 'Failed to deny unsigned image or failed with unexpected error. Output:'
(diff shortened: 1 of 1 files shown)


tests/integration/upgrade-integration-test.sh
```diff
@@ -28,17 +29,15 @@ else
...
-if [[ "$(cat output.log)" != 'pod/dpod created' ]]; then
+if [[ "$(cat output.log)" != 'pod/dpod-${RAND} created' ]]; then
   echo 'Failed to allow signed image. Output:'
   cat output.log
   exit 1
 else
   echo 'Successfully allowed usage of signed image'
 fi
 
-kubectl delete pods --all || { echo 'Failed to delete test pods'; exit 1; }
```
Starkteetje, line 41: Can you put this behind a GITHUB_ACTIONS flag like here: https://github.com/sse-secure-systems/connaisseur/blob/develop/tests/integration/load-test.sh#L45

Then you can still run it locally without polluting your env, but still gain speed in the CI?

xopham: done

tests/integration/upgrade-integration-test.sh
```diff
@@ -3,8 +3,9 @@ set -euo pipefail
 
 # This script is expected to be called from the root folder of Connaisseur
 
+RAND=$(head -c 5 /dev/urandom | hexdump -ve '1/1 "%.2x"')
```
Starkteetje, line 6: I'd add a comment here explaining why you're doing this

tests/integration/upgrade-integration-test.sh
```diff
@@ -28,17 +29,20 @@ else
...
+if [[ "$(cat output.log)" != "pod/dpod-${RAND} created" ]]; then
   echo 'Failed to allow signed image. Output:'
   cat output.log
   exit 1
 else
   echo 'Successfully allowed usage of signed image'
 fi
 
-kubectl delete pods --all || { echo 'Failed to delete test pods'; exit 1; }
+# Cleanup can be skipped in GitHub Actions, since uninstall is already tested in other integration test
```
Starkteetje, line 42: Should probably add the uninstall as well to do proper cleanup and make the comment apply

xopham: You mean uninstall of Connaisseur? The integration test does not install Connaisseur, so I would not uninstall it as well

tests/integration/upgrade-integration-test.sh
xopham: done

tests/integration/upgrade-integration-test.sh
Starkteetje: Oh, I assumed it was more or less the same as the others, but then no install it is ;)

Starkteetje (approved):

PR #421, opened 2021-11-19
Refactor imports, comments, docstrings and some types

Starkteetje:
This commit refactors the import order (PEP-8), some comments, many docstrings (PEP-257), a regex and some types where there were inconsistencies present


connaisseur/alert.py
```diff
@@ -2,9 +2,10 @@
 import logging
 import os
 from datetime import datetime
+from typing import Optional
```
Starkteetje, line 5: That's probably something that can be debated whether we want to include it at all

phbelitz (dismissed):

phbelitz (approved):

PR #422, opened 2021-11-23
feat: support custom LogLevel

sf-jmarcou:
Add `defaultLogLevel` helm value.
No breaking change.

### DIFF
--- a/helm/templates/env.yaml
+++ b/helm/templates/env.yaml
@@ -12,7 +12,7 @@ data:
   {{- if .Values.debug }}
   LOG_LEVEL: DEBUG
   {{- else }}
-  LOG_LEVEL: INFO
+  LOG_LEVEL: {{.Values.logLevel | default "INFO"}}
   {{- end }}
   {{- if .Values.detectionMode }}
   DETECTION_MODE: "1"
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -149,6 +149,12 @@ automaticChildApproval:
 
 # debug: true
 
+
+# The "logLevel" configuration option adds a partial redundancy to the `debug` setting.
+# Removing the `debug` setting is a breaking change though - we are going to remove the `debug` setting in the context of a larger refactoring to avoid multiple breaking releases.
+# Option to configure the log level. Either one of `DEBUG`, `INFO`, `WARNING`, `ERROR`, `CRITICAL`. Defaults to `INFO`
+logLevel: INFO
+
 # alerting is implemented in form of simple POST requests with json payload
 # you can use and/or adapt the predefined Slack/OpsGenie/Keybase templates and the examples below
 # to channel alert notifications to Slack/OpsGenie/Keybase or create a custom template for a customized alert


helm/templates/env.yaml
```diff
@@ -12,7 +12,7 @@ data:
   {{- if .Values.debug }}
   LOG_LEVEL: DEBUG
   {{- else }}
-  LOG_LEVEL: INFO
+  LOG_LEVEL: {{.Values.defaultLogLevel}}
```
annekebr, line 15:
```suggestion
  LOG_LEVEL: {{.Values.logLevel | default "INFO"}}
```

helm/values.yaml
```diff
@@ -148,6 +148,7 @@ automaticChildApproval:
   enabled: true
 
 # debug: true
+defaultLogLevel: INFO
```
annekebr, line 151:
```suggestion
# The "logLevel" configuration option adds a partial redundancy to the `debug` setting.
# Removing the `debug` setting is a breaking change though - we are going to remove the `debug` setting in the context of a larger refactoring to avoid multiple breaking releases.
# logLevel: INFO
```

helm/templates/env.yaml
annekebr: I'd suggest to have an explicit default at this point to avoid having to specify `logLevel` in the `helm/values.yaml`.

annekebr (commented):
Hey @sf-jmarcou! Thanks a lot for your contribution! A configurable log level is a great idea :+1: I had a few comments to keep the minimal required config as small as possible.

annekebr (commented):
I changed the target branch to point to `develop` to schedule this PR for the next release according to our workflow :)

Please pull the repo again and rebase your branch to the latest develop :)

helm/templates/env.yaml
sf-jmarcou: Hi @annekebr. Thanks for your review!

In an helm chart, the values.yaml file represents the API for consumer to know what parameters are supported.
For that reason, I suggest to keep and define the `logLevel` yaml key in the values.yaml file, with its default `INFO` value. This is a good place as well for users to know what is the default value just by reading the values.yaml file.

helm/values.yaml
sf-jmarcou: Indeed, `debug` and `logLevel` are confusing and redundant. Removing it would be a breaking change.
That's the reason I suggested `defaultLogLevel` which implicitly means the logLevel if not override by `debug` mode.
But, it depends on the breaking change management strategy.

helm/templates/env.yaml
annekebr: Hi @sf-jmarcou. I agree the `helm/values.yaml` should be self-explanatory. I edited my suggestion below to reflect the default value in a comment.

We really want to keep our `helm/values.yaml` as minimal as possible and make additional configuration options optional. :)

helm/values.yaml
annekebr: Yeah, we discussed that internally and want to remove the `debug` option with the next breaking change - and decided to add an explanatory comment to reflect that :)

annekebr (approved):
Thanks a lot for your contribution!

PR #423, opened 2021-11-23
Fix/bump app version

phbelitz:


xopham (commented):
maybe we should call the commit "update"

xopham (approved):
lgtm

Issue #426, opened 2021-11-26
[Elastic Common Schema] Alert Template for Elastic SIEM


operatorequals:
**Describe the feature**
Where I work we use the Elastic SIEM and gather all Security Events in there. As we consider denying a Container Image a serious event (as the CI/CD pipeline only works with signed images - denying means manual tampering), we need to ingest it into the SIEM so we can keep record of the following:
- Time of the event
- Cluster
- Namespace
- Image URL that was tried
- Hopefully Kubernetes user that tried the deployment

To add this feature to our stack, I created an Alert Template that follows the ECS [1]. This produces a JSON POST request
directly consumable by all HTTP/S log collector plugins [2][3], so the logs end up in an Elasticseach index (or wherever).

Finally, that format is very useful as SIEM Rules [5] can be created to optionally deliver alerts to Slack, or end up in Incident Response tools, using Kibana Integrations or Praeco integrations [4].

----

I will create another Issue for addition of Kubernetes ServiceAccount in the alert context [6], as well as the single-word messages (such as `allow`, `deny` to be used in appropriate places in the Schema [7])

[1] : https://www.elastic.co/guide/en/ecs/1.12/index.html
[2] : https://docs.fluentd.org/input/http
[3] : https://www.elastic.co/guide/en/logstash/current/plugins-inputs-http.html
[4] : https://github.com/johnsusek/praeco
[5] : https://github.com/elastic/detection-rules
[6] : https://github.com/sse-secure-systems/connaisseur/blob/master/connaisseur/alert.py#L89
[7] : https://www.elastic.co/guide/en/ecs/1.12/ecs-event.html#field-event-type

xopham:
@operatorequals this issue has been resolved by releasing #427 , right?

operatorequals:
Yes indeed!

PR #427, opened 2021-11-26
feat: create ECS alert template

operatorequals:
An ECS template can be used as an alert template, so Connaisseur
generates JSON logs valid in Elastic SIEM (with ECS schema 1.12.0 -
latest [1]).

This is part of the Issue: https://github.com/sse-secure-systems/connaisseur/issues/426

The values mapped are taken from the below fields:
- https://www.elastic.co/guide/en/ecs/1.12/ecs-orchestrator.html
- https://www.elastic.co/guide/en/ecs/1.12/ecs-container.html
- https://www.elastic.co/guide/en/ecs/1.12/ecs-agent.html
- https://www.elastic.co/guide/en/ecs/1.12/ecs-ecs.html
- https://www.elastic.co/guide/en/ecs/1.12/ecs-event.html

[1] : https://www.elastic.co/guide/en/ecs/1.12/index.html

## Checklist

- [X] PR follows [Contributing Guide](../docs/CONTRIBUTING.md)

image


helm/alert_payload_templates/ecs.json
```diff
@@ -0,0 +1 @@
+{"orchestrator":{"type":"kubernetes","cluster":{"name":"{{ cluster }}"},"namespace":"{{ namespace }}","resource":{"type":"Pod"}},"container":{"image":{"name":"{{ images }}"}},"agent":{"id":"{{ connaisseur_pod_id }}","type":"connaisseur"},"event":{"created":"{{ timestamp }}","id":"{{ request_id }}","severity":"{{ priority }}","category":"intrusion_detection","type":"denied","action":"{{ alert_message }}"},"ecs":{"version":"1.12.0"}}
```
xopham, line 1: the only thing that seems problematic to me is that the ECS version is pinned in the last entry. That would make this seem ECS version specific. Do you see a way to avoid this?
Maybe @annekebr also has an idea.

xopham (commented):
@operatorequals neat PR! Very pragmatic extension.

Generally looks good. I only have one smaller comment that should be addressed.
If that can be resolved, it seems this can be rebased, signed and merged.

Thanks for contributing :pray:

helm/alert_payload_templates/ecs.json
annekebr: Does ECS require the `"ecs":{"version":<version>}}` object and does it reject the paylaod if it's not present?

If so, I would suggest either 
a) making the version configurable in the `helm/values.yaml` assuming that the schema of accepted JSON payloads did not/will not change too often and adding a comment with the ECS versions that are known to be compatible right now
b) adjust the docs to reflect that this template is for this specific version only and if someone wants to use another ECS version, they can use the template as template and create a custom template changing the version and potentially other fields in the JSON payload, if the used version requires a different format. 

What do you think?

operatorequals: Hello @annekebr !
First of all, the `ecs.version` is required, and this is documented from elastic [1] (check the `core` level).

I believe that `a)` is a bit dirty, in the sense that we are not providing any other version, and we'll need to maintain ECS versions along the Helm chart.

I prefer `b)` and we could even rename the file to `ecs-1-12-0.json` to be very explicit.

----

[1] : https://www.elastic.co/guide/en/ecs/1.12/ecs-ecs.html#field-ecs-version

operatorequals: Also, the `orchestrator` field is not available before ECS `1.10`.

annekebr: b) is perfectly fine and renaming the file sounds like a good idea :-) Please adapt the docs accordingly and add a short sentence what one would need to change to use another version (create custom template with adapted version field plus potential changes in the payload compatible with the version). 

Thanks a lot for your input! :-)

operatorequals: Added some info on how to use ECS:
https://github.com/sse-secure-systems/connaisseur/pull/427/files#diff-dfedb6f792e2c252547cfc7fba0a5a3301668aaeb06f5b3a6332bde593990578R38

Transforming to different ECS is quite long scope for the doc. As there are so so many fields that could potentially be relevant. I suggest that, as ECS before `1.10.0` is unusable (as `orchestrator` map does not exist), we can wait for PRs for different ECS versions ( > `1.12.0`).

docs/features/alerting.md
```diff
@@ -19,7 +19,7 @@ Currently, Connaisseur supports alerting on either admittance of images, denial
...
 | -------------------------------------------------- | ----------------------------------------------------  | ----------------- | ------------------ | -------------------------------------------------------------------------------------------------- |
 | `alerting.cluster_identifier`                      | string                                                | `"not specified"` |                    | Cluster identifier used in alert payload to distinguish between alerts from different clusters.     |
-| `alerting.<category>.template`                     | `opsgenie`, `slack`, `keybase` or custom<sup>*</sup>  | -                 | :heavy_check_mark: | File in `helm/alert_payload_templates/` to be used as alert payload template.           |
+| `alerting.<category>.template`                     | `opsgenie`, `slack`, `keybase`, `ecs` or custom<sup>*</sup>  | -                 | :heavy_check_mark: | File in `helm/alert_payload_templates/` to be used as alert payload template.           |
```
annekebr, line 22:
```suggestion
| `alerting.<category>.template`                     | `opsgenie`, `slack`, `keybase`, `ecs-1-12-0` or custom<sup>*</sup>  | -                 | :heavy_check_mark: | File in `helm/alert_payload_templates/` to be used as alert payload template.           |
```

annekebr (approved):
Thanks for your contribution!

PR #428, opened 2021-11-26
feat: multi-signer support for cosign

### LINKED ISSUE #375: cosign + connaisseur - image with multiple cosign signatures to be verified before deployment
### Discussed in https://github.com/sse-secure-systems/connaisseur/discussions/371

Implement verification against multiple trust roots in the policy. Could be done by for example making the `trust_root` key in the policy a list:
```yaml
- pattern: "docker.io/my-app/*:*"
  validator: my-validator
  with:
    trust_root: ["vulnerability-scanner","qa-scanner"]
```
This could be extended by a `required` key to enforce some signatures or a `threshold` to set a minimum of signatures present

xopham:
Fixes #375

## Description

- a container image can be signed by multiple signers using different keys
- this feature enables validation against all, a threshold number of or specific required trust roots of a validator
- concurrent processing of signatures via asyncio is currently blocked by a asyncio bug
  - [Stackoverflow discussion](https://stackoverflow.com/questions/70624413/how-to-read-from-dev-stdin-with-asyncio-create-subprocess-exec)
  - [Python bug report](https://bugs.python.org/issue46364)
- concurrency is (for now) implemented via [`concurrent.futures`](https://docs.python.org/3/library/concurrent.futures.html)
  - futures yield a significant speedup as shown below
  - once the asyncio bug is fixed, both should be benchmarked again

#### asyncio (non-functional)
```bash
$ time k run unsigned --image=docker.io/securesystemsengineering/testimage:unsigned
Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: Image not compliant with validation policy (threshold of '2' not reached). The following errors occurred (please check the logs for more information):
* trust root 'alice': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'bob': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'charlie': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
kubectl run unsigned   0.18s user 0.07s system 3% cpu 7.411 total
```

#### concurrent.futures
```bash
$ time k run unsigned --image=docker.io/securesystemsengineering/testimage:unsigned
Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: Image not compliant with validation policy (threshold of '2' not reached). The following errors occurred (please check the logs for more information):
* trust root 'alice': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'bob': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'charlie': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
kubectl run unsigned   0.18s user 0.06s system 6% cpu 3.456 total
```

#### asyncio (non-functional)
```bash
$ time k run unsigned --image=docker.io/securesystemsengineering/testimage:unsigned
Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: Image not compliant with validation policy (threshold of '2' not reached). The following errors occurred (please check the logs for more information):
* trust root 'alice': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'bob': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'charlie': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'doug': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'eve': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'fein': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
kubectl run unsigned   0.22s user 0.04s system 1% cpu 14.861 total
```

#### concurrent.futures
```bash
$ time k run unsigned --image=docker.io/securesystemsengineering/testimage:unsigned
Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: Image not compliant with validation policy (threshold of '2' not reached). The following errors occurred (please check the logs for more information):
* trust root 'alice': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'bob': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'charlie': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'doug': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'eve': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'fein': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
kubectl run unsigned   0.20s user 0.05s system 6% cpu 3.535 total
```

#### asyncio (non-functional)
```bash
$ time k run unsigned --image=docker.io/securesystemsengineering/testimage:unsigned
Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: Image not compliant with validation policy (threshold of '2' not reached). The following errors occurred (please check the logs for more information):
* trust root 'alice': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'bob': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'charlie': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'doug': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'eve': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'fein': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'george': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'harry': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'ines': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
kubectl run unsigned   0.22s user 0.04s system 1% cpu 22.449 total
```

#### concurrent.futures
```bash
$ time k run unsigned --image=docker.io/securesystemsengineering/testimage:unsigned
Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: Image not compliant with validation policy (threshold of '2' not reached). The following errors occurred (please check the logs for more information):
* trust root 'alice': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'bob': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'charlie': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'doug': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'eve': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'fein': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'george': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'harry': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
* trust root 'ines': No trust data for image "docker.io/securesystemsengineering/testimage:unsigned".
kubectl run unsigned   0.20s user 0.04s system 4% cpu 4.838 total

```

## Checklist

### DIFF
--- a/connaisseur/validators/cosign/cosign_validator.py
+++ b/connaisseur/validators/cosign/cosign_validator.py
@@ -1,8 +1,11 @@
+# import asyncio
 import json
 import logging
 import os
 import re
 import subprocess  # nosec
+
+from concurrent.futures import ThreadPoolExecutor
 
 from connaisseur.crypto import load_key
 from connaisseur.exceptions import (
@@ -21,42 +24,115 @@ from connaisseur.validators.interface import ValidatorInterface
 class CosignValidator(ValidatorInterface):
     name: str
     trust_roots: list
+    vals: dict  # validations, a dict for each required trust root containing validated digests or errors
     k8s_keychain: bool
 
     def __init__(self, name: str, trust_roots: list, auth: dict = None, **kwargs):
         super().__init__(name, **kwargs)
         self.trust_roots = trust_roots
+        self.vals = {}
         self.k8s_keychain = False if auth is None else auth.get("k8s_keychain", False)
-
-    def __get_key(self, key_name: str = None):
-        key_name = key_name or "default"
-        try:
-            key = next(
-                key["key"] for key in self.trust_roots if key["name"] == key_name
-            )
-        except StopIteration as err:
-            msg = 'Trust root "{key_name}" not configured for validator "{validator_name}".'
-            raise NotFoundException(
-                message=msg, key_name=key_name, validator_name=self.name
-            ) from err
-        return "".join(key)
 
     async def validate(
         self, image: Image, trust_root: str = None, **kwargs
     ):  # pylint: disable=arguments-differ
-        key = self.__get_key(trust_root)
-        return self.__get_cosign_validated_digests(str(image), key).pop()
+        required = kwargs.get("required", [])
+        # if not configured, `threshold` is 1 if trust root is not "*" or
+        # `required` is specified and number of trust roots otherwise
+        threshold = kwargs.get(
+            "threshold",
+            1 if trust_root != "*" or any(required) else len(self.trust_roots),
+        )
 
-    def __get_cosign_validated_digests(self, image: str, key: str):
+        self.vals = self.__get_pinned_keys(trust_root, required, threshold)
+
+        # use concurrent.futures for now
+        # tasks = [self.__validation_task(k, str(image)) for k in self.vals.keys()]
+        # await asyncio.gather(*tasks)
+
+        # prepare executor
+        num_workers = len(self.vals)
+        executor = ThreadPoolExecutor(num_workers)
+        # prepare tasks
+        arguments = [(k, str(image)) for k in self.vals.keys()]
+        futures = [executor.submit(self.__validation_task, *arg) for arg in arguments]
+        # await results (output dropped as `self.vals` is updated within function)
+        for future in futures:
+            future.result()
+
+        return CosignValidator.__apply_policy(
+            vals=self.vals, threshold=threshold, required=required
+        )
+
+    def __get_pinned_keys(self, key_name: str, required: list, threshold: int):
+        """
+        Extract the pinned key(s) selected for validation from the list of trust roots.
+        """
+        key_name = key_name or "default"
+        available_keys = list(map(lambda k: k["name"], self.trust_roots))
+
+        # generate list of pinned keys
+        if key_name == "*":
+            if len(required) >= threshold:
+                pinned_keys = required
+            else:
+                pinned_keys = available_keys
+        else:
+            pinned_keys = [key_name]
+
+        # check if pinned keys exist in available trust roots
+        missing_keys = set(pinned_keys) - set(available_keys)
+        if missing_keys:
+            msg = 'Trust roots "{key_names}" not configured for validator "{validator_name}".'
+            raise NotFoundException(
+                message=msg,
+                key_names=", ".join(missing_keys),
+                validator_name=self.name,
+            )
+
+        # construct key validation dictionary for pinned keys
+        keys = {
+            k["name"]: {
+                "name": k["name"],
+                "key": "".join(k["key"]),
+                "digest": None,
+                "error": None,
+            }
+            for k in self.trust_roots
+            if k["name"] in pinned_keys
+        }
+
+        return keys
+
+    # async def __validation_task(self, trust_root: str, image: str):
+    def __validation_task(self, trust_root: str, image: str):
+        """
+        Async task for each validation to gather all required validations,
+        execute concurrently and update results.
+        """
+        try:
+            # self.vals[trust_root]["digest"] = await self.__get_cosign_validated_digests(
+            self.vals[trust_root]["digest"] = self.__get_cosign_validated_digests(
+                image, self.vals[trust_root]
+            )
+        except Exception as err:
+            self.vals[trust_root]["error"] = err
+            logging.info(err)
+
+    # async def __get_cosign_validated_digests(self, image: str, trust_root: dict):
+    def __get_cosign_validated_digests(self, image: str, trust_root: dict):
         """
         Get and process Cosign validation output for a given `image` and `key`
         and either return a list of valid digests or raise a suitable exception
         in case no valid signature is found or Cosign fails.
         """
-        returncode, stdout, stderr = self.__invoke_cosign(image, key)
+        # returncode, stdout, stderr = await self.__invoke_cosign(image, trust_root["key"])
+        returncode, stdout, stderr = self.__invoke_cosign(image, trust_root["key"])
+
         logging.info(
-            "COSIGN output for image: %s; RETURNCODE: %s; STDOUT: %s; STDERR: %s",
-            image,
+            "COSIGN output of trust root '%s' for image'%s': RETURNCODE: %s; STDOUT: %s; STDERR: %s",
+            trust_root["name"],
+            str(image),
             returncode,
             stdout,
             stderr,
@@ -79,7 +155,13 @@ class CosignValidator(ValidatorInterface):
                             "received by Cosign: {err_type}: {err}"
                         )
                         raise UnexpectedCosignData(
-                            message=msg, err_type=type(err).__name__, err=str(err)
+                            message=msg,
+                            err_type=type(err).__name__,
+                            err=str(err),
+                            trust_data_type="dev.cosignproject.cosign/signature",
+                            stderr=stderr,
+                            image=str(image),
+                            trust_root=trust_root["name"],
                         ) from err
                     # remove prefix 'sha256'
                     digests.append(digest.removeprefix("sha256:"))
@@ -92,6 +174,8 @@ class CosignValidator(ValidatorInterface):
                 message=msg,
                 trust_data_type="dev.cosignproject.cosign/signature",
                 stderr=stderr,
+                image=str(image),
+                trust_root=trust_root["name"],
             )
         elif "Error: no matching signatures:\n\nmain.go:" in stderr:
             msg = 'No trust data for image "{image}".'
@@ -100,6 +184,7 @@ class CosignValidator(ValidatorInterface):
                 trust_data_type="dev.cosignproject.cosign/signature",
                 stderr=stderr,
                 image=str(image),
+                trust_root=trust_root["name"],
             )
         else:
             msg = 'Unexpected Cosign exception for image "{image}": {stderr}.'
@@ -108,30 +193,29 @@ class CosignValidator(ValidatorInterface):
                 trust_data_type="dev.cosignproject.cosign/signature",
                 stderr=stderr,
                 image=str(image),
+                trust_root=trust_root["name"],
             )
         if not digests:
             msg = (
                 "Could not extract any digest from data received by Cosign "
                 "despite successful image verification."
             )
-            raise UnexpectedCosignData(message=msg)
-        return digests
+            raise UnexpectedCosignData(
+                message=msg,
+                trust_data_type="dev.cosignproject.cosign/signature",
+                stderr=stderr,
+                image=str(image),
+                trust_root=trust_root["name"],
+            )
+        return digests.pop()
 
-    def __invoke_cosign(self, image, key):
+    # async def __invoke_cosign(self, image: str, key: str):
+    def __invoke_cosign(self, image: str, key: str):
         """
         Invoke the Cosign binary in a subprocess for a specific `image` given a `key` and
         return the returncode, stdout and stderr. Will raise an exception if Cosign times out.
         """
         pubkey_config, env_vars, pubkey = CosignValidator.__get_pubkey_config(key)
-
-        env = os.environ.copy()
-        # Extend the OS env vars only for passing to the subprocess below
-        env["DOCKER_CONFIG"] = f"/app/connaisseur-config/{self.name}/.docker/"
-        if safe_path_func(
-            os.path.exists, "/app/certs/cosign", f"/app/certs/cosign/{self.name}.crt"
-        ):
-            env["SSL_CERT_FILE"] = f"/app/certs/cosign/{self.name}.crt"
-        env.update(env_vars)
 
         cmd = [
             "/app/cosign/cosign",
@@ -145,7 +229,7 @@ class CosignValidator(ValidatorInterface):
 
         with subprocess.Popen(  # nosec
             cmd,
-            env=env,
+            env=self.__get_envs().update(env_vars),
             stdin=subprocess.PIPE,
             stdout=subprocess.PIPE,
             stderr=subprocess.PIPE,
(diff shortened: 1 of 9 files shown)


connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -182,6 +203,44 @@ def __get_pubkey_config(key: str):
...
+    @staticmethod
+    def __apply_policy(roots: dict, threshold: int, required: list):
+        """
+        TODO: applies the validation policy
+
+        TODO: Raises Exception if not compliant
+        """
+
+        # verify threshold
+        signed_digests = [k['digests'] for i,k in roots.items() if k['digests'] is not None]
```
phbelitz, line 215:
```suggestion
        signed_digests = [k['digests'] for i,k in roots.items() if not isinstance(k['digests'], Exception)]
        exceptions =  [k['digests'] for i,k in roots.items() if isinstance(k['digests'], Exception)]
```

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -182,6 +203,44 @@ def __get_pubkey_config(key: str):
...
+        # check that same digest present 'threshold' times
+        if not len(set(signed_digests)) == 1 or not len(signed_digests) >= threshold:
+            msg = "Image not compliant to validation policy (threshold of '{threshold}' not reached)"
+            raise ValidationError(
+                message=msg,
+                trust_data_type="dev.cosignproject.cosign/signature",
+                stderr='TODO',
+                threshold=str(threshold),
```
phbelitz, line 223:
```suggestion
                threshold=str(threshold),
                exc="\n".join([str(e) for e in exceptions])
```

phbelitz (commented):
maybe something like this works better for you?

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -37,13 +38,33 @@ def __get_key(self, key_name: str = None):
...
+        if trust_root == "*":
+            roots = {k['name']:{'name': k['name'], 'key': "".join(k['key'])} for k in self.trust_roots}
+        else:
+            roots = self.__get_key(trust_root)
+        logging.warning(roots)
+        for name, root in roots.items():
+            try:
+                roots[name]['digests'] = self.__get_cosign_validated_digests(str(image), root['key']).pop()
+            except Exception as err:
+                roots[name]['digests'] = None
```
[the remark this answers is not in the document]
xopham: good idea, I think I should not use `'digests'` but could go for an additional field `'errors'` that I print later in some form. What would you think?

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -21,44 +22,110 @@
...
+        # if not configured, `threshold` is 1 if trust root is not "*" or
+        # `required` is specified and number of trust roots otherwise
+        threshold = kwargs.get(
+            "threshold",
+            (int(any(required)) or len(self.trust_roots)) if trust_root == "*" else 1,
+        )
+
+        self.vals = self.__get_pinned_keys(trust_root, required)
+
+        tasks = [self.__validation_task(k, image) for k in self.vals.keys()]
```
peterthomassen: You could also do:
```python
tasks = map(lambda k: self.__validation_task(k, image), self.vals.keys())
```

This returns an iterator, preventing in-memory construction of the list.

Now imagine you define a function `self.__validation_task_by_image_callback(image)` that returns a function (!) that only takes `k` as an argument, but returns `self.__validation_task(k, image)`:

```python
def __validation_task_by_image_callback(image):
    return lambda k: self.__validation_task(k, image)  # not sure if this syntax works; think of it as pseudocode
```

Essentially this is the same function, but it has the `image` argument statically bound.

... then you could simply do:

```python
validation_callback = self.__validation_task_by_image_callback(image)
tasks = map(validation_callback, self.vals.keys())
```

This would very clearly show that each item of `keys()` is mapped onto a function call.

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -21,44 +22,110 @@
...
+
+            # check if all `required` are present in trust roots
+            not_found = set(required) - {k["name"] for k in self.trust_roots}
+            if not_found:
+                msg = 'Trust roots "{key_names}" not configured for validator "{validator_name}".'
+                raise NotFoundException(
+                    message=msg,
+                    key_names=", ".join(not_found),
+                    validator_name=self.name,
+                )
```
peterthomassen, line 92: This can go before the `keys` assignment above

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -21,44 +22,110 @@
...
+                    message=msg, key_name=key_name, validator_name=self.name
+                ) from err
+            keys = {
+                key_name: {
+                    "name": key_name,
+                    "key": Key("".join(key)),
+                    "digest": None,
+                    "error": None,
+                }
+            }
```
peterthomassen, line 108: Code duplication from above. Suggesting to construct `keys` only once (via iteration), and do the special-case handling when preparing the iteration input:

```python
trust_roots = self.trust_roots
if key_name != "*":
    trust_roots = filter(lambda key: key["name"] = key_name, trust_roots)
if not trust_roots:
    msg = "..."
    raise ...

return {
    ...
    for key in trust_roots
}
```

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -21,44 +22,110 @@
 class CosignValidator(ValidatorInterface):
     name: str
     trust_roots: list
+    vals: dict
```
peterthomassen, line 27: What does `vals` stand for? If it's just "values", I'd prefer a more telling name; otherwise, please explain in a comment.

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -21,44 +22,110 @@
...
+        Async task for each validation to gather all required validations,
+        execute concurrently and update results.
+        """
+        try:
+            self.vals[trust_root]["digest"] = await self.__get_cosign_validated_digests(
+                image, self.vals[trust_root]
+            )
+        except Exception as err:
+            self.vals[trust_root]["error"] = err
+            logging.info(err)
```
peterthomassen, line 124: It's better to pass `self.vals[trust_root]` into this function, instead of just passing the key. -- If done this way, you might actually be able to get rid of `self.vals` as a class attribute, and keep `vals` local to the function that constructs it.

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -169,6 +252,77 @@ def __get_envs(self):
...
+        Validates the signature verification outcome against the policy for
+        threshold and required trust roots.
+
+        Raises an exception if not compliant.
+        """
+
+        # verify threshold
+        signed_digests = [k["digest"] for k in vals.values() if k["digest"] is not None]
+        # raise exception if the same digest does not appear 'threshold' times
+        if not len(set(signed_digests)) == 1 or not len(signed_digests) >= threshold:
```
peterthomassen, line 284: Under which circumstance could the first condition fail?

docs/validators/sigstore_cosign.md
```diff
@@ -223,6 +233,86 @@ Such environment variables can be injected into Connaisseur via `deployment.envs
     VAULT_TOKEN: secrettoken
 ```
 
+### Multi-signature verification
+
+> :warning: This is currently an experimental feature that might unstable over time.
+> As such, it is not part of our semantic versioning guarantees and we take the liberty to adjust or remove it with any version at any time without incrementing MAJOR or MINOR.
+> Central consideration for the experimental state is a bug in Python's [asyncio module](https://bugs.python.org/issue46364) by which multi-signature validation cannot be carried out concurrently in a secure manner.
```
peterthomassen: That sound like Python has a security bug?

docs/validators/sigstore_cosign.md
```diff
@@ -223,6 +233,86 @@ Such environment variables can be injected into Connaisseur via `deployment.envs
...
+> As a result the current implementation uses sequential execution by which validation time scales linearly with the number of trust roots in scope.
+
+Connaisseur can verify multiple signatures for a single image.
+It is possible to configure a threshold number and specific set of required valid signatures.
+This allows to implement several advanced use cases (and policies):
+
+* Five maintainers of a repository are able to sign a single derived image, however at least 3 signatures are required for the image to be valid.
+* In a CI pipeline, a container image is signed directly after pushing by the build job and at a later time upon passing quality gates such as security scanners or integration tests, each with their own key (trust root). Validation requires all of these signatures for deployment to enforce integrity and quality gates.
+* A mixture of the above use cases whereby several specific trust roots are enforced (e.g. automation tools) and the overall number of signatures has to surpass a certain threshold (e.g. at least one of the testers admits).
+* Key rotation is possible by adding a new key as an additional key and require at least one valid signature.
```
peterthomassen, line 248: Nice! Wouldn't have thought of that use case. That's a cool list!

docs/validators/sigstore_cosign.md
```diff
@@ -223,6 +233,86 @@ Such environment variables can be injected into Connaisseur via `deployment.envs
...
+      -----END PUBLIC KEY-----
+  - name: charlie
+    key: |
+      -----BEGIN PUBLIC KEY-----
+      MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEEHBUYJVrH+aFYJPuryEkRyE6m0m4
+      ANj+o/oW5fLRiEiXp0kbhkpLJR1LSwKYiX5Toxe3ePcuYpcWZn8Vqe3+oA==
+      -----END PUBLIC KEY-----
+```
+
+The trust roots `alice`, `bob`, and `charlie` are all included for verification in case `.policy[*].with.trust_root` is set to `"*"` (note that this is a special flag not a real wildcard):
```
peterthomassen, line 278: `flag, not`

docs/validators/sigstore_cosign.md
```diff
@@ -223,6 +233,86 @@ Such environment variables can be injected into Connaisseur via `deployment.envs
...
+
+```yaml
+- pattern: "*:*"
+  validator: multicosigner
+  with:
+    trust_root: "*"
+    threshold: 2
+```
+
+In this case, valid signatures two or more out of the three trust roots are required for admittance.
```
peterthomassen: broken grammar

helm/values.yaml
```diff
@@ -1,7 +1,7 @@
 # configure Connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v2.4.1
+  image: securesystemsengineering/connaisseur:v2.5.0
```
peterthomassen, line 4: Why is this a MINOR increase? (It's experimental, so a change to the feature would not trigger a MINOR change. Why would its introduction?)

Also, the version change should be a separate commit, so that it can be easily adjusted if the PR gets delayed beyond another release, without touching the feature commit.

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -45,8 +47,18 @@ async def validate(
...
+        # use concurrent.futures for now
+        # tasks = [self.__validation_task(k, image) for k in self.vals.keys()]
+        # await asyncio.gather(*tasks)
+
+        # prepare executor
+        num_workers = len(self.vals)
+        executor = ThreadPoolExecutor(num_workers)
+        # prepare tasks
+        arguments = [(k, image) for k in self.vals.keys()]
+        futures = [executor.submit(self.__validation_task, *arg) for arg in arguments]
```
peterthomassen, line 59: Well, then the callable thing I commented on the other commit is not needed.

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -45,8 +47,18 @@ async def validate(
...
+        # await results (output dropped as `self.vals` is updated within function)
+        _ = [future.result() for future in futures]
```
peterthomassen, line 61: Why do an assignment?

Seems like you intend to do:
```python
for future in futures:
    future.result()
```

docs/validators/sigstore_cosign.md
```diff
@@ -308,7 +308,18 @@ Using the `required` key, it is possible to enforce specific trusted roots:
...
+```yaml
+- pattern: "*:*"
+  validator: multicosigner
+  with:
+    trust_root: "*"
+    threshold: 3
+    required: ["alice", "bob"]
+```
+
+Thus, at least 3 valid signatures are required, however among `alice` and `bob` must be among those.
```
peterthomassen, line 322: `among` twice

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -169,6 +252,77 @@ def __get_envs(self):
...
+        Raises an exception if not compliant.
+        """
+
+        # verify threshold
+        signed_digests = [k["digest"] for k in vals.values() if k["digest"] is not None]
+        # raise exception if the same digest does not appear 'threshold' times
+        if not len(set(signed_digests)) == 1 or not len(signed_digests) >= threshold:
+            # simply raise the specific error if single specified trust root
+            if len(vals) == 1:
+                raise list(vals.values()).pop()["error"]
```
peterthomassen, line 287: `vals.values()[0]["error"]`?

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -169,6 +252,77 @@ def __get_envs(self):
...
+            )
+            raise ValidationError(
+                message=msg,
+                trust_data_type="dev.cosignproject.cosign/signature",
+                threshold=str(threshold),
+                required=required,
+                errors=errs,
+            )
+
+        digest = set(signed_digests).pop()
```
peterthomassen, line 310: `digest = signed_digests[0]`

peterthomassen (changes requested):

connaisseur/validators/cosign/cosign_validator.py
xopham: I had previously constructed a single-line solution and pylint was complaining about unassigned outcome

connaisseur/validators/cosign/cosign_validator.py
xopham: would fail if the list of signed digests contains different digests or none or if the number of digests is less than threshold

docs/validators/sigstore_cosign.md
xopham: thanks :slightly_smiling_face:

helm/values.yaml
xopham: you are correct. Opened #562 to update versions in accordance with release scope

connaisseur/validators/cosign/cosign_validator.py
xopham: took a bit of different path...let me know what you think

connaisseur/validators/cosign/cosign_validator.py
xopham: actually requires `list(vals.values())[0]["error"]` as `.values()` returns an iterator

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -64,48 +64,46 @@ async def validate(
...
-            # check if all `required` are present in trust roots
-            not_found = set(required) - {k["name"] for k in self.trust_roots}
-            if not_found:
-                msg = 'Trust roots "{key_names}" not configured for validator "{validator_name}".'
-                raise NotFoundException(
-                    message=msg,
-                    key_names=", ".join(not_found),
-                    validator_name=self.name,
-                )
+            if len(required) >= threshold:
```
peterthomassen, line 78: 1. How can the threshold be lower than the number of required signatures?
2. `len` and `>=` doesn't work if either side is `None`.

peterthomassen (commented):
I like the new approach in `__get_pinned_keys`!

connaisseur/validators/cosign/cosign_validator.py
xopham: 1. as implemented in the `validate` method (you also made a proposal for that), I chose for implementation that `threshold` is 1 if `required` contains a list of required trust roots and there is no direct specification of `threshold`. that means if you only provide a setting for `required` in `helm/values.yaml`, threshold is set to 1, however, connaisseur will enforce those 2 signatures. Here, it is used to shorten the required number of validations if the validator contains a lot of trust roots but only a few are actually required and no other threshold is used. I think I also had an actual reason which goes beyond throwing the exception in __apply_policy, but cannot recall rn
2. based on `validate` method neither can be `None`, but I agree that this is misleading with the defaults. Would you advise to drop the defaults?

connaisseur/validators/cosign/cosign_validator.py
xopham: just noticed the change got lost

connaisseur/validators/cosign/cosign_validator.py
peterthomassen: 2. Yes!

xopham: thx for advice. changed

connaisseur/validators/cosign/cosign_validator.py
xopham: fixed as per comment and change below

connaisseur/validators/cosign/cosign_validator.py
xopham: in a stupid way, but now it works :stuck_out_tongue_winking_eye:

connaisseur/validators/cosign/cosign_validator.py
xopham: decided to keep it as is in order to maintain easier results handling. updating gets rid of result collection and re-arranging

phbelitz (approved):

PR #437, opened 2021-12-03
feat: cosign private registry

### LINKED ISSUE #324: Enable selfsigned certs for cosign OCI registries
Extend the `cert` flag for cosign
---

Hi there,
I could sign (cosign) some images in my nexus.test.com:5000 self hosted registry that is using a self signed certificate. I can verify with cosign on my laptop. Now I would like to verify it in the kubernetes cluster using connaisseur.
Unfortunately, I am getting this
```
x509: certificate is valid for connaisseur-svc .... connaisseur-svc.connaisseur.svc.cluster.local, not nexus.test.com
```
I think I need to provide the root cert used to generate the nexus.test.com cert, can't find a way to do that within the values.yaml. There is this `cert: |` section under the cosign validators I configured, but I didn't get much luck using it.

_Originally posted by @joelapatatechaude in https://github.com/sse-secure-systems/connaisseur/discussions/320_

phbelitz:
For cosign validators, a new field (`cert`) has been added so that cosign signatures can be validated against registries with selfsigned certificates.

fixes #324

### DIFF
--- a/helm/templates/_helpers.tpl
+++ b/helm/templates/_helpers.tpl
@@ -6,6 +6,7 @@ Expand the name of the chart.
 {{- default .Chart.Name .Values.nameOverride | trunc 63 | trimSuffix "-" -}}
 {{- end -}}
 
+
 {{/*
 Create a default fully qualified app name.
 We truncate at 63 chars because some Kubernetes name fields are limited to this (by the DNS naming spec).
@@ -23,6 +24,7 @@ If release name contains chart name it will be used as a full name.
 {{- end -}}
 {{- end -}}
 {{- end -}}
+
 
 {{/*
 Create chart name and version as used by the chart label.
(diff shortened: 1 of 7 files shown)


xopham (commented):
Lgtm! Was it tested? If I change the cert, a helm upgrade will restart the pods?

xopham:
Note that there is also an insecure registry flag for cosign: https://github.com/sigstore/cosign/issues/311
I tend to like a selfsigned cert better though

phbelitz:
I tested this on a local harbor registry.

helm/templates/_helpers.tpl
```diff
@@ -139,3 +142,36 @@ Extract Kubernetes Minor Version.
 {{ include (print $.Template.BasePath "/env.yaml") . }}
 {{ include (print $.Template.BasePath "/alertconfig.yaml") . }}
 {{- end -}}
+
```
xopham, line 145: one too many empty lines

xopham (approved):
one minor finding otherwise lgtm

xopham (approved):
lgtm

PR #485, opened 2022-01-14
ci: removed stress test

phbelitz:
The stress test was supposed to test the performance of the different server implementations of connaisseur (cheroot vs uwsgi vs ...). Since we settled with cheroot for now, the stress test isn't needed anymore for the regular pipeline run, especially because it's result was discarded anyways. In order to have a "successful" pipeline again, the stress test was removed.

## Checklist

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -202,7 +202,6 @@ jobs:
       matrix:
         integration-test-arg:
           [
-            "stress-test",
             "regular",
             "cosign",
             "namespace-val",
--- a/.github/workflows/nightly-scans.yaml
+++ b/.github/workflows/nightly-scans.yaml
@@ -1,14 +1,19 @@
 name: nightly-scans
 
 on:
+  pull_request:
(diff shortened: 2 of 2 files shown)


Starkteetje:
Well, it was not only for comparing servers. It was also for measuring stability. As such it does still make sense. At least, I've never had the stress-test fail so far since it was merged to dev, so I don't see the reason to remove it (though it takes a ton of time, so splitting it into its two parts would make sense)

phbelitz:
Ok, I moved the stress test into the nightly run, where the long runtime isn't an issue. This way we could still observe any stability issues.

Starkteetje:
Then we'd have feedback only after a problematic merge to whatever branch nightly runs on (I'd guess master, in which case the feedback would be VERY late and it wouldnt be easy to discern what PR the degradation comes from. Maybe https://github.com/sse-secure-systems/connaisseur/pull/486 fixes your issue with time?

phbelitz:
closing in favor of #486

PR #501, opened 2022-01-24
feat: silent healthz endpoints logging

sf-jmarcou:
## Issue
In INFO loglevel, the application healthz endpoints prints lot of logs (because of K8s live/ready-ness) which make hard to read the logs and consume logging storage.
A solution would be to disable the INFO loglevel, but this hides the genuine logs such as image verification operations.

Some flask module suggest to disable the logs on these endpoints, for example: https://pypi.org/project/flask-healthz/, using `no_log` option.

## Description
This PR changes the loglevel of the HTTP requests handled by the healthz endpoints to DEBUG, to avoid cluttering the web log files with automated requests.

### DIFF
--- a/connaisseur/logging_wrapper.py
+++ b/connaisseur/logging_wrapper.py
@@ -31,7 +31,10 @@ class ConnaisseurLoggingWrapper:
         result = self.app(environ, custom_start_response)
         # the server can "change it's mind" before sending the first response if exc_info is provided
         # hence we need to use the last status code provided
-        self.logger.info(_format_log(status_codes[-1], environ))
+        if environ.get("REQUEST_URI") in ["/ready", "/health"]:
+            self.logger.debug(_format_log(status_codes[-1], environ))
+        else:
+            self.logger.info(_format_log(status_codes[-1], environ))
         return result
 
 


xopham (commented):
I see where you are coming from and agree that it clutters the logs. On the other hand, it seems wrong to drop it alltogether. Another option would be to change the loglevel of health to `debug`. Would need to check what is recommended/best way forward here.

sf-jmarcou:
That sounds a good idea to me!
I am happy to update the PR to make it at DEBUG level instead of disabling.
Just let me know.

Starkteetje:
Hi @sf-jmarcou 
I'm with @xopham on this and wouldn't drop them completely. While right now they're kind of devoid of meaning, in the past at least /ready had its uses (when we checked health of the configured notary) and might get it again in the future. But I definitely see the pain, so I'd suggest we do solve it via DEBUG level. If you change it and rebase, we'd be happy to merge :)

sf-jmarcou:
Hi @Starkteetje, make total sense. I'll update the PR ASAP. Cheers.

xopham:
@sf-jmarcou feel free to let us know if you need any help. We aim to cut a feature release next week and we'd be happy to include this :slightly_smiling_face:

xopham (approved):
@sf-jmarcou made the minor change to set `/health` endpoint loglevel to DEBUG and think it should be ready to merge.
Thanks for the awesome contribution :pray:

Issue #496, opened 2022-01-21
Refactor helm deployment


annekebr:
When touching the helm deployment the next time, please consider the following things:

- The way how we allow configurations of Kubernetes resource settings differs from resource to resource. We have an own section `service` to configure the Kubernetes `Service`, but configuration for the webhook is handled in the `deployment` section (which I would have assumed to cover only settings for the Kubernetes `Deployment` resource judging from the additional section for the `Service` resource. I'd suggest to decide for one option for consistency reasons.

- Why do we allow the `Service` type to be configurable? Did we ever test Connaisseur with another type than `ClusterIP`? Are we sure that our security assumptions hold for all service types equally?

- In certain circumstances (with a high probability), Connaisseur deploys empty resources (e.g. `env` or `alertconfig` secret, there may be more). Are we sure that we want that? -> https://github.com/sse-secure-systems/connaisseur/pull/958 ✔️ 

- Do we want to explicitly enable features by adding an extra yaml subfield `enabled: true` like we do for e.g. `automaticChildApproval`? However we decide, I'd suggest to do it consistently - right now, `detectionMode` behaves differently than `automaticChildApproval` for example. -> https://github.com/sse-secure-systems/connaisseur/pull/984 ✔️ 

- Tying into the point above: How do we want to handle keys, that are commented out? Do we want to be able to achieve the same behaviour with explicit values? Right now, there is no way for e.g. alerting to be turned off if not having the config block commented out.

- We once decided to allow no trust root in `notaryv1` validator instances for the preconfig to work. This results in inconsistent trust root validations. I'd like to question the decision again, maybe we should rethink that.

- [_Naming_] Alerting has a `templates` list. I'd suggest to rename it to `receivers`. -> https://github.com/sse-secure-systems/connaisseur/pull/984 ✔️ 

- ~[_Naming_] Should the `policy` section rather be `policies`?~ -> https://github.com/sse-secure-systems/connaisseur/issues/496#issuecomment-1495810148

xopham:
Consider deprecating `default` for trust roots in light of #428

xopham:
Consider dropping dockerhub official public key

annekebr:
See also https://github.com/sse-secure-systems/connaisseur/issues/383 - might become more relevant when users can configure their own side cars etc.

xopham:
normalize naming either camel-case or snake case

annekebr:
[...]

PR #508, opened 2022-02-01
feat: prometheus metrics

masaruhoshi:
## Description

Introduces simple metrics for Flask application using prometheus-flask-exporter for mutate endpoints. The metrics are available from the /metrics endpoint.

It also introduces the possibility to specify a different port for the mutating webhook based on the service port.

## Checklist

- [-] Extended README/Documentation (if necessary)

### DIFF
--- a/connaisseur/flask_application.py
+++ b/connaisseur/flask_application.py
@@ -4,6 +4,7 @@ import os
 import traceback
 
 from flask import Flask, jsonify, request
+from prometheus_flask_exporter import PrometheusMetrics, NO_PREFIX
 
 from connaisseur.admission_request import AdmissionRequest
 from connaisseur.alert import send_alerts
@@ -24,6 +25,15 @@ sends its response back.
 CONFIG = Config()
 DETECTION_MODE = os.environ.get("DETECTION_MODE", "0") == "1"
 
+metrics = PrometheusMetrics(
+    APP,
+    defaults_prefix=NO_PREFIX,
+    buckets=(0.1, 0.25, 0.5, 0.75, 1.0, 2.5, 5.0, 10.0, 15.0, 20, 30.0, float("inf")),
+)
+"""
+Provides metrics for the Flask application
+"""
+
 
 @APP.errorhandler(AlertSendingError)
 def handle_alert_sending_failure(err):
@@ -41,6 +51,14 @@ def handle_alert_config_error(err):
 
 
 @APP.route("/mutate", methods=["POST"])
+@metrics.counter(
+    "mutate_requests_total",
+    "Total number of mutate requests",
+    labels={
+        "allowed": lambda r: metrics_label(r, "allowed"),
+        "status_code": lambda r: metrics_label(r, "status_code"),
+    },
+)
 def mutate():
     """
     Handle the '/mutate' path and accept CREATE and UPDATE requests.
@@ -74,8 +92,19 @@ def mutate():
     return jsonify(response)
 
 
+def metrics_label(response, label):
+    json_response = response.get_json(silent=True)
+    if json_response:
+        if label == "allowed":
+            return json_response["response"]["allowed"]
+        elif label == "status_code":
+            return json_response["response"]["status"]["code"]
+    return json_response
+
+
 # health probe
 @APP.route("/health", methods=["GET", "POST"])
+@metrics.do_not_track()
 def healthz():
     """
     Handle the '/health' endpoint and check the health status of the web server.
@@ -87,6 +116,7 @@ def healthz():
 
 # readiness probe
 @APP.route("/ready", methods=["GET", "POST"])
+@metrics.do_not_track()
 def readyz():
     return "", 200
 
--- a/docs/README.md
+++ b/docs/README.md
@@ -17,9 +17,10 @@ It is built to be extendable and currently aims to support the following signing
 
 It provides several additional features:
 
+- [Metrics](features/metrics.md): *get prometheus metrics at `/metrics`*
+- [Alerting](features/alerting.md): *send alerts based on verification result*
 - [Detection Mode](features/detection_mode.md): *warn but do not block invalid images*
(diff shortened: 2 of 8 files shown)


tluimes:
Addresses feature issue: https://github.com/sse-secure-systems/connaisseur/issues/511

docs/README.md
```diff
@@ -139,6 +139,83 @@ Otherwise, Connaisseur translates all image references in the original request t
...
+# HELP mutate_requests_total Total number of mutate requests
+# TYPE mutate_requests_total counter
+mutate_requests_total{allowed="False",status_code="403"} 4.0
+mutate_requests_total{allowed="True",status_code="202"} 5.0
+# HELP mutate_requests_created Total number of mutate requests
+# TYPE mutate_requests_created gauge
+mutate_requests_created{allowed="False",status_code="403"} 1.643760946491879e+09
+mutate_requests_created{allowed="True",status_code="202"} 1.6437609592007663e+09
+```
+
```
xopham, line 218: would prefer to add this to the docs as a feature such as `alerting`, see e.g. [here](https://sse-secure-systems.github.io/connaisseur/v2.4.0/features/). I can make the basic changes to setup the docs structure...

xopham (commented):
@masaruhoshi @tluimes great work! Really like the feature :rocket: 

I have some comments on e.g. the service port (which I think is valid, but not well scoped to the PR), the docs update (will provide basic infrastructure)...

Thanks for the contribution :pray:

helm/templates/certificate_webhook-conf.yaml
```diff
@@ -71,6 +71,7 @@ webhooks:
         name: {{ .Chart.Name }}-svc
         namespace: {{ .Release.Namespace }}
         path: /mutate
+        port: {{ .Values.service.port }}
```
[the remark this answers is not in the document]
Starkteetje: I'm not aware of more thoughts around port or service than what's in #496 :man_shrugging:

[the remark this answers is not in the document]
xopham: ok, then let's move this to a separate PR with respective scope @masaruhoshi

docs/README.md
xopham: have added a new section for the metrics :man_technologist: if you want to get a preview for testing, you can use the following btw:
```bash
docker run --rm -it -p 8000:8000 -v ${PWD}:/docs squidfunk/mkdocs-material
```

helm/templates/certificate_webhook-conf.yaml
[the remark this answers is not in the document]
masaruhoshi: it makes sense. Sorry about that, guys. I'll move it to another PR

docs/README.md
masaruhoshi: thanks you, @xopham

xopham: just made another fixup to add it to the features/overview. hope i did not intercept your work :wink:

connaisseur/flask_application.py
```diff
@@ -24,6 +25,15 @@
 CONFIG = Config()
 DETECTION_MODE = os.environ.get("DETECTION_MODE", "0") == "1"
 
+metrics = PrometheusMetrics(
+    APP,
+    defaults_prefix=NO_PREFIX,
+    buckets=(0.1, 0.25, 0.5, 0.75, 1.0, 2.5, float("inf")),
```
xopham, line 31: I am not 100% certain about the buckets, but if I see that correctly, it will allow to get a request duration histogram, right?
Essentially, it should sort requests into buckets based on their duration, right?

If that is so, I would propose to adjust the default buckets. The maximum response duration of an [admission webhook is limited to 30s](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#timeouts). Thus, the following may make more sense to cover the range:

```suggestion
    buckets=(0.1, 0.25, 0.5, 1.0, 2.5, 5.0, 10.0, 15.0, 20.0, float("inf")),
```
thus, we'd be covering the whole range and see how close we get to timeout

xopham (commented):
generally, looks good to me know. I have one remaining comment. Once that is resolved and the `pytests` are fixed, I'd be happy to merge it.
There is some help on how to run pytest locally: https://sse-secure-systems.github.io/connaisseur/develop/CONTRIBUTING/#test-changes

Let me know if you need any help :slightly_smiling_face:

connaisseur/flask_application.py
masaruhoshi: > Essentially, it should sort requests into buckets based on their duration, right?
Yep! That's exactly that.

Usually, I suggest buckets created based on the general assumption of latency. In this case, anything taking between 0 <= x<= 2.5 seconds is considered "acceptable" and > 2.5s would be "too slow". That's a personal preference, But I have no issues adding more buckets.

xopham: I see your point and agree for a common request, but in this case Connaisseur has to perform quite some verification and both `notary` as well as `cosign` take some time to verify the signatures. From personal experience, that can easily take a few seconds.
However, I might be wrong here as latency might just mean the time it takes for the request to reach the hook and I am probably talking about processing time as in `latency + processing time = response time` as discussed [here](https://stackoverflow.com/questions/58082389/what-is-the-difference-between-latency-and-response-time).

masaruhoshi: @xopham, I agree with your points. I also added the `0.75` to the list you suggested 👍

masaruhoshi:
@xopham, @tluimes and I had a few backs and forth fighting with the tests but we finally managed to get everything green. Thanks for the great help, @tluimes

xopham (dismissed):
lgtm! thanks for the great work. Very much appreciate the contribution :pray:

xopham (approved):
preparing merge: squashed :hammer_and_wrench:  and signed :writing_hand:

PR #510, opened 2022-02-02
fix: corrected typo in ClusterRole helm template

hsuchan:
# General summary
"replicacontrollers" is not a valid kubernetes object.
Authors probably meant "replicationcontrollers".

## Description
In my connaisseur pods, I see error logs such as:
"requests.exceptions.HTTPError: 403 Client Error: Forbidden" for replicationcontrollers api calls.
After editing clusterrole connaisseur-role, replacing "replicacontrollers" with "replicationcontrollers", and doing a rolling restart, those errors msg went away.

## Checklist

- [x] PR is rebased to/aimed at branch `develop`
- [x] PR follows [Contributing Guide](../docs/CONTRIBUTING.md)
- [ ] Added tests (if necessary)
- [x] Extended README/Documentation (if necessary)
- [x] Adjusted versions of image and Helm chart in `values.yaml` and `Chart.yaml` (if necessary)

PR #514, opened 2022-02-04
fix: updated REST path for core api group

### LINKED ISSUE #513: Issue with parent resource of type pod
**Describe the bug**
Connaisseur is making invalid API calls when checking for parent resource, when the parent resource is of type pod.

**Detailed description**
We are using the kuberhealthy project (https://github.com/kuberhealthy/kuberhealthy), which is an operator for synthetic monitoring on Kubernetes. Kuberhealthy schedules checks (i.e. pods) at specified intervals, those pods run some monitoring tasks, and those pods terminate with a "Completed" status.

Checking for the parent of any of those temporary monitoring pods show:
```
kubectl get po dns-status-internal-1643848976 -n kuberhealthy -o json | jq ".metadata.ownerReferences"
[
  {
    "apiVersion": "v1",
    "kind": "Pod",
    "name": "kuberhealthy-7d47c7c9d6-m2skw",
    "uid": "f7852c4f-78bc-4573-b72b-1353104d7ba8"
  }
]
```

In connaisseur logs you will see the following entry:
```
requests.exceptions.HTTPError: 403 Client Error: Forbidden for url: https://10.233.0.1:443/apis/v1/namespaces/kuberhealthy/pods/kuberhealthy-7d47c7c9d6-m2skw
```

This error comes from:
https://github.com/sse-secure-systems/connaisseur/blob/eaa16437426d1dab5335d995ed952bf13d46acfa/connaisseur/workload_object.py#L61-L62

However this is not a valid API call.
```
curl -s $APISERVER/apis/v1 --header "Authorization: Bearer $TOKEN" --cacert /tmp/ca.crt
404 page not found
```

My EKS 1.21 cluster shows the following api-resources:

```
pods                              po                       v1                                          true         Pod
replicationcontrollers            rc                       v1                                          true         ReplicationController
daemonsets                        ds                       apps/v1                                     true         DaemonSet
deployments                       deploy                   apps/v1                                     true         Deployment
replicasets                       rs                       apps/v1                                     true         ReplicaSet
statefulsets                      sts                      apps/v1                                     true         StatefulSet
cronjobs                          cj                       batch/v1beta1                               true         CronJob
jobs                                                       batch/v1                                    true         Job
```
RestAPI calls targeting the `v1` apiVersion (core group) typically use `https://$APISERVER/api/...` instead of `https://$APISERVER/apis/...`

I suspect the same error impacts replicationcontrollers objects.

hsuchan:
## General summary
Connaisseur is making invalid API calls when checking for parent resource, when the parent resource is of type pod.

Fixes #513 

## Description
Adds rest_path variable that takes either following values:
- "api" if the apiVersion is "v1"
- "apis" otherwise

... to make sure the proper API call is made

## Checklist

### DIFF
--- a/connaisseur/workload_object.py
+++ b/connaisseur/workload_object.py
@@ -58,8 +58,13 @@ class WorkloadObject:
             name = owner["name"]
             uid = owner["uid"]
 
+            if api_version == "v1":
+                rest_path = "api"
+            else:
+                rest_path = "apis"
+
             parent = k_api.request_kube_api(
-                f"apis/{api_version}/namespaces/{self.namespace}/{kind}/{name}"
+                f"{rest_path}/{api_version}/namespaces/{self.namespace}/{kind}/{name}"
             )
 
             if parent["metadata"]["uid"] != uid:


xopham (approved):
@hsuchan great catch!
Bug was also reproduced during testing of #510 and not only affects the `pod` but also the `replicationset` api group.
Issue wasn't caught within the integration tests, as those currently only confirm the resources as allowed, but fail to check whether all child resources spin up. We raise this in #516 .

Great work! Thx :pray:

xopham (approved):

PR #518, opened 2022-02-04
feat: add reinvocationPolicy option

caiconkhicon:
Add a value in the Helm chart to set `reinvocationPolicy` of `MutatingWebhookConfiguration`

## Description

It should be possible to set `reinvocationPolicy` (https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#reinvocation-policy) for the MutatingWebhookConfiguration resource. 

My use case is:
- I want to make sure that only images from my private registry (which are signed) are used in my cluster
- Thus, for public images, I pull them, sign and then push them to my private registry
- I use `kyverno` to mutate all public images to my private images (e.g. `bitnami/bitnami-shell:10-debian-10-r99` become `my.private.registry/bitnami/bitnami-shell:10-debian-10-r99`
- So, I need `reinvocationPolicy: IfNeeded` to let Connaisseur reevaluate the request after it was mutated by `kyverno`.

## Checklist

### DIFF
--- a/helm/templates/certificate_webhook-conf.yaml
+++ b/helm/templates/certificate_webhook-conf.yaml
@@ -33,6 +33,7 @@ metadata:
 webhooks:
   - name: {{ .Chart.Name }}-svc.{{ .Release.Namespace }}.svc
     failurePolicy: Ignore
+    reinvocationPolicy: {{ .Values.deployment.reinvocationPolicy | default "Never" }}
     clientConfig:
       service:
         name: {{ .Chart.Name }}-svc
@@ -61,11 +62,8 @@ metadata:
     "helm.sh/hook": post-install, post-upgrade, post-rollback
 webhooks:
   - name: {{ .Chart.Name }}-svc.{{ .Release.Namespace }}.svc
-    {{- if .Values.deployment.failurePolicy }}
-    failurePolicy: {{ .Values.deployment.failurePolicy }}
-    {{- else }}
-    failurePolicy: Fail
-    {{- end }}
+    failurePolicy: {{ .Values.deployment.failurePolicy | default "Fail" }}
+    reinvocationPolicy: {{ .Values.deployment.reinvocationPolicy | default "Never" }}
     clientConfig:
       service:
         name: {{ .Chart.Name }}-svc
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -8,6 +8,12 @@ deployment:
   # imagePullSecrets:
   # - name: "my-container-secret"
   failurePolicy: Fail  # use 'Ignore' to fail open if Connaisseur becomes unavailable
+  # Use 'reinvocationPolicy: IfNeeded' to be called again as part of the admission evaluation if the object
+  # being admitted is modified by other admission plugins after the initial webhook call
+  # Note that if Connaisseur is invoked a second time, the policy to be applied might change in between.
+  # Make sure, your Connaisseur policies are set up to handle multiple mutations of the image originally
+  # specified in the manifest, e.g. my.private.registry/image:1.0.0 and my.private.registry/image@sha256:<hash-of-1.0.0-image>
+  reinvocationPolicy: Never
   resources:
     limits:
       cpu: 1000m
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -55,12 +55,15 @@ The upper `deployment` section offers some general Kubernetes typical configurat
 Noteworthy configurations are:
 
 - `deployment.failurePolicy`: Failure policy allows configuration whether the mutating admission webhook should fail closed (`Fail`, *default*) or open (`Ignore`) should the Connaisseur service become unavailable. While Connaisseur is configured to be secure by default, setting the [failure policy](https://kubernetes.io/docs/reference/access-authn-authz/extensible-admission-controllers/#failure-policy) to `Ignore` allows to prioritize cluster access[^1].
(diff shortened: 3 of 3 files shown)


helm/templates/certificate_webhook-conf.yaml
```diff
@@ -33,6 +33,11 @@ metadata:
 webhooks:
   - name: {{ .Chart.Name }}-svc.{{ .Release.Namespace }}.svc
     failurePolicy: Ignore
+    {{- if .Values.deployment.reinvocationPolicy }}
+    reinvocationPolicy: {{ .Values.deployment.reinvocationPolicy }}
```
Starkteetje, line 37: I think this could be simplified to `{{ .Values.deployment.reinvocationPolicy | default "Never" }}`

helm/values.yaml
```diff
@@ -8,6 +8,9 @@ deployment:
   # imagePullSecrets:
   # - name: "my-container-secret"
   failurePolicy: Fail  # use 'Ignore' to fail open if Connaisseur becomes unavailable
+  # Use 'reinvocationPolicy: IfNeeded' to be called again as part of the admission evaluation if the object
```
Starkteetje, line 11: Not having thought on it for a long time, it may lead to different policies being executed by Connaisseur. I _assume_
```
policy:
  - pattern: "*:*"
    validator: deny
  - pattern: "my-image:my-tag"
    validator: allow
```
would mutate to `my-image@sha` in the first invocation and deny the second. If that's the case, it should be mentioned in the comment for the option that applied policies may change

helm/templates/certificate_webhook-conf.yaml
caiconkhicon: I agree. I just want to follow the same style that are being used for `failurePolicy`

helm/values.yaml
caiconkhicon: Oh yes, I didn't think about this situation. Do you have any suggestion?

caiconkhicon: Shouldn't Connaisseur "automatically" translate from the tag to the SHA and allow the SHA?

helm/templates/certificate_webhook-conf.yaml
Starkteetje: Ah, makes sense. I'd vote for refactoring the failurePolicy expression unless @phbelitz has a different opinion

helm/values.yaml
Starkteetje: It currently cannot and I would think Connaisseur can conceptually not handle this:
As an example, lets say you have a policy for `*:debug` which always denies since you don't want debug images in your cluster and a policy for `*@sha:123` to always allow as part of a allowlisting baseimages.
If presented with the `*@sha123` image, tagged with `debug` and put into a k8s manifest in the tag form, Connaisseur has to honor your decision to deny it, whereas if you had deployed the image@sha it has to honor your decision to allowlist the image.

For your particular problem, iirc you can also sign digests (at least our notary v1 impl has `__search_image_targets_for_digest` and I think cosign does it mainly), so you could for each incoming image sign both tag and digest and make the policy the same for both tags and digests, i.e. `pattern: yourregistryandimage:*`

caiconkhicon: So my PR here is still useful, right? What I need to do is to make sure that both `my.private.registry/image:tag` ang `my.private.registry/image@sha256:hash-of-tag` are signed, right?

Starkteetje: Yes. My comment wasn't meant to dissuade the PR, I do see the use case. More along the line of, the comment for the reinvocationPolicy should highlight the problem that can arise

Maybe `Note that if Connaisseur is invoked a second time, the policy to be applied might change in between. Make sure, your policies are set up to handle multiple mutations of the image originally specified in the manifest.` or something

helm/templates/certificate_webhook-conf.yaml
phbelitz: I agree with @Starkteetje and we should go with using the `default` template function. We'll probably rework the helm chart at some point and swicth styles for `failurePolicy` and the likes as well.

caiconkhicon: ok, let me update it 
Renault ZOE INTENS 41kWh zzgl. Batterie Miete KLIMAAUTO
7 / 29 Vergrößern

caiconkhicon: ok, let me update it

caiconkhicon: updated for both `failurePolicy` and `reinvocationPolicy` :)

caiconkhicon:
@Starkteetje @phbelitz : I updated the PR following your recommendations. Is there anything else I need to do to get the PR approved?

xopham:
@caiconkhicon generally looks good to me. Before merging, think it makes sense to make a small note on this in the docs [here](https://sse-secure-systems.github.io/connaisseur/v2.4.1/basics/#configure) just below the comment on the `failurePolicy`. Please keep it in a similar form and link to the external documentation of `reinvocationPolicy`.

Very nice contribution. I like it!

caiconkhicon:
> @caiconkhicon generally looks good to me. Before merging, think it makes sense to make a small note on this in the docs [here](https://sse-secure-systems.github.io/connaisseur/v2.4.1/basics/#configure) just below the comment on the `failurePolicy`. Please keep it in a similar form and link to the external documentation of `reinvocationPolicy`.
> 
> Very nice contribution. I like it!

Hi, done :)

xopham (changes requested):
@caiconkhicon looks awesome! If you squash all changes and rebase to develop, I am happy to approve and merge. Will definitely be added to the next release!

Thanks for the great contribution :pray:

xopham (approved):

Issue #99, opened 2021-03-05
Support for rsa, rsa-x509 and ecdsa-x509 on targets keys


victorfrancax1:
Hey guys,

We're implementing connaisseur at our company, and we noticed that it doesn't support the use of `rsa`, `rsa-x509` or`ecdsa-x509` keys, only `ecdsa`, in delegation roles, which correspond to targets in Notary's key hierarchy. Is there any particular reason for that?

Our understanding is that these should be supported, since it's stated at Content Trust's official [tutorial](https://docs.docker.com/engine/security/trust/trust_delegation/#manually-generating-keys) that both cryptosystems could be used. The example in the documentation not only uses `rsa`, but its `x509` form, in order to add the signer.

If this is something that cannot be implemented right now, I think it would be nice if connaisseur's documentation had this information stated somewhere. We had to manually compare our `targets.json` with [targets_schema.json]( https://github.com/sse-secure-systems/connaisseur/blob/master/connaisseur/res/targets_schema.json) to find what we were doing wrong when trying to verify a signature.

Thanks in advance!

phbelitz:
Hi there! Thank you very much for the issue. 

The reason we don't support `rsa`, `rsa-x509` and `ecdsa-x509` (yet) is because the default key generation using Docker Content Trust only uses `ecdsa` and we didn't anticipate someone using other means on creating the keys. But that doesn't mean it has to stay that way! We'll gladly look into the issue and rollout a new release, supporting the new key types at some later time. 

If you want, you can also prepare a pull request to speed things up. I'd be very excited to see one! Cheers.

xopham:
also mentioned in #246

MageshSrinivasulu:
@phbelitz Is the support for RSA rolled out ?

Starkteetje:
With the release 3.4.0, support for all TUF key types through the Go notary package was added

PR #540, opened 2022-02-11
feat: key interface

phbelitz:
This commit adds an easy way to add additional key types into connaisseur. The first set of supported keys are: ECDSA, RSA[cosign], KMS[cosign] and Keyless[cosign] ande each gets automatically recognized.

## Description

We eventually want to support more types of keys and even apply additional logic to specific types, while also have a common interface so that the types can be easily exchanged. 

## Checklist


connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -158,29 +147,15 @@ def __invoke_cosign(self, image, key):
...
-        except ValueError:
-            pass
-
-        # key is KMS reference
-        if re.match(r"^\w{2,20}://[\w:/-]{3,255}$", key):
-            return ["--key", key], {}, b""
-
-        msg = "Public key (reference) '{input_str}' does not match expected patterns."
-        raise InvalidFormatException(message=msg, input_str=key)
+    def __get_envs(self):
```
xopham, line 150: maybe docstring

xopham (commented):
some first minor comments

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+from connaisseur.image import Image
+from connaisseur.exceptions import NoSuchClassError, WrongKeyError
+
+KMS_REGEX = r"(?:awskms|gcpkms|azurekms|hashivault|k8s):\/{2,3}[a-zA-Z0-9_.+\/:-]+"
+KEYLESS_REGEX = r"^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$"
+
+
+class Key:
+    """
+    Abstract Key class, containing the public key used for verifying the image signatures.
```
Starkteetje, line 19: The class does not necessarily contain the public key, does it? Afaik the idea of KMS is to have a reference, not the key itself, so this is probably false.

Does the key abstraction necessitate that the key is a public one? Couldn't, using KMS, there be an HMAC key?

Maybe rather in the direction of `Key object represents a key and can be used in conjunction with a validator to verify objects in that validator's domain.`

Starkteetje: Additionally from [PEP257](https://www.python.org/dev/peps/pep-0257/): `The docstring for a class should summarize its behavior and list the public methods and instance variables. If the class is intended to be subclassed, and has an additional interface for subclasses, this interface should be listed separately (in the docstring).`

Not sure what exactly that entails, but I would highlight that each subclass needs to implement verify, which kind of ties into making it an abstractmethod as per comment above

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+        try:
+            key_cls = Key.__get_key_type_cls(data)
+            return super(Key, cls).__new__(key_cls)
+        except Exception as err:
+            msg = "Unable to find find class {class_name}."
+            raise NoSuchClassError(message=msg, class_name=key_cls) from err
+
+    @staticmethod
+    def __get_key_type_cls(data: str):
+        # key gets automatically identified
```
Starkteetje, line 34: If this is supposed to be a docstring, use triple quotes and command sentence, otherwise I'd say it's a superfluous comment

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+        elif Key.__is_ecdsa_key(data):
+            return ECDSAKey
+        elif Key.__is_rsa_key(data):
+            return RSAKey
+        return None
+
+    @staticmethod
+    def __is_ecdsa_key(data: str):
+        try:
+            ecdsa.VerifyingKey.from_pem(data)
```
Starkteetje, line 48: Don't know whether there's a better way, but this leads to VerifyingKey.from_pem(data) being called twice. Once to get key type and then to get key. Maybe there's a reasonable way to pass the key object up, without breaking the abstraction, maybe not :man_shrugging:

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+            return True
+        except (ecdsa.der.UnexpectedDER, binascii.Error, TypeError, AttributeError):
```
Starkteetje, line 50: Why these errors in particular? Given https://github.com/tlsfuzzer/python-ecdsa/blob/065b40c43e066de6df7f263ee43ade80d86d79f2/src/ecdsa/keys.py#L453 there seems to be at least the possibility for an AssertionError(?).
Wouldn't we want to catch any exception and return False?

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+            return False
+
+    @staticmethod
+    def __is_rsa_key(data: str):
+        try:
+            rsa.PublicKey.load_pkcs1_openssl_pem(data)
+            return True
+        except (ValueError, PyAsn1Error):
```
Starkteetje, line 58: Same here, why not any exception?

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+    def __str__(self) -> str:
+        return self.value
+
+
+class ECDSAKey(Key):
+    def __init__(self, data: str) -> None:
+        self.value = ecdsa.VerifyingKey.from_pem(data)
+
+    def verify(self, validator_type: str, **kwargs):
+        if validator_type == "notaryv1":
```
Starkteetje, line 76: Should this and "cosign" strings rather be constants?

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+KMS_REGEX = r"(?:awskms|gcpkms|azurekms|hashivault|k8s):\/{2,3}[a-zA-Z0-9_.+\/:-]+"
+KEYLESS_REGEX = r"^[a-zA-Z0-9_.+-]+@[a-zA-Z0-9-]+\.[a-zA-Z0-9-.]+$"
+
+
+class Key:
+    """
+    Abstract Key class, containing the public key used for verifying the image signatures.
+    """
+
+    value: str
```
Starkteetje, line 22: Not a string. For e.g. ECDSAKey you use value.verify from VerifyingKey object

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+        self.value = rsa.PublicKey.load_pkcs1_openssl_pem(data)
+
+    def verify(self, validator_type: str, **kwargs):
+        if validator_type == "notaryv1":
+            pass
+        elif validator_type == "cosign":
+            pass
+        return super().verify(validator_type, **kwargs)
+
+    def __verify_cosign(
```
Starkteetje, line 107: Why does this exist if its unused?

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+class KeyLessKey(Key):
+    def __init__(self, data: str) -> None:
+        self.value = data
+
+    def verify(self, validator_type: str, **kwargs):
+        if validator_type == "cosign":
+            pass
+        return super().verify(validator_type, **kwargs)
+
+    def __verify_cosign(
```
Starkteetje, line 138: As above, why this implemented, but not called?

connaisseur/keys.py
Starkteetje: Seeing Keyless, it doesn't even represent a Key, maybe the name is not correct in the first place. Maybe TrustAnchor?

Starkteetje (changes requested):

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -117,41 +119,33 @@ def __get_cosign_validated_digests(self, image: str, key: str):
             raise UnexpectedCosignData(message=msg)
         return digests
 
-    def __invoke_cosign(self, image, key):
+    def __cosign_callback(self, image: Image, key_args: list):
         """
-        Invoke the Cosign binary in a subprocess for a specific `image` given a `key` and
-        return the returncode, stdout and stderr. Will raise an exception if Cosign times out.
+        Callback function that is passed on to any `key.verify()` that is using a cosign
```
Starkteetje, line 124: Docstring comments should be commands

Should first say what it does and if necessary, then what the context is

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -161,29 +155,19 @@ def __invoke_cosign(self, image, key):
...
-    def __get_pubkey_config(key: str):
+    def __get_envs(self):
         """
-        Return a tuple of the used Cosign verification command (flag-value list), a
-        dict of potentially required environment variables and public key in binary
-        PEM format to be used as stdin to Cosign based on the format of the input
-        key (reference).
-
-        Raise InvalidFormatException if none of the supported patterns is matched.
+        Sets up environment variables used by cosign for potential authentication or TLS
```
Starkteetje, line 160: Docstring as command

is the use `for potential authentication or TLS verification against private registries.` relevant to the comment?

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -161,29 +155,19 @@ def __invoke_cosign(self, image, key):
...
-
-        # key is KMS reference
-        if re.match(r"^\w{2,20}://[\w:/-]{3,255}$", key):
-            return ["--key", key], {}, b""
-
-        msg = "Public key (reference) '{input_str}' does not match expected patterns."
-        raise InvalidFormatException(message=msg, input_str=key)
+        env = os.environ.copy()
+        # Extend the OS env vars only for passing to the subprocess below
+        env["DOCKER_CONFIG"] = f"/app/connaisseur-config/{self.name}/.docker/"
```
Starkteetje, line 165: Should this be set even if "DOCKER_CONFIG" already exists?

connaisseur/keys.py
```diff
@@ -0,0 +1,141 @@
...
+
+    @staticmethod
+    def __is_rsa_key(data: str):
+        try:
+            rsa.PublicKey.load_pkcs1_openssl_pem(data)
+            return True
+        except (ValueError, PyAsn1Error):
+            return False
+
+    def verify(self, validator_type: str, **kwargs):  # pylint: disable=unused-argument
```
Starkteetje, line 61: Something else that occurred to me. Should the key be the object that verifies?

The key itself doesnt really have any connection to what validator is used and hence to how verification is to be done. On the other hand, the validator does know what key to use and how to use said key. While the key doing it doesn't feel bad (I noticed it really late through the review), thinking about it does not feel good either and validator seems like a way better place. Thoughts?

phbelitz: I think logically speaking your right. In the current implementation I didn't need to do a differentiation between the keys within the validators, since the keys do the verification, but I see how this is both confusing and logically wrong.

I'll change up the logic so that a validator looks for the key type it's getting and then switches up verification accordingly. Fine with this flow?

connaisseur/validators/cosign/cosign_validator.py
phbelitz: yes, cause different validators could have different configs.

connaisseur/keys.py
```diff
@@ -0,0 +1,93 @@
...
+class Key(KeyInterface):
+    """
+    Abstract Key class, containing the public key used for verifying the image signatures.
+    """
+
+    value: object
+
+    def __new__(cls, data: str):
+        try:
+            key_cls, key_data = Key.__get_key_type_cls(data)
```
Starkteetje, line 39: This function seems to do more than get the type

connaisseur/keys.py
```diff
@@ -0,0 +1,93 @@
...
+            return key_cls.__new__(key_cls, key_data)
+        except Exception as err:
+            msg = "Cannot discern key type."
+            raise NoSuchClassError(message=msg, key_type=key_cls) from err
```
Starkteetje, line 43: Would key_cls not always be expected to be None? If the misnamed function returns class and data, instantiation should always work, shouldn't it? And in that case, discerning the type wouldn't be the problem, but rather instantiation of the key object or am I missing sth?

connaisseur/keys.py
```diff
@@ -0,0 +1,93 @@
...
+        elif re.match(KMS_REGEX, data):
+            return KMSKey, data
+        elif key := Key.__is_ecdsa_key(data):
+            return ECDSAKey, key
+        elif key := Key.__is_rsa_key(data):
+            return RSAKey, key
+        return None, None
+
+    @staticmethod
+    def __is_ecdsa_key(data: str):
```
Starkteetje, line 58: This is a misnomer now, since it also returns the key

connaisseur/keys.py
```diff
@@ -0,0 +1,93 @@
...
+        try:
+            return ecdsa.VerifyingKey.from_pem(data)
+        except Exception:
+            return None
+
+    @staticmethod
+    def __is_rsa_key(data: str):
```
Starkteetje, line 65: This is a misnomer now, since it also returns the key

connaisseur/keys.py
```diff
@@ -0,0 +1,93 @@
...
+            return None
+
+
+class ECDSAKey(KeyInterface):
+    def __str__(self) -> str:
+        return base64.b64encode(self.value.to_der()).decode("utf-8")
+
+
+class RSAKey(KeyInterface):
+    # reminder when implementing RSA validation:
```
Starkteetje, line 78: This seems to be cosign specific, while RSA shouldn't be. See #99 . Could be part of notary as well, so this comment (if useful?) is misplaced

connaisseur/keys.py
```diff
@@ -0,0 +1,93 @@
...
+    def __str__(self) -> str:
+        return base64.b64encode(self.value.save_pkcs1("DER")).decode("utf-8")
+
+
+class KMSKey(KeyInterface):
+    pass
+
+
+class KeyLessKey(KeyInterface):
+    # reminder when implementing Keyless validation:
```
Starkteetje, line 90: While this is more or less guaranteed to be cosign specific, the key still shouldn't hold this info I'd argue

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -217,35 +216,50 @@ def __get_cosign_validated_digests(self, image: str, trust_root: dict):
...
+            return self.__invoke_cosign(image, ["--key", key.value, b""])
+        msg = (
+            "The key type {key_type} is unsupported for a validator of type {val_type}."
+        )
+        raise WrongKeyError(message=msg, key_type=type(key), val_type="cosign")
 
+    def __invoke_cosign(self, image: str, key_args: list):
+        """
+        Invoke the Cosign binary in a subprocess for a specific `image` given a `key` and
+        return the returncode, stdout and stderr. Will raise an exception if Cosign times
```
Starkteetje, line 239: nit: `Will raise` is descriptive instead of commanding again

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -255,33 +269,9 @@ def __invoke_cosign(self, image: str, key: str):
...
-        msg = "Public key (reference) '{input_str}' does not match expected patterns."
-        raise InvalidFormatException(message=msg, input_str=key)
-
     def __get_envs(self):
         """
-        Sets up environment variables used by cosign.
+        Set up environment variables used by cosign.
         """
         env = os.environ.copy()
         # Extend the OS env vars only for passing to the subprocess below
```
Starkteetje, line 277: Since I see it from here but Github doesn't allow me to comment there, ` __apply_policy` docstring descriptive

connaisseur/validators/notaryv1/trust_data.py
```diff
@@ -93,11 +94,28 @@ def validate_signature(self, keystore: KeyStore):
             sig = signature["sig"]
 
             try:
-                verify_signature(pub_key, sig, msg)
+                TrustData.__validate_signature_with_key(sig, msg, pub_key)
```
Starkteetje, line 97: Since we've in cosign gone away from public key, do you think it makes sense to go to key from pub_key here? I guess its mostly still PKC, but I think I've read stuff about x509 in #99 . Without having looked into it, pub_key may then be a certificate. While still mainly having the pubkey as interesting property, this would then be somewhat inaccurate

Since I don't see us implementing x509 soon though, we can also not do that now and stick with the name that better fits the current use case.

No strong stakes, just noticed. Thy choice

connaisseur/validators/notaryv1/trust_data.py
```diff
@@ -93,11 +94,28 @@ def validate_signature(self, keystore: KeyStore):
...
             except Exception as err:
                 msg = "Failed to verify signature of trust data {trust_data_kind}."
                 raise ValidationError(message=msg, trust_data_kind=self.kind) from err
 
+    @staticmethod
+    def __validate_signature_with_key(signature: str, payload: str, key: Key):
+        if isinstance(key, ECDSAKey):
+            return TrustData.__validate_signature_with_ecdsa(signature, payload, key)
+        msg = (
```
Starkteetje, line 106: Given that this function can already render the error string, is there a reason to delegate rendering the error string to the error class?

Starkteetje (changes requested):

connaisseur/keys.py
phbelitz: The instantiation will still fail, since `None.__new__(None, data)` will raise a `TypeError`.

connaisseur/validators/notaryv1/trust_data.py
phbelitz: i don't understand ...?

Starkteetje: "The key type {key_type} is unsupported for a validator of type {val_type}." is not rendered as the error string, but rather is the template. The function, where I commented does provide the template and the data inserted into the template. Why let the error class handle rendering the error message instead of simply providing the error that this function apparently knows it wants to output?

connaisseur/trust_root.py
```diff
@@ -0,0 +1,89 @@
...
+    """
+
+    value: object
+
+    def __new__(cls, data: str):
+        try:
+            tr_cls, tr_data = TrustRoot.__get_tr_type_cls_and_data(data)
+            return tr_cls.__new__(tr_cls, tr_data)
+        except Exception as err:
+            msg = "Cannot discern trust anchor type."
```
Starkteetje, line 44: Root not anchor?

connaisseur/keys.py
Starkteetje: Putting the key data there is dangerous as we may have symmetric secrets as root at some point and then this logs a secret. If you want to notify about which key is "bad", log the name from the helm.yml comfig

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -217,35 +221,58 @@ def __get_cosign_validated_digests(self, image: str, trust_root: dict):
...
+        msg = (
+            "The trust_root type {tr_type} is unsupported for a validator of type"
+            "{val_type}."
+        )
+        raise WrongKeyError(message=msg, tr_type=type(trust_root), val_type="cosign")
+
+    def __invoke_cosign(self, image: str, tr_args: list):
+        """
+        Invoke the Cosign binary in a subprocess for a specific `image` given a
+        `trust_root` and return the returncode, stdout and stderr.
```
Starkteetje, line 251: This comment seems somewhat outdated

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -217,35 +221,58 @@ def __get_cosign_validated_digests(self, image: str, trust_root: dict):
...
+
+        Raises an exception if Cosign times out.
+        """
+        option_kword, inline_tr, trust_root = tr_args
```
Starkteetje, line 255: This seems to assume a very specifc config of the tr_args, which is not apparent from the interface. Is __function, so maybe alright, but maybe could be improved

Starkteetje (changes requested):

connaisseur/validators/notaryv1/trust_data.py
phbelitz: created issue #646

connaisseur/validators/cosign/cosign_validator.py
phbelitz: tried to make it more clear in the functions description

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -217,35 +221,61 @@ def __get_cosign_validated_digests(self, image: str, trust_root: dict):
...
+            "The trust_root type {tr_type} is unsupported for a validator of type"
+            "{val_type}."
+        )
+        raise WrongKeyError(message=msg, tr_type=type(trust_root), val_type="cosign")
+
+    def __invoke_cosign(self, image: str, tr_args: list):
+        """
+        Invoke the Cosign binary in a subprocess for a specific `image` given trust root
+        argument list (`tr_args`) and return the returncode, stdout and stderr. The trust
+        root argument list includes a cosign option keyword and the trust root itself,
```
Starkteetje, line 252:
```suggestion
        root argument list includes a Cosign option keyword and the trust root itself,
```

connaisseur/trust_root.py
```diff
@@ -0,0 +1,90 @@
...
+        self.value = data
+
+    def __str__(self) -> str:
+        return self.value
+
+
+class TrustRoot(TrustRootInterface):
+    """
+    Abstract TrustRoot class used to represent key material or similar entities, used in
+    signature verification processes.
```
Starkteetje, line 33:
```suggestion
    verification processes.
```
?

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -217,35 +221,72 @@ def __get_cosign_validated_digests(self, image: str, trust_root: dict):
...
+            "{val_type}."
+        )
+        raise WrongKeyError(message=msg, tr_type=type(trust_root), val_type="cosign")
+
+    def __invoke_cosign(self, image: str, tr_args: dict):
         """
-        pubkey_config, env_vars, pubkey = CosignValidator.__get_pubkey_config(key)
+        Invoke the Cosign binary in a subprocess for a specific `image` given trust root
+        argument list (`tr_args`) and return the returncode, stdout and stderr. The trust
+        root argument list includes a Cosign option keyword and the trust root itself,
```
Starkteetje, line 264:
```suggestion
        root argument dict includes a Cosign option keyword and the trust root itself,
```

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -217,35 +221,72 @@ def __get_cosign_validated_digests(self, image: str, trust_root: dict):
...
+        either as inline argument or pipeable input with an inline reference. The
+        composition of the list is dependant on the type of trust root.
```
Starkteetje, line 266:
```suggestion
        composition of the dict is dependant on the type of trust root.
```

Starkteetje (approved):

PR #497, opened 2022-01-22
ci: fix release pipeline

phbelitz:
Fixes the release pipeline.

## Description

The `ImagePullPolicy` for the release pipeline was accedentially set to `Never`, thus no image could be loaded. This was fixed.

## Checklist

- [x] PR is rebased to/aimed at branch `develop`
- [x] PR follows [Contributing Guide](../docs/CONTRIBUTING.md)
- [x] Added tests (if necessary)
- [x] Extended README/Documentation (if necessary)
- [ ] Adjusted versions of image and Helm chart in `values.yaml` and `Chart.yaml` (if necessary)

PR #542, opened 2022-02-11
test: fix unset variable in preconfig integration test

annekebr:
As the actual update of the config values happens in its own bash function, the `IPP` variable is no longer set when performing the values.yaml update resulting in an empty value in the yaml file (`imagePullPolicy: ""`).

Fixes #

## Checklist

### DIFF
--- a/tests/integration/integration-test.sh
+++ b/tests/integration/integration-test.sh
@@ -282,7 +282,7 @@ case $1 in
   deployment_int_test
   ;;
 "pre-config")
-  IPP=`yq e '.deployment.imagePullPolicy' tests/integration/update.yaml`
+  export IPP=`yq e '.deployment.imagePullPolicy' tests/integration/update.yaml`
   update_values '.deployment.imagePullPolicy=strenv(IPP)'
   helm_install
   pre_config_int_test


xopham (commented):
happy to approve this. however, does it actually cause an error?

annekebr:
@xopham I'm pretty sure that it was intended to actually replace the value for the `imagePullPolicy` in https://github.com/sse-secure-systems/connaisseur/blob/master/tests/integration/integration-test.sh#L286 by the value `IPP` that seems to be set only for this purpose, but as the actual update is invoked in its [own function|https://github.com/sse-secure-systems/connaisseur/blob/master/tests/integration/integration-test.sh#L173] the variable is no longer set. I was wondering why the test still works as Kubernetes says that it only allows `Never`, `IfNotPresent` and `Always` as imagePullPolicy according to their docs, but in my local tests an empty string seems to be valid as well. Didn't allow that in https://github.com/sse-secure-systems/connaisseur/blob/master/tests/integration/integration-test.sh#L173 though, that's why I stumbled across it. I updated https://github.com/sse-secure-systems/connaisseur/blob/master/tests/integration/integration-test.sh#L173 accordingly, but I thought that this might be worth fixing anyway before it actually causes issues.

annekebr:
@xopham I'm pretty sure that it was intended to actually replace the value for the `imagePullPolicy` in https://github.com/sse-secure-systems/connaisseur/blob/master/tests/integration/integration-test.sh#L286 by the value `IPP` that seems to be set only for this purpose, but as the actual update is invoked in its [own function|https://github.com/sse-secure-systems/connaisseur/blob/master/tests/integration/integration-test.sh#L173] the variable is no longer set. I was wondering why the test still works as Kubernetes says that it only allows `Never`, `IfNotPresent` and `Always` as imagePullPolicy according to their docs, but in my local tests an empty string seems to be valid as well. Didn't allow that in https://github.com/sse-secure-systems/connaisseur/blob/master/tests/integration/integration-test.sh#L173 though, that's why I stumbled across it. I updated https://github.com/sse-secure-systems/connaisseur/blob/master/tests/integration/integration-test.sh#L173 accordingly, but I thought that this might be worth fixing anyway before it actually causes issues.

xopham:
Thx for updating 🙏
Might that have caused #497 ?

xopham (approved):
Lgtm

annekebr:
It seems that the issue was introduced in https://github.com/sse-secure-systems/connaisseur/pull/497, but as `imagePullPolicy=""` defaults to `Always` it still worked as intended :)

PR #545, opened 2022-02-15
test: added tests for logging wrapper

phbelitz:
Adds missing tests for the logging wrapper

## Checklist

### DIFF
--- a/tests/test_logging_wrapper.py
+++ b/tests/test_logging_wrapper.py
@@ -1,0 +1,61 @@
+import pytest
+import time
+
+from . import conftest as fix
+
+import connaisseur.logging_wrapper as lw
+
+
+@pytest.fixture
+def mock_time(monkeypatch):
+    monkeypatch.setattr(time, "strftime", lambda X, y: "1/1/1970 00:00:00")
+
+
+@pytest.mark.parametrize(
+    "status_code, environ, out",
+    [
+        (
+            200,
+            {
+                "REMOTE_ADDR": "127.0.0.1",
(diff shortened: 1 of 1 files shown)


connaisseur/logging_wrapper.py
```diff
@@ -11,7 +11,10 @@ def __init__(self, app, log_level):
         self.logger = logging.getLogger(
             "connaisseurRequestLogger"
         )  # has no handler and hence propagates logs to root logger
-        self.logger.setLevel(log_level)
+        try:
+            self.logger.setLevel(log_level)
+        except ValueError:
```
Starkteetje, line 16: Would you want to log this misconfig?

Starkteetje (commented):
See comment, otherwise lgtm

connaisseur/logging_wrapper.py
phbelitz: hm ... I see now that this is dead code ... a wrong log level already raises an exception in connaisseurs `__main__.py` ... so i guess i revert this back

tests/test_logging_wrapper.py
```diff
@@ -0,0 +1,61 @@
+from tracemalloc import start
```
Starkteetje, line 1: I don't see start used, is it a superfluous import?

Starkteetje: And also, import order is meh

Afaik PEP says python modules emptyline external modules emptyline modules from current project

Starkteetje (approved):

Starkteetje (approved):

PR #560, opened 2022-02-28
Alert - Support for authentication when calling receiver endpoint.

andreee94:
The changes implement more authentication options without requiring to hard-code the secret inside the url.

Until now is not possible to call an alert webhook which requires a basic or a bearer authentication.

The secrets should not be passed directly inside the configuration, instead they can be injected as environmental variables or files. 

## Description

Three authentication options have been implemented:
- **basic** from environmental variables
- **bearer** token from environmental variables
- **bearer** token from file (for example the service account token file in kubernetes)

Is however possible to specify a custom header prefix other than *Basic* and *Bearer*.

The **validation schema** has been updated to support the new options and the **unit test** has been written. All tests pass (running them inside the alpine container as suggested by the documentation).

The **documentation** reports a description of the new functionalities and few examples.

The new connaisseur image has been installed manually (forked from master) inside a K3S cluster and it worked correctly .

## Checklist

- [x] Updated schema validation (if necessary) 

### DIFF
--- a/connaisseur/alert.py
+++ b/connaisseur/alert.py
@@ -48,6 +48,166 @@ class AlertingConfiguration:
         return bool(self.config.get(event_category))
 
 
+class AlertReceiverAuthentication:
+    """
+    Class to store authentication information for securely sending events to the alert receiver.
+    """
+
+    class AlertReceiverBasicAuthentication:
+        """
+        Class to store authentication information for basic authentication type with username and password.
+        """
+        username: str
+        password: str
+        authentication_type: str
+        authorization_prefix: str = "Basic"
+
+        def __init__(self, alert_receiver_config: dict):
+            basic_authentication_config = alert_receiver_config.get(
+                "receiver_authentication_basic", None
+            )
+
+            if (
+                basic_authentication_config is None
+            ):  # TODO maybe remove this check since it is included in the json validation?
+                raise ConfigurationError("No basic authentication configuration found.")
+
+            username_env = basic_authentication_config.get("username_env", None)
+            password_env = basic_authentication_config.get("password_env", None)
+
+            if (
+                username_env is None or password_env is None
+            ):  # TODO maybe remove this check since it is included in the json validation?
+                raise ConfigurationError(
+                    "No username_env or password_env configuration found."
+                )
+
+            self.username = os.environ.get(username_env, None)
+            self.password = os.environ.get(password_env, None)
+
+            if self.username is None or self.password is None:
+                raise ConfigurationError(
+                    f"No username or password found from environmental variables {username_env} and {password_env}."
+                )
+
+            self.authorization_prefix = basic_authentication_config.get(
+                "authorization_prefix", "Basic"
+            )
+            # TODO maybe validate authorization prefix
+
+        def get_header(self) -> dict:
+            return {
+                "Authorization": f"{self.authorization_prefix} {self.username}:{self.password}"
+            }
+
+    class AlertReceiverBearerAuthentication:
+        """
+        Class to store authentication information for bearer authentication type which uses a token.
+        """
+        token: str
+        authorization_prefix: str = "Bearer"  # default is bearer
+
+        def __init__(self, alert_receiver_config: dict):
+            bearer_authentication_config = alert_receiver_config.get(
+                "receiver_authentication_bearer", None
+            )
+
+            if (
+                bearer_authentication_config is None
+            ):  # TODO maybe remove this check since it is included in the json validation?
+                raise ConfigurationError(
+                    "No bearer authentication configuration found."
+                )
+
+            token_env = bearer_authentication_config.get("token_env", None)
+            token_file = bearer_authentication_config.get("token_file", None)
+
+            if (
+                token_env is None and token_file is None
+            ):  # TODO maybe remove this check since it is included in the json validation?
+                raise ConfigurationError(
+                    "No token_env and token_file configuration found."
+                )
+
+            if (
+                token_env is not None and token_file is not None
+            ):  # TODO maybe remove this check since it is included in the json validation?
+                raise ConfigurationError(
+                    "Both token_env and token_file configuration found. Only one is required."
+                )
+
+            if token_env is not None:
+                self.token = os.environ.get(token_env, None)
+
+                if self.token is None:
+                    raise ConfigurationError(
+                        f"No token found from environmental variable {token_env}."
+                    )
+            else:
+                try:
+                    with open(token_file, "r") as token_file:
+                        self.token = token_file.read()
+                except FileNotFoundError:
+                    raise ConfigurationError(f"No token file found at {token_file}.")
+                except Exception as err:
+                    raise ConfigurationError(
+                        f"An error occurred while loading the token file {token_file}: {str(err)}"
+                    )
+
+            self.authorization_prefix = bearer_authentication_config.get(
+                "authorization_prefix", "Bearer"
+            )
+            # TODO maybe validate authorization prefix
+
+        def get_header(self) -> dict:
+            return {"Authorization": f"{self.authorization_prefix} {self.token}"}
+
+    def __init__(self, alert_receiver_config: dict):
+        self.authentication_type = alert_receiver_config.get(
+            "receiver_authentication_type", "none"
+        )
+
+        if self.is_basic():
+            self.__init_basic_authentication(alert_receiver_config)
+        elif self.is_bearer():
+            self.__init_bearer_authentication(alert_receiver_config)
+
+    def is_basic(self):
+        return self.authentication_type == "basic"
+
+    def is_bearer(self):
+        return self.authentication_type == "bearer"
+
+    def is_none(self):
+        return self.authentication_type == "none"
+
+    def __init_bearer_authentication(self, alert_receiver_config: dict):
+        self.bearer_authentication = (
+            AlertReceiverAuthentication.AlertReceiverBearerAuthentication(
+                alert_receiver_config
+            )
+        )
+
+    def __init_basic_authentication(self, alert_receiver_config: dict):
+        self.basic_authentication = (
+            AlertReceiverAuthentication.AlertReceiverBasicAuthentication(
+                alert_receiver_config
+            )
+        )
+
+    def get_auth_header(self) -> dict:
+        if self.is_basic():
+            return self.basic_authentication.get_header()
+        elif self.is_bearer():
+            return self.bearer_authentication.get_header()
+        elif self.is_none():
+            return {}
+        else:
+            raise ConfigurationError(
+                "No authentication type found."
+            )  # hopefully this never happens
+
+
 class Alert:
     """
     Class to store image information about an alert as attributes and a sending
@@ -59,6 +219,7 @@ class Alert:
 
     template: str
     receiver_url: str
+    receiver_authentication: AlertReceiverAuthentication
     payload: str
     headers: dict
 
@@ -101,12 +262,13 @@ class Alert:
             "images": images,
         }
         self.receiver_url = receiver_config["receiver_url"]
+        self.receiver_authentication = AlertReceiverAuthentication(receiver_config)
         self.template = receiver_config["template"]
         self.throw_if_alert_sending_fails = receiver_config.get(
             "fail_if_alert_sending_fails", False
         )
         self.payload = self.__construct_payload(receiver_config)
-        self.headers = self.__get_headers(receiver_config)
+        self.headers = self.__get_headers(receiver_config, self.receiver_authentication)
 
     def __construct_payload(self, receiver_config: dict) -> str:
         try:
@@ -153,13 +315,18 @@ class Alert:
         return response
 
     @staticmethod
-    def __get_headers(receiver_config):
+    def __get_headers(
+        receiver_config: dict, receiver_authentication: AlertReceiverAuthentication
+    ) -> dict:
         headers = {"Content-Type": "application/json"}
         additional_headers = receiver_config.get("custom_headers")
         if additional_headers is not None:
             for header in additional_headers:
                 key, value = header.split(":", 1)
                 headers.update({key.strip(): value.strip()})
+        auth_header = receiver_authentication.get_auth_header()
+        if auth_header:  # not None and not empty
+            headers.update(auth_header)
         return headers
 
 
--- a/connaisseur/alert.py
+++ b/connaisseur/alert.py
@@ -59,15 +59,15 @@ class AlertReceiverAuthentication:
     class AlertReceiverAuthenticationInterface:
         def __init__(self, alert_receiver_config: dict, authentication_config_key: str):
             if authentication_config_key is not None:
-                self.authentication_config = alert_receiver_config.get(
-                    authentication_config_key
-                )
-
-                if self.authentication_config is None:
+                try:
+                    self.authentication_config = alert_receiver_config[
+                        authentication_config_key
+                    ]
+                except KeyError as err:
                     raise ConfigurationError(
                         "No authentication configuration found for dictionary key:"
                         f"{authentication_config_key}."
-                    )
+                    ) from err
 
                 self.authentication_scheme = self.authentication_config.get(
                     "authentication_scheme", self.authentication_scheme
@@ -109,24 +109,22 @@ class AlertReceiverAuthentication:
         def __init__(self, alert_receiver_config: dict):
             super().__init__(alert_receiver_config, "receiver_authentication_basic")
 
-            username_env = self.authentication_config.get("username_env")
-            password_env = self.authentication_config.get("password_env")
-
-            if (
-                username_env is None or password_env is None
-            ):  # This should not happen since it is included in the json validation
+            try:
+                username_env = self.authentication_config["username_env"]
+                password_env = self.authentication_config["password_env"]
+            except KeyError as err:
                 raise ConfigurationError(
                     "No username_env or password_env configuration found."
-                )
+                ) from err
 
-            self.username = os.environ.get(username_env, None)
-            self.password = os.environ.get(password_env, None)
-
-            if self.username is None or self.password is None:
+            try:
+                self.username = os.environ[username_env]
+                self.password = os.environ[password_env]
+            except KeyError as err:
                 raise ConfigurationError(
                     "No username or password found from environmental variables "
                     f"{username_env} and {password_env}."
-                )
+                ) from err
 
         def get_header(self) -> dict:
             return {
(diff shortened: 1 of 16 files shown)


connaisseur/alert.py
```diff
@@ -48,6 +48,166 @@ def alerting_required(self, event_category: str) -> bool:
...
+        Class to store authentication information for basic authentication type with username and password.
+        """
+        username: str
+        password: str
+        authentication_type: str
+        authorization_prefix: str = "Basic"
+
+        def __init__(self, alert_receiver_config: dict):
+            basic_authentication_config = alert_receiver_config.get(
+                "receiver_authentication_basic", None
```
peterthomassen, line 67: `None` is the default for `dict.get`. Same for other occurrences.

connaisseur/alert.py
```diff
@@ -48,6 +48,166 @@ def alerting_required(self, event_category: str) -> bool:
...
+            )
+
+            if (
+                basic_authentication_config is None
+            ):  # TODO maybe remove this check since it is included in the json validation?
```
peterthomassen, line 72: the check is fine

connaisseur/alert.py
```diff
@@ -48,6 +48,166 @@ def alerting_required(self, event_category: str) -> bool:
...
+            self.username = os.environ.get(username_env, None)
+            self.password = os.environ.get(password_env, None)
+
+            if self.username is None or self.password is None:
+                raise ConfigurationError(
+                    f"No username or password found from environmental variables {username_env} and {password_env}."
+                )
+
+            self.authorization_prefix = basic_authentication_config.get(
+                "authorization_prefix", "Basic"
```
peterthomassen, line 94: Should the default be `self.authorization_prefix`, as the value is hardcoded at the class?

connaisseur/alert.py
```diff
@@ -48,6 +48,166 @@ def alerting_required(self, event_category: str) -> bool:
...
+            )
+            # TODO maybe validate authorization prefix
```
peterthomassen, line 96: Sounds like a good idea

connaisseur/alert.py
```diff
@@ -48,6 +48,166 @@ def alerting_required(self, event_category: str) -> bool:
...
+            ):  # TODO maybe remove this check since it is included in the json validation?
+                raise ConfigurationError(
+                    "No token_env and token_file configuration found."
+                )
+
+            if (
+                token_env is not None and token_file is not None
+            ):  # TODO maybe remove this check since it is included in the json validation?
+                raise ConfigurationError(
+                    "Both token_env and token_file configuration found. Only one is required."
```
peterthomassen, line 136: `Only one can be given.`

connaisseur/alert.py
```diff
@@ -48,6 +48,166 @@ def alerting_required(self, event_category: str) -> bool:
...
+            bearer_authentication_config = alert_receiver_config.get(
+                "receiver_authentication_bearer", None
+            )
+
+            if (
+                bearer_authentication_config is None
+            ):  # TODO maybe remove this check since it is included in the json validation?
+                raise ConfigurationError(
+                    "No bearer authentication configuration found."
+                )
```
peterthomassen, line 120: This is boilerplate (together with assigning `self.authentication_prefix` at the end of the method).

Suggestion: make a base class, introduce a variable for the lookup key (e.g. `"receiver_authentication_bearer"`), and run all this via `super().__init__(*args, **kwargs)`.

connaisseur/alert.py
```diff
@@ -48,6 +48,166 @@ def alerting_required(self, event_category: str) -> bool:
...
+
+    def __init__(self, alert_receiver_config: dict):
+        self.authentication_type = alert_receiver_config.get(
+            "receiver_authentication_type", "none"
+        )
+
+        if self.is_basic():
+            self.__init_basic_authentication(alert_receiver_config)
+        elif self.is_bearer():
+            self.__init_bearer_authentication(alert_receiver_config)
```
peterthomassen, line 173:
```python
init_map = {
    "basic": AlertReceiverAuthentication.AlertReceiverBasicAuthentication,
    "bearer": AlertReceiverAuthentication.AlertReceiverBearerAuthentication,
}
try:
    self.auth_instance = init_map[self.authentication_type](alert_receiver_config)
except AttributeError:
    ....
```

Then, it `get_auth_header()`, you don't need to do the case handling, but can just return `{} if self.auth_instance is None self.auth_instance.get_header()`.
```

File: connaisseur/res/alertconfig_schema.json [hunk 8]
```diff
@@ -89,4 +235,4 @@
       ]
     }
   }
-}
+}
```
peterthomassen, line 238: missing newline

peterthomassen (commented):
Thanks a lot for this awesome PR! This is really useful, and also good quality.

I added a few suggestions on how to improve the code, especially structure it so that there is less repetition.

@xopham I did not review schema definitions, tests, and docs.

connaisseur/alert.py
andreee94: I cannot find an actual validation rule.
The only thing I can found is: [https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization).
I have assumed that the authentication scheme (previously called authorization prefix) is not null or empty and does not contain any space.

andreee94:
@peterthomassen I have updated the alert classes accordingly to your suggestions. 
Let me know if I can make something better.

connaisseur/alert.py
```diff
@@ -53,31 +54,66 @@ class AlertReceiverAuthentication:
     Class to store authentication information for securely sending events to the alert receiver.
     """
 
-    class AlertReceiverBasicAuthentication:
+    authentication_config: dict = None
+    authentication_scheme: str = None
+
+    class AlertReceiverAuthenticationInterface:
+        def __init__(self, alert_receiver_config: dict, authentication_key: str):
+            self.authentication_config = alert_receiver_config.get(authentication_key)
```
peterthomassen, line 62: `authentication_key` sounds like a credential

connaisseur/alert.py
```diff
@@ -53,31 +54,66 @@ class AlertReceiverAuthentication:
...
+            )
+            self._validate_authentication_scheme()
+
+        def _validate_authentication_scheme(self) -> None:
+            if not self.authentication_scheme:
+                raise ConfigurationError(
+                    "The authentication scheme cannot be null or empty."
+                )
+
+            if " " in self.authentication_scheme:
```
peterthomassen, line 80: You might want to check that it contains only a-zA-Z or something like that. (`" "` is very specific, doesn't catch tabs, newlines, ...)

connaisseur/alert.py
```diff
@@ -53,31 +54,66 @@ class AlertReceiverAuthentication:
...
+
+    class AlertReceiverNoneAuthentication(AlertReceiverAuthenticationInterface):
+        """
+        Placeholder class for AlertReceiver without authentication.
+        """
+        def __init__(self, alert_receiver_config: dict):
+            pass
+
+        def get_header(self) -> dict:
+            return {}
```
peterthomassen, line 98: This could be the default implementation (instead of `@abstractmethod`). (Just a thought, certainly not mandatory.)

connaisseur/alert.py
```diff
@@ -154,58 +177,37 @@ def __init__(self, alert_receiver_config: dict):
...
-            )
-        )
-
-    def __init_basic_authentication(self, alert_receiver_config: dict):
-        self.basic_authentication = (
-            AlertReceiverAuthentication.AlertReceiverBasicAuthentication(
-                alert_receiver_config
-            )
-        )
+        return self.init_map.get(self.authentication_type)
```
peterthomassen, line 208:
```python
try:
    return self.init_map[self.authentication_type]
except KeyError:
    raise ConfigurationError(...)
```

This is a Python paradigm called EAFP. For context, see https://sauravmaheshkar.github.io/blog/Introduction-To-Duck-Typing-and-EAFP/.

One benefit is that EAFP gets ride of many [TOCTOU](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use) race conditions (although not the case here, as the state is purely local).

xopham (commented):
@andreee94 real neat contribution :rocket: like the work you are contributing here :pray: 

there is still a few pylint findings:
```
connaisseur/alert.py:102:0: C0301: Line too long (107/105) (line-too-long)
connaisseur/alert.py:127:0: C0301: Line too long (116/105) (line-too-long)
connaisseur/alert.py:205:0: C0301: Line too long (117/105) (line-too-long)
connaisseur/alert.py:94:8: W0231: __init__ method from base class 'AlertReceiverAuthenticationInterface' is not called (super-init-not-called)
connaisseur/alert.py:172:25: W1514: Using open without explicitly specifying an encoding (unspecified-encoding)
connaisseur/alert.py:175:20: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
connaisseur/alert.py:177:20: W0707: Consider explicitly re-raising using the 'from' keyword (raise-missing-from)
connaisseur/alert.py:203:43: C0201: Consider iterating the dictionary directly instead of calling .keys() (consider-iterating-dictionary)
```
will take a closer look at the pr once @peterthomassen is done.

connaisseur/alert.py
andreee94: I was not aware of this python principle. Thank you.

connaisseur/alert.py
andreee94: Yes, I agree. It will be called `authentication_config_key`.

connaisseur/alert.py
```diff
@@ -58,48 +57,49 @@ class AlertReceiverAuthentication:
...
 
-            self.authentication_scheme = self.authentication_config.get(
-                "authentication_scheme", self.authentication_scheme
-            )
-            self._validate_authentication_scheme()
+                if self.authentication_config is None:
+                    raise ConfigurationError(
+                        "No authentication configuration found for dictionary key:"
+                        f"{authentication_config_key}."
+                    )
```
peterthomassen, line 70: Here, you can also `try: self.authentication_config = alert_receiver_config[authentication_config_key]`, and then raise the error in the `except:` block.

andreee94: Fixed also other places where EAFP can be applied.

connaisseur/alert.py
```diff
@@ -48,6 +48,169 @@ def alerting_required(self, event_category: str) -> bool:
...
+                    ]
+                except KeyError as err:
+                    raise ConfigurationError(
+                        "No authentication configuration found for dictionary key:"
+                        f"{authentication_config_key}."
+                    ) from err
+
+                self.authentication_scheme = self.authentication_config.get(
+                    "authentication_scheme", self.authentication_scheme
+                )
```
peterthomassen, line 74: What's the second argument here? `self.authentication_scheme` would be `None` at this point, which is the default for `dict.get`.

peterthomassen (commented):
Cool, thanks! I think the code looks great now. Let's hand over to @xopham who will think about this more from the functionality side.

Hope you didn't mind my nit-picking :-)

docs/features/alerting.md
```diff
@@ -52,6 +61,54 @@ alerting:
...
+        receiver_authentication_type: bearer
+        receiver_authentication_bearer: 
+          token_env: CONNAISSEUR_ADMIT_REQUEST_WEBHOOK_AUTH_TOKEN
+```
+
+You then have to set the `CONNAISSEUR_ADMIT_REQUEST_WEBHOOK_AUTH_TOKEN` environment variable referencing the bearer token secret you want to use into the connaisseur deployment.
+
+Or if you would like to use the service account token as the bearer token, you can use the following snippet:
+
+```
```
xopham, line 83: here and in other cases should be denoted as YAML
```suggestion
```yaml
```

xopham (commented):
@andreee94 thanks for contributing! Really neat work. I think reworking the authentication is definitely required for the alerting. When first introduced, it was intentionally left open via the `custom_header` field as authentication can be very divers and is hard to pin down without just accepting anything. However, as such any authorization header must be provided within the `values.yaml` which is not ideal for a secret.
Overall the alerting configuration is somewhat complex and may need reworking eventually.

I made a few comments, however I would like to discuss the general layout:
* there is a wealth of types of authorization headers which makes this difficult to design. However, as a fallback people can always use the `custom_headers` field. Thus, I think we are fine restricting us here to `Bearer` or `Basic`
* in the PR the fields `receiver_authentication_type`, `receiver_authentication_bearer` and `receiver_authentication_basic` are introduced which has redundancy in the authentication type (essentially, specifying `receiver_authentication_bearer` already defines the `receiver_authentication_type` which is thus obsolete)
* there is two ways how secrets can be specified:
  * via environment variables: while better than plain in the `values.yaml`, this is still not a native way of providing a secret in Kubernetes. The intentional solution would be a [kubernetes secret](https://kubernetes.io/docs/concepts/configuration/secret/). If we are to improve the security, we should go for the intended handling of secrets.
  * via the service account token: Unless there is concrete use-case (e.g. intentional usage of the account token in a managed kubernetes as a workload identity), I would caution against this, as (1) in the current form this allows accessing a random path on the pod and (2) this would exfiltrate the service account token that has a much broader permission out of the scope of the cluster in question.
* the field names seem long in the current form

Providing a concise, clear and compatible form seems to be very challenging and this is further complicated by the complex configuration of alerting. I would propose using the following form:
```yaml
alerting:
  admit_request:
    templates:
      - template: ecs-1-12-0 
        receiver_url: https://your.custom.domain.com/webhook/admit
        authorization_header:
          scheme: Bearer  # scheme (e.g. `Bearer`, `Basic`) used in authorization header
          # either (preferred)
          secret_name: alladins-secret  # specify the name of a Kubernetes secret to be used. The secret take the following form: `auth: base64(opensesame)` (in case of basic auth use base64(username:password)
          # or
          auth: opensesame  # secret to add after the scheme in the authorization header (in case of basic auth use base64(username:password)
```
The proposal is similar to the auth for validators. It reduces redundancy, avoids having to parse username and password that are provided as a single token anyways and allows either rendering the `auth` field in during deployment via `helm install ... -set ...` or employing an actual kubernetes secret that must be configured separately. In case, the validation for `scheme` was chosen as any string containing only letters and '-' with at maximum 20 characters, this would also allow more exotic auth headers such as OpsGenie's "authorization: GenieKey reallybadchoicetobeuniqueinaworldofstandars`. The kubernetes secret could still be added to Connaisseur as an env variable.

I apologize for the long response, however it took me some time to wrap my head around this and I am still not certain whether this is the optimal solution. Let me know what you think :pray:

docs/features/alerting.md
```diff
@@ -52,6 +61,54 @@ alerting:
...
+alerting:
+  admit_request:
+    templates:
+      - template: ecs-1-12-0 
+        receiver_url: https://your.custom.domain.com/webhook/admit
+        receiver_authentication_type: bearer
+        receiver_authentication_bearer: 
+          token_file: /var/run/secrets/kubernetes.io/serviceaccount/token
```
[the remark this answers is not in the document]
andreee94: This is just an example of loading the token from the service account. Since the service account token is just a file inside the pod I prefer not to explicitly have a bool to control this behavior but to let it generic.

For what concern the service account as authentication method, I am not sure if it makes sense to use it in a production environment with the webhook inside the cluster (To not let the token to leak outside). 
I have just tried it to experiment with the `TokenReview` and the `system:auth-delegator` cluster role.

Taking the token from a file may be used in general if the secret has been mounted as file inside the pod.

To make the use case more clear I can change the example to something more generic like `/etc/webhook/token`.

docs/features/alerting.md
```diff
@@ -52,6 +61,54 @@ alerting:
...
+  admit_request:
+    templates:
+      - template: ecs-1-12-0 
+        receiver_url: https://your.custom.domain.com/webhook/admit
+        receiver_authentication_type: bearer
+        receiver_authentication_bearer: 
+          token_env: CONNAISSEUR_ADMIT_REQUEST_WEBHOOK_AUTH_TOKEN
+```
+
+You then have to set the `CONNAISSEUR_ADMIT_REQUEST_WEBHOOK_AUTH_TOKEN` environment variable referencing the bearer token secret you want to use into the connaisseur deployment.
```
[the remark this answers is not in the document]
andreee94: Actually I am modifying the `deployment.yaml' file in the helm chart to include the possibility to load envs from secrets:

```yaml
          {{- if .Values.deployment.env }}
            {{- range $e := .Values.deployment.env }}
            - name: {{ $e.name }}
              value: {{ $e.value }}
            {{- end }}
          {{- end }}
          {{- if .Values.deployment.envValueFrom}}
            {{- range $e := .Values.deployment.envValueFrom }}
            - name:  {{ $e.name }}
              valueFrom:
                {{- if $e.valueFrom.secretKeyRef }}
                secretKeyRef:
                  name: {{ $e.valueFrom.secretKeyRef.name }}
                  key: {{ $e.valueFrom.secretKeyRef.key }}
                {{- else if $e.valueFrom.configMapKeyRef }}
                configMapKeyRef:
                  name: {{ $e.valueFrom.configMapKeyRef.name }}
                  key: {{ $e.valueFrom.configMapKeyRef.key }}
                {{- else if $e.valueFrom.fieldRef }}
                fieldRef:
                  fieldPath: {{ $e.valueFrom.fieldRef.fieldPath }}
                {{- end }}
            {{- end }}
          {{- end }}
```

```yaml

deployment:
  # Add the environmental variables to the pod in the deployment
  # NOT use this for secrets, use the envValueFrom section instead
  env:
    - name: CONNAISSEUR_ENV
      value: "admin"

  # Add environmental variables to the pod in the deployment
  # referencing secrets, configMaps, or fields.
  envValueFrom:
    - name: CONNAISSEUR_WEBHOOK_USERNAME
      valueFrom:
        secretKeyRef:
          name: connaisseur-webhook-user
          key: username
    - name: CONNAISSEUR_WEBHOOK_PASSWORD
      valueFrom:
        secretKeyRef:
          name: connaisseur-webhook-user
          key: password
```

I will commit the code shortly.

The `.deployment.envs` option is fine for non secrets environmental variables.

tests/test_alert.py
```diff
@@ -57,6 +58,76 @@
     "template": "custom",
 }
 
+receiver_config_bearer_env = {
```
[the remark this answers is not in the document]
andreee94: Most of the not tested lines are part of the code that should not be reached since the validation is performed with the schema validation:
- 66-67
- 115-116
- 151
- 158
- 205

This is in case the file reading fails for unknown reason (FileNotFoundError tested separately).
The approach mirrors the code already used to load a file in lines 277-288. 
- 177-178

These are lines not concerning this pull request:
- 241-243
- 285-286

xopham (commented):
@andreee94 sorry for taking so long with feedback. I was very busy and we had to fix a few bugs with high priority recently. I hope to get back to this next week.

PR #571, opened 2022-03-04
ci: fix and improve completeness of nightly scans

### LINKED ISSUE #527: run nightly scans on actual released connaisseur image
**Describe the feature**
currently, the trivy vulnerability scan runs on a freshly build image, but should at least in addition run on the actual remote image.

xopham:
fixes #527

## Description

* make sure nightly scans work and provide a full picture
* example see here: https://github.com/sse-secure-systems/connaisseur/runs/5424795881?check_suite_focus=true

## Checklist

### DIFF
--- a/tests/integration/integration-test.sh
+++ b/tests/integration/integration-test.sh
@@ -134,6 +134,18 @@ helm_install() {
   echo -e "${SUCCESS}"
 }
 
+helm-repo_install() {
+  # will install unconfigured Connaisseur
+  echo -n "Installing Connaisseur..."
+  helm repo add connaisseur https://sse-secure-systems.github.io/connaisseur/charts >/dev/null
+  helm install connaisseur connaisseur/connaisseur --atomic --create-namespace \
+    --namespace connaisseur >/dev/null || {
+    echo -e "${FAILED}"
+    exit 1
+  }
+  echo -e "${SUCCESS}"
+}
+
 ### UPGRADING CONNAISSEUR ####################################
 make_upgrade() {
   echo -n 'Upgrading Connaisseur...'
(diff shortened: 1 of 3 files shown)


tests/integration/integration-test.sh
```diff
@@ -153,6 +165,17 @@ helm_upgrade() {
   echo -e "${SUCCESS}"
 }
 
+helm-repo_upgrade() {
```
Starkteetje, line 168: This doesn't seem to be called anywhere

xopham: yes, extended here for completeness as otherwise there is no way to use the artifact repo with custom values.yaml and I felt I may just directly increase functionality here. Should I remove?

xopham: the way the above helm-repo_install works is deploying the artifacthub charts as is. this part allows later upgrade to a specific `values.yaml` to allow testing with non-default settings. I consider this useful, but have not implemented a test yet

Starkteetje: You consider it useful as in you do use it locally? In that case, I'd see the value but would ask that you explain the use case in a comment

xopham: see below in line 170. will move 1 line up

Starkteetje: I think for the shell code, we did comments ahead of their definition.

Would add context to the comment:
When testing locally and you installed Conaisseur via helm-repo_install (why -_ btw?), you can call this function in an adapted or added test (because you cant call this from outside this file, right?) to set ...

xopham: ok, will remove

Starkteetje (approved):

PR #578, opened 2022-03-09
[Docs] Add clarity re generating registry credential secrets

funkypenguin:
Hey guys,

This PR clarifies the process of generating / supplying registry secrets for cosign to verify images in private repositories. It's the result of some trial-and-error and debugging efforts today!

## Checklist

- [X] Added tests (if necessary)

### DIFF
--- a/docs/validators/sigstore_cosign.md
+++ b/docs/validators/sigstore_cosign.md
@@ -144,10 +144,22 @@ The secret can for example be created directly from your local `config.json` (fo
 ```bash
 kubectl create secret generic my-secret \
   --from-file=.dockerconfigjson=path/to/config.json \
-  --type=kubernetes.io/dockerconfigjson -n connaisseur
+  -n connaisseur
 ```
 
-In the above case, the secret name in Connaisseur configuration would be `secret_name: my-secret`.
+The secret can also be generated directly from supplied credentials (which may differ from your local `config.json`, using:
+
+```bash
+kubectl create secret docker-registry my-secret \
+  --docker-server=https://index.docker.io/v1/ \
+  --docker-username='<your username>' \
+  --docker-password='<your password>' \
+  -n connaisseur
+```
+
+> :octicons-light-bulb-16: **Note**: At present, it [seems to be necessary](https://github.com/sigstore/cosign/issues/587#issuecomment-1062510930) to suffix your registry server URL with `/v1/`. This may become unnecessary in the future.
+
+In the above cases, the secret name in Connaisseur configuration would be `secret_name: my-secret`.
 It is possible to provide one Kubernetes secret with a `config.json` for authentication to multiple private registries and referencing this in multiple validators.
 
 #### k8s_keychain
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -17,8 +17,8 @@ The image policy defines a set of rules which maps different images to those val
 This is done via glob matching of the image name which for example allows to use different validators for different registries, repositories, images or even tags.
 This is specifically useful when using public or external images from other entities like Docker's official images or different keys in a more complex development team.
 
-:octicons-light-bulb-16: **Note**: Typically, the public key of a known entity is used to validate the signature over an image's content in order to ensure integrity and provenance.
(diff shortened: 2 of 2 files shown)


docs/validators/sigstore_cosign.md
```diff
@@ -144,9 +144,22 @@ The secret can for example be created directly from your local `config.json` (fo
...
+```bash
+kubectl create secret docker-registry my-secret \
+  --docker-server=https://index.docker.io/v1/ \
+  --docker-username='<your username>' \
+  --docker-password='<your password>' \
+  -n connaisseur
+```
+
+!!! note
+    At present, it [seems to be necessary](https://github.com/sigstore/cosign/issues/587#issuecomment-1062510930) to suffix your registry server URL with `/v1/`. This may become unnecessary in future.
```
xopham, line 161:
```suggestion
    At present, it [seems to be necessary](https://github.com/sigstore/cosign/issues/587#issuecomment-1062510930) to suffix your registry server URL with `/v1/`. This may become unnecessary in the future. 
```
also: we try to do one sentence per line to improve editing  experience, as the newline will be interpreted as a single space when rendering

docs/validators/sigstore_cosign.md
```diff
@@ -144,9 +144,22 @@ The secret can for example be created directly from your local `config.json` (fo
...
+
+```bash
+kubectl create secret docker-registry my-secret \
+  --docker-server=https://index.docker.io/v1/ \
+  --docker-username='<your username>' \
+  --docker-password='<your password>' \
+  -n connaisseur
+```
+
+!!! note
```
xopham, line 160: I think, we should stick to a common form here as used for example here: https://github.com/sse-secure-systems/connaisseur/blob/master/docs/basics.md#admission-control-validators-and-image-policy
This renders as: https://sse-secure-systems.github.io/connaisseur/v2.5.1/basics/#admission-control-validators-and-image-policy

xopham (commented):
@funkypenguin thanks for contributing :pray: 
I have three minor comments and would take a moment to test your change.

Great spirit to directly share your learning. Much appreciated :rocket:

xopham:
@funkypenguin would be happy to support the changes. Feel free to let me know :slightly_smiling_face:

docs/validators/sigstore_cosign.md
funkypenguin: Hey @xopham! No worries, I updated to match the existing syntax. The reason I went for the admonition-style syntax, is because I noted that the admonitions plugin is installed, at https://github.com/sse-secure-systems/connaisseur/blob/master/mkdocs.yml#L36 - these would render as described at https://squidfunk.github.io/mkdocs-material/reference/admonitions/

Anyway, consistency is better than flashiness, so I've reverted to the existing syntax :)

D

xopham (approved):
lgtm! thanks for the contribution :pray: 

(took the liberty to fix some style typos introduced by the reference I provided :sweat_smile: )

xopham (approved):

docs/validators/sigstore_cosign.md
```diff
@@ -144,10 +144,22 @@ The secret can for example be created directly from your local `config.json` (fo
 ```bash
 kubectl create secret generic my-secret \
   --from-file=.dockerconfigjson=path/to/config.json \
-  --type=kubernetes.io/dockerconfigjson -n connaisseur
+  -n connaisseur
```
xopham, line 147: @funkypenguin was there a specific reason for this change or did it cause any errors? I noticed that dropping this line works with connaisseur, but means that the secret cannot anymore be used as an `imagePullSecret` since those need to have the `type` specified, see e.g. [here](https://kubernetes.io/docs/tasks/configure-pod-container/pull-image-private-registry/#registry-secret-existing-credentials).
If that is not the case, I will re-introduce it in #585

funkypenguin: It didn't seem necessary, so I removed it for simplicity. I didn't realize that the intention was to reuse it as an imagePullSecret in the connaisseur namespace, so if that's the case then it should be reintroduced :)

xopham: let's say it could be re-used as an image pull secret or at least has the same form which probably makes it easier to work with. will fix :wink:

PR #589, opened 2022-03-22
fix: digest confusion at high load

### LINKED ISSUE #583: ImageRef is mutated with the wrong digest
**Describe the bug**

This issue may be precipitated by https://github.com/sse-secure-systems/connaisseur/issues/582. Our CI platform routinely deploys about 40 pods into a fresh namespace to test code changes. We noticed that after deploying connaisseur, some of the time, the pods in these namespaces fail with an error like this:

```
  Warning  Failed       12s                kubelet            Failed to pull image "docker.io/gergorg/rtdb@sha256:8f76cb854b018213597d910fb4a1ad7d5d5ed775fbe328594b9d6347a48c38f9": rpc error: code = NotFound
 desc = failed to pull and unpack image "docker.io/myorg
/rtdb@sha256:8f76cb854b018213597d910fb4a1ad7d5d5ed775fbe328594b9d6347a48c38f9": failed to resolve reference 
"docker.io/myorg/rtdb@sha256:8f76cb854b018213597d910fb4a1ad7d5d5ed775fbe328594b9d6347a48c38f9": docker.io
/myorg/rtdb@sha256:8f76cb854b018213597d910fb4a1ad7d5d5ed775fbe328594b9d6347a48c38f9: not found
```

The digest ref is the result of a deployment being mutated by connaisseur, having the image tag changed into the digest format. 

**However**... the actual digest applied belongs to a **different** image in a **different** repository, which was deployed by CI at the same time into the namespace:

Harbor 2022-03-17 16-26-56

It would seem that connaisseur "mixed up" the digests of the various deployments/pods which were deployed at the same time (potentially due to an error parsing the CRs), and substituted the digest for one image, with another.

**Expected behavior**

Connaisseur only mutates imageRefs matching the repository and image which was originally intercepted

**Optional: To reproduce**

This is hard, since the issue seems to be intermittent. It **seems** to co-inside with the errors generated by https://github.com/sse-secure-systems/connaisseur/issues/582. To reproduce, try deploying about 20 different images simultaneously, mixed in with some CRDs which include an image ref.

 - Connaisseur: 2.5.1

xopham:
Fixes #583

## Description

* it seems that as of v2.5.0 digest confusion can occur
  * a referenced image gets the digest of another image assigned and admission request is patched accordingly
  * causes deployments potentially falsely allowed, but failing due to broken request patching
* the root issue appears to be unexpectatly related to updating a class variable inside concurrent futures / asyncio
* the class variable is removed and inputs are passed into the concurrent tasks, results returned and then used to update the resuts

## Checklist

### DIFF
--- a/connaisseur/validators/cosign/cosign_validator.py
+++ b/connaisseur/validators/cosign/cosign_validator.py
@@ -24,13 +24,11 @@ from connaisseur.validators.interface import ValidatorInterface
 class CosignValidator(ValidatorInterface):
     name: str
     trust_roots: list
-    vals: dict  # validations, a dict for each required trust root containing validated digests or errors
     k8s_keychain: bool
 
     def __init__(self, name: str, trust_roots: list, auth: dict = None, **kwargs):
         super().__init__(name, **kwargs)
         self.trust_roots = trust_roots
-        self.vals = {}
         self.k8s_keychain = False if auth is None else auth.get("k8s_keychain", False)
 
     async def validate(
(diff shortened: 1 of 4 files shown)


connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -43,25 +41,25 @@ async def validate(
...
         executor = ThreadPoolExecutor(num_workers)
         # prepare tasks
-        arguments = [(k, str(image)) for k in self.vals.keys()]
+        arguments = [(k, v.copy(), str(image)) for k, v in vals.items()]
         futures = [executor.submit(self.__validation_task, *arg) for arg in arguments]
-        # await results (output dropped as `self.vals` is updated within function)
+        # await results (output dropped as `vals` is updated within function)
         for future in futures:
-            future.result()
+            vals.update(future.result())
```
Starkteetje, line 59: Does then the last error message win out? Should we be a bit more careful there and combine them to give accurate feedback?

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -43,25 +41,25 @@ async def validate(
...
         # tasks = [self.__validation_task(k, str(image)) for k in self.vals.keys()]
         # await asyncio.gather(*tasks)
 
         # prepare executor
-        num_workers = len(self.vals)
+        num_workers = len(vals)
         executor = ThreadPoolExecutor(num_workers)
         # prepare tasks
-        arguments = [(k, str(image)) for k in self.vals.keys()]
+        arguments = [(k, v.copy(), str(image)) for k, v in vals.items()]
```
Starkteetje, line 55: I'd add a comment there to justify the usage of v.copy as I wouldn't say its obvious. May help someone in the future touching this

xopham: Will add a comment

Starkteetje (approved):

Starkteetje (approved):

PR #606, opened 2022-03-29
Add optional objectValidation to webhook

funkypenguin:
Hey guys,

Here's a PR which implements "objectValidation", much the same way that "namespacedValidation" is implemented. It allows the user to optionally include/ignore specific objects by attaching the `securesystemsengineering.connaisseur/webhook` to them via their parent controller/objects (_deployment, statefulset, etc_).

## Description

Why do we need this? There may be many more edge cases, but in my case I use FoundationDB Operator to manage multiple database clusters. The operator deploys sidecars to database pods. When performing a version upgrade, the operator identifies targets for upgrade by finding pods running the not-latest version of the sidecar, and attempts to update them.

With Connaisseur in the mix, all database pods get mutated, and don't have the expected image tag, so I get stuck in a perpetual loop of "lets-upgrade-the-non-latest-images", with the operator "battling" Connaisseur for how the sidecar image should be specified!

An easy fix (_which I'm now using locally_) has been to add the `objectSelector` field to the webhook, so that these particularly tricky pods can be excluded altogether from connaisseur validation and mutation.

I've updated the chart version, docs, and templates, but I might need assistance updating any necessary tests!

Cheers!
D

## Checklist

### DIFF
--- a/docs/features/namespaced_validation.md
+++ b/docs/features/namespaced_validation.md
@@ -1,6 +1,6 @@
 # Namespaced Validation
 
-Namespaced validation allows restricting validation to specific namespaces.
+Namespaced validation allows restricting validation to specific namespaces, and can operate alongside [object validation](./object_validation.md).
 Connaisseur will only verify trust of images deployed to the configured namespaces.
 This can greatly support initial rollout by stepwise extending the validated namespaces or excluding specific namespaces for which signatures are unfeasible.
 
--- a/docs/features/object_validation.md
+++ b/docs/features/object_validation.md
@@ -1,0 +1,46 @@
+# Object Validation
+
+Object validation allows restricting validation to specific objects, and can operate alongside [namespace validation](./namespaced_validation.md).
+Connaisseur can optionally include/exclude the verification trust of images based on pod labels.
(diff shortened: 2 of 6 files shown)


phbelitz:
Hello @funkypenguin! Thank you very much for your contribution, the PR look pretty good. 

However I still need some clarification to fully understanding the problem. Your goal is to exclude very specific pods and not just image patterns, do I understand correctly? So setting up a policy rule like this:

```yaml
policy:
  - pattern: "<database-image>:*"
    validator: allow
```

doesn't work in your use-case, since this would skip all database images, regardless of version and where they are deployed. You only want to skip a selected few pods, yes?

funkypenguin:
Hey @phbelitz! I'd assumed that using an "allow" validator as above would still mutate the image into the `image@sha256:digest` format. If that's true, then this wouldn't fix my issue, since the database operator requires the original image tag in order to determine the target version.

If not, then using an `allow` validator probably **would** work, for my use-case :)

Cheers!
D

xopham:
@funkypenguin `allow` validator completely skips validation and should thus fix your issue. Will close accordingly. Feel free to re-open if otherwise :slightly_smiling_face:

PR #620, opened 2022-04-08
Fix/pin and update prometheus dependencies

Starkteetje:
## Description

Fixes break of Connaisseur by incompatible versions of prometheus-flask-exporter and prometheus-client (indirect dependency, not pinned by prometheus-flask-exporter)

Match is exact on purpose, as the change may be reverted https://github.com/rycus86/prometheus_flask_exporter/pull/129#issuecomment-1092382980 breaking installations again if the change is breaking without being marked as such and prometheus-flask-exporter doesn't implement version pinning since then (and we update to it for all releases we do in the meantime)

If they introduce pinning, we can remove the indirect dependency again.

## Checklist

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -2,8 +2,8 @@ apiVersion: v2
 name: connaisseur
 description: Helm chart for Connaisseur - a Kubernetes admission controller to integrate container image signature verification and trust pinning into a cluster.
 type: application
-version: 1.3.2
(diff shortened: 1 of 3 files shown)


Starkteetje:
Since its breaking current installations, this may actually be a fix version for Connaisseur. Thoughts @xopham @phbelitz ?

phbelitz (approved):

PR #622, opened 2022-04-08
ci: use ghcr, signatures, sboms, slsa, reusable workflows, ossf

### LINKED ISSUE #624: GitHub Container Registry for CI
**Describe the feature**
GitHub offers the [GitHub Container Registry](https://docs.github.com/en/packages/working-with-a-github-packages-registry/working-with-the-container-registry) (ghcr) for container images (private and public). This can be used within the CI to store images.

Using a container registry is a more native approach to pass images between different CI steps and can simplify testing and publication.

* prepare more re-usable actions
* [security harden github actions](https://docs.github.com/en/github-ae@latest/actions/security-guides/security-hardening-for-github-actions)
* use BuildKit/Buildx to speed up building
* use GITHUB_TOKEN and only-read tokens for security during testing (e.g. imagePullSecrets)
* sign/verify using cosign
* build branch images (master, tag, dev, <any-branch>), however make sure to not create image confusion (digest for reference)
* inject connaisseur image reference, validator and policy into `helm/values.yaml` to declutter testing
* potentially provide release candidates or dev builds
* check storage limits
* ensure frequent cleanup (on branch merge, as cron)

Please add further ideas below!

xopham:
Fixes #624 

## Description

Major changes:
- new build system
  - implement GHCR to store images as packages in pipeline
  - use buildkit for image build -> SBOMs, Provenance
  - sign packages using cosign
- modular CI ([reusable workflows](https://docs.github.com/en/actions/using-workflows/reusing-workflows) and actions)
- refactor CI

Improvements:
- integrated with GitHub Security
- modular CI
- faster builds
- signed builds
- nightly builds
- provenance and SBOMs added
- broader code scanning coverage

TBD:
- [ ] summary update
- [ ] rm trivy-released-image-scan into sca job
- [ ] join docs into main workflows
- [ ] update "required tests"
- [ ] matrix strategy for `dockerhub-check`

Open Question:
- OK to add registry to helm chart? seems more explicit and less reliant on a random default
- OK to rename "nightly-scans" to "nightly"?
- can we drop tag engineering?
- should we make everything reusable with selector for which jobs to drop?
- now only relying on Github security not pipeline fail
- IMPORANT: mitigate risk of overwriting tags!

Missing (create issues for remaining):
- [ ] release workflow + test
- [x] somehow `getRoot` fails on nightly
- [ ] update public getRoot image
- [ ] rework all actions
- [ ] consider caching for build and integration tests
- [ ] speedup by reduction of integration tests
- [x] remove `update` upload
- [x] consider usage of https://github.com/marketplace/actions/docker-metadata-action
- [ ] use GH action specific parsing: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions
- [ ] add verify tag functionality for cosign validation to connaisseur
- [ ] sign scans
- [ ] add semgrep
- [x] add checkov
- [ ] automated local testing with remote builds
- [ ] confirm dependabot update actions by commitSHA (checkout actions are purposely outdated by two minor versions)
- [ ] finetune [checkov](https://github.com/marketplace/actions/checkov-github-action)
- [ ] resolve OSSF Scorecard findings
- [ ] publish OSSF Scorecard findings
- [ ] merge upgrade int test into bash script
- [ ] why has cosign [output changed](https://github.com/sse-secure-systems/connaisseur/actions/runs/4258789437/jobs/7410395811)? 
- [ ] harmonize usage of `yq` (see https://github.com/sse-secure-systems/connaisseur/pull/622#discussion_r1109874661)
- [ ] consider dropping `Uninstall` when not necessary (see https://github.com/sse-secure-systems/connaisseur/pull/622#discussion_r1116865206)
- [ ] switch back to strict github security evaluation

## Checklist


.github/workflows/release.yaml
```diff
@@ -2,33 +2,51 @@ name: release
...
-          IMAGE_TAG=$(yq e '.deployment.image' helm/values.yaml| cut -d ':' -f2)
+          IMAGE_TAG=${{ needs.context.outputs.original_tag }}
           for COMMIT_TAG in $(git tag --points-at $(git rev-parse HEAD)); do
             if [[ "$IMAGE_TAG" == "$COMMIT_TAG" ]]; then
               exit 0
             fi
           done
           echo "Tag '$IMAGE_TAG' is not within tags of commit: $(git tag --points-at $(git rev-parse HEAD))"
-          exit 1
+          exit 0
```
phbelitz, line 45: why 0? shouldn't the pipeline fail if there is a mismatch in versions?

.github/workflows/nightly-scans.yaml
```diff
@@ -193,20 +229,58 @@ jobs:
...
         run: |
           sudo snap install yq
       - uses: ./.github/actions/k8s-version-config
         name: Setup k8s cluster
         with:
           k8s-version: ${{ matrix.k8s-version }}
-          load-local-image: false
       - name: Run pre-config and workload integration tests
         run: |
-          bash tests/integration/integration-test.sh "nightly-pre-and-workload"
```
phbelitz, line 206: is the `"nightly-pre-and-workload"` step in the integration test still needed after those changes?

tests/integration/integration-test.sh
```diff
@@ -168,8 +175,27 @@ load_test() { #
   rm output.log
 }
 
+### CREATE IMAGEPULLSECRET ####################################
+create_imagepullsecret() { # NAMESPACE # CREATE
+  local CREATE=${2:-true}
+  if $CREATE; then
+    echo -n "Creating Namespace '${1}'..."
```
phbelitz, line 182: nit: consider using the `${SUCCESS}` and `${FAILED}` shortcuts. or do it your way 🤷 same for the other `echo` outputs.

.github/workflows/nightly-scans.yaml
```diff
@@ -193,20 +229,58 @@ jobs:
...
+          cut-off: four hours ago UTC+1
+          timestamp-to-use: updated_at
+          account-type: org
+          org-name: sse-secure-systems
+          token: ${{ secrets.GHCR_PAT }}
+      - name: Cleanup all connaisseur images
+        uses: snok/container-retention-policy@v1
+        with:
+          image-names: connaisseur
+          skip-tags: develop, v*, sha256-*
```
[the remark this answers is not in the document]
xopham: yes, we need to keep the cosign signatures. See: https://github.com/sse-secure-systems/connaisseur/pkgs/container/connaisseur

I do not yet have a deletion strategy for the signatures yet, but will look into it in the future. For now all those packages are private.

.github/workflows/cicd.yaml
```diff
@@ -142,69 +164,44 @@ jobs:
...
 
   trivy-vuln-scan:
     name: trivy vuln scan
+    permissions:
+      packages: read
     runs-on: ubuntu-latest
     container:
       image: docker:stable
-    needs: [build]
+    needs: [build,context]
```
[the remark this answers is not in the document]
xopham: I believe it is needed to use the outputs. Which I believe fails if the respective job is not explicitly mentioned.

.github/workflows/cicd.yaml
```diff
@@ -337,32 +351,50 @@ jobs:
...
+        run: |
+          kubectl create ns connaisseur
+          kubectl create secret generic ${imagepullsecret} \
+            --from-file=.dockerconfigjson=$HOME/.docker/config.json \
+            --type=kubernetes.io/dockerconfigjson \
+            --namespace=connaisseur
       - uses: actions/checkout@v3
         with:
           ref: "master"
+      - name: Checkout required current files
```
Starkteetje, line 387: If it's about current files, why does the checkout action above not suffice?

Also, why do we need a 3rd party integration for this?

.github/workflows/cicd.yaml
```diff
@@ -337,32 +351,50 @@ jobs:
...
+        uses: Bhacaz/checkout-files@v2
+        with:
+          files: tests/integration/ghcr-values.yaml
+          branch: ${{ github.head_ref || github.ref_name }}
       - name: Configure Connaisseur (master)
         run: |
-          yq e '.detectionMode=true' -i helm/values.yaml
+          envsubst < tests/integration/ghcr-values.yaml > update
+          yq '. *+ load("update")' -i helm/values.yaml
```
Starkteetje, line 395: Not knowing what this does, I'd assume its similar to below. If so, why different syntax. If not, maybe a comment would be nice

.github/workflows/nightly-scans.yaml
xopham: why would it not? This will test the latest build

xopham: oh, hold on this may have been a mishap

.github/workflows/nightly-scans.yaml
```diff
@@ -193,20 +229,58 @@ jobs:
...
+    steps:
+      - name: Cleanup test images in 'connaisseur-test'
+        uses: snok/container-retention-policy@v1
+        with:
+          image-names: connaisseur-test
+          cut-off: three weeks ago UTC+1
+          timestamp-to-use: updated_at
+          account-type: org
+          org-name: sse-secure-systems
+          token: ${{ secrets.GHCR_PAT }}
```
Starkteetje, line 270: I guess that's a personal token of yours which is only allowed to read/write the registry?

.github/workflows/release.yaml
```diff
@@ -2,33 +2,51 @@ name: release
...
         run: |
-          IMAGE_TAG=$(yq e '.deployment.image' helm/values.yaml| cut -d ':' -f2)
+          IMAGE_TAG=${{ needs.context.outputs.original_tag }}
           for COMMIT_TAG in $(git tag --points-at $(git rev-parse HEAD)); do
             if [[ "$IMAGE_TAG" == "$COMMIT_TAG" ]]; then
               exit 0
             fi
           done
           echo "Tag '$IMAGE_TAG' is not within tags of commit: $(git tag --points-at $(git rev-parse HEAD))"
-          exit 1
```
Starkteetje, line 27: Why change this? Seems like a guardrail that was put in place for a reason

tests/integration/integration-test.sh
```diff
@@ -16,6 +16,13 @@ RETRY=3
 ## Backup helm/values.yaml
 cp helm/values.yaml values.yaml.backup
 
+## Join ghcr integration yaml
+if ! [[ -z "${IMAGE+x}" || -z "${imagepullsecret+x}" ]]; then
```
Starkteetje, line 20: if not (x unset or y unset) is way harder on cognition than if x and y

tests/integration/integration-test.sh
```diff
@@ -168,8 +175,27 @@ load_test() { #
...
 }
 
+### CREATE IMAGEPULLSECRET ####################################
+create_imagepullsecret() { # NAMESPACE # CREATE
+  local CREATE=${2:-true}
+  if $CREATE; then
+    echo -n "Creating Namespace '${1}'..."
+    kubectl create ns ${1}
+  fi
+  if ! [[ -z "${imagepullsecret+x}" ]]; then
```
Starkteetje, line 191: Wouldn't 

```suggestion
  if [[ -n "${imagepullsecret+x}" ]]; then
```

also work and be way easier to comprehend?

tests/integration/integration-test.sh
```diff
@@ -168,8 +175,27 @@ load_test() { #
   rm output.log
 }
 
+### CREATE IMAGEPULLSECRET ####################################
+create_imagepullsecret() { # NAMESPACE # CREATE
+  local CREATE=${2:-true}
+  if $CREATE; then
+    echo -n "Creating Namespace '${1}'..."
+    kubectl create ns ${1}
```
Starkteetje: Creation of a NS is a weird and unintuitive consequence of creating a secret

tests/integration/update-for-workloads.yaml
```diff
@@ -1,5 +1,4 @@
 deployment:
-  imagePullPolicy: Never
```
Starkteetje, line 2: It looks like this shouldn't break local integration tests, but there may also have been a reason why we set it to `Never` instead of default. Have you tried running local integration tests with this setup?

.github/actions/context/action.yaml
```diff
@@ -0,0 +1,105 @@
...
+        elif [ "${{ github.ref_name }}" = "develop" ]; then
+          BUILD_TAG="develop"
+        else
+          BUILD_REPO="${BUILD_REPO}-test"
+          BUILD_TAG=${{ github.run_id }}-${{ github.run_number }}-${{ github.run_attempt }}
+        fi
+
+        BUILD_IMAGE=${BUILD_REPO}:${BUILD_TAG}
+
+        echo COSIGN_VERSION=${COSIGN_VERSION} >> $GITHUB_OUTPUT
```
Starkteetje, line 71: nit
```suggestion
        echo COSIGN_VERSION=${COSIGN_VERSION} >> ${GITHUB_OUTPUT}
```

.github/actions/build/action.yml
```diff
@@ -0,0 +1,120 @@
+name: build
+description: 'build connaisseur image'
```
Starkteetje, line 2: Do we do Capitalization in actions?

Starkteetje: bc below it's `Install Cosign`

.github/actions/context/action.yaml
Starkteetje: Same below and other action file(s)

.github/actions/build/action.yml
```diff
@@ -0,0 +1,120 @@
...
+      id: build
+      uses: docker/build-push-action@v4
+      with:
+        push: true
+        tags: ${{ inputs.image-repo }}:${{ inputs.image-tag }}
+        file: docker/Dockerfile
+        build-args: COSIGN_VERSION=${{ inputs.cosign-version }}
+        sbom: true
+        provenance: true
+    - name: Sign image with a key
```
Starkteetje, line 48: We did in the past conciously decide against signing the image with the notary key within the pipeline. Reasons were that we run stuff within the pipeline that's not under our control and accept PRs from contributors who could then exfil the key, making the signatures void. Those arguments still seem to apply

Starkteetje (changes requested):

.github/actions/build/action.yml
xopham: Yes, I agree. However, this does not replace the release signing so far, but only signs for the internal pipeline for now. This may evolve into a pipeline that allows for automated signed releases and then will have to either reconsider or require an additional signature by an offline key.
However, so far this only affects the intermediate testing images and we could for example consider to make the nightly images public with a "less secure" key that is stored in github. Still, the images uploaded so far are private and not publicly available

.github/actions/context/action.yaml
```diff
@@ -0,0 +1,111 @@
...
+
+        if [ "${{ github.ref_name }}" = "master" ]; then
+          BUILD_TAG="${ORIGINAL_TAG}"
+          if [ "${{ github.workflow }}" = "nightly-scans" ]; then
+            BUILD_TAG=${BUILD_TAG}-nightly
+          fi
+        elif [ "${{ github.ref_name }}" = "develop" ]; then
+          BUILD_TAG="develop"
+        else
+          #TODO: just for test
```
Starkteetje, line 59: remaining TODO

.github/actions/context/action.yaml
xopham: I do believe that this does actually not work

.github/actions/context/action.yaml
```diff
@@ -0,0 +1,111 @@
...
+        CHART_VERSION=${{ steps.get_chart_version.outputs.result }}
+        COSIGN_VERSION=$(grep -Eo '^COSIGN_VERSION = .*' Makefile | grep -Eo '[0-9]+\.[0-9]+\.[0-9]+')
+        ORIGINAL_IMAGE=${{ steps.get_original_image.outputs.result }}
+        ORIGINAL_REPO=$(echo "${ORIGINAL_IMAGE}" | cut -d ":" -f 1)
+        ORIGINAL_TAG=$(echo "${ORIGINAL_IMAGE}" | cut -d ':' -f 2)
+        BUILD_REPO=ghcr.io/${{ github.repository }}
+
+        if [ "${{ github.ref_name }}" = "master" ]; then
+          BUILD_TAG="${ORIGINAL_TAG}"
+          if [ "${{ github.workflow }}" = "nightly-scans" ]; then
```
Starkteetje, line 53: If you intend to ever switch to ghcr as an additional or replacement registry for the main Connaisseur image, this looks super frail

.github/actions/build/action.yml
xopham: don't think we have much consistency. I am planning to do a full refactoring after this is merged. Also the above is a description and `install cosign` is a job name....

.github/workflows/cicd.yaml
xopham: utlimately, it overwrites/extends the other yaml with another yaml provided via 'load'. The other style of writing still exists throughout our code at numerous places and below I essentially just copy-pasted (I believe). Want to clean that up in a general refactoring after this MR

xopham: I would not think this is worthy of a comment

.github/workflows/nightly-scans.yaml
xopham: correct, actually only `packages:delete` which includes `read`

.github/workflows/release.yaml
xopham: yeap...that was by accident as is being fixed. Good catch

tests/integration/integration-test.sh
xopham: sorry, I don't think I fully understand the proposal

tests/integration/update-for-workloads.yaml
xopham: nope, only CI

tests/integration/integration-test.sh
xopham: done

tests/integration/integration-test.sh
xopham: please see my fix

tests/integration/integration-test.sh
xopham: yes, but cannot use an imagePullSecret from another NS and wanted to avoid code duplication or unintuitively adding a namespace all the time. What do you propose?

.github/actions/context/action.yaml
xopham: yes, I added it to run all workflows again. Sorry, that was my rebase

.github/actions/context/action.yaml
xopham: Why?

.github/actions/context/action.yaml
xopham: was wrong, it does work :sweat:

.github/actions/context/action.yaml
xopham: we could use the `docker metadata` action for that which I newly added and do it based on a schedule, but I am not entirely certain

.github/workflows/cicd.yaml
```diff
@@ -4,396 +4,83 @@ on:
   push:
     branches:
       - master
+      - develop
```
Starkteetje, line 7: Why this addition? To have the image:develop image "published"?

.github/workflows/.reusable-sast.yml
```diff
@@ -0,0 +1,139 @@
...
+  checkov:
+    permissions:
+      security-events: write
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
+      - name: Render Helm charts
+        run: |
+          rm -rf tests
```
phbelitz, line 126: why? tests didn't do anything to you.

tests/integration/integration-test.sh
Starkteetje: Maybe rename to something that at least hints at the dependency on ns, like `create_imagepullsecret_in_ns`

Otherwise would just do the creation and fail if ns not present and add ns in respective locations explicitly (and if unintuitive, add comment)

.github/workflows/cicd.yaml
xopham: I don't think I fully understand the comment. What do you mean by "It's really different"? You mean it looks different or it creates different output?
Ultimately, the former version appears to use `*` at its core and I use `*+`. Mine extends, the former overwrites arrays. My plan was to harmonize the writings in the next round when I really get to the `integration-test.sh`

tests/integration/integration-test.sh
xopham: OK, renamed it to be more explicit since it is tests (not production code) and code duplication is more painful than explicitness is valuable in tests. hope that is fine

tests/integration/integration-test.sh
```diff
@@ -263,14 +299,23 @@ update_values() { # [EXPRESSION...]
   done
 }
 
+update_values_minimal() {
```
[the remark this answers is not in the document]
xopham: good catch. this remained from an earlier version where the `ghcr-values` was created in those function. I removed `rm`

.github/workflows/.reusable-integration-test.yml
```diff
@@ -0,0 +1,248 @@
...
+      - name: Login with registry
+        uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a # v2.1.0
+        with:
+          registry: ${{ inputs.build_registry }}
+          username: ${{ inputs.repo_owner }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Install yq and bash
+        run: |
+          sudo snap install yq
+          sudo apt update
```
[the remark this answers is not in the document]
xopham: no idea, removed it. planning to fully go over the workflows and integration tests again as mentioned above.

.github/workflows/.reusable-integration-test.yml
```diff
@@ -0,0 +1,248 @@
...
+        run: |
+          kubectl get pods -n connaisseur -o jsonpath="{.items[*].spec.containers[*].image}"
+      - name: Run integration tests (branch)
+        run: |
+          bash tests/integration/upgrade-integration-test.sh
+      - name: Display k8s logs if integration test failed
+        if: failure()
+        run: |
+          kubectl logs -n connaisseur -lapp.kubernetes.io/name=connaisseur --prefix=true
+      - name: Uninstall Connaisseur
```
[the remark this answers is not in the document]
xopham: we could though it feels a bit incomplete :thinking: but I get your point. Maybe let's add it to my list and I have a strong opinion next week.

.github/workflows/cicd.yaml
xopham: that and upload security artifacts on all protected branches to ensure we can understand the state of security. I could however drop the integration tests bc those already ran on both branches and were a duplicate during release in any case

.github/workflows/.reusable-sast.yml
xopham: added a comment

.github/workflows/cicd.yaml
Starkteetje: If `yq '. *+ load("update")' -i helm/values.yaml` and `yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' helm/values.yaml update` do the same, why use both and not just one of those consistently?

Starkteetje: What I meant is that `yq '. *+ load("update")' -i helm/values.yaml` has a way different yq syntax from `yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)'`  and seems to do a similar job, in which case we shouldn't use different syntax.
But if yours updates and extends and the select syntax only updates, then there's a good reason to use "your" syntax instead of the old one (though if it's just more powerful we should also switch the other's over (at some point))

xopham: will be in the next refactor

Starkteetje (approved):

phbelitz (approved):

PR #632, opened 2022-04-22
build: Remove transitive dependency pinning

Starkteetje:
## Description

In light of #620 we pinned a transitive dependency to fix incompatibility introduced by a breaking change in said transitive dependency. While being a reasonable quick fix at the time, it imposes undue limits on our dependency. Thus this commit removes the pinning of the transitive dependency and should they again break (the transitive dependency took notice and the dependency took steps to mitigate), we may need to reeavluate the dependency as a whole instead of pinning dependencies that are not ours to pin.

## Checklist

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -5,7 +5,6 @@ Flask~=2.1.1
 Jinja2~=3.1.1
 jsonschema~=4.4.0
 parsedatetime~=2.6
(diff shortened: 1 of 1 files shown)


xopham (approved):
fair point

xopham (approved):
lgtm

PR #663, opened 2022-05-23
Small change to make connoisseur running under OpenShift / OKD 4

pflaeging:
This is needed for OpenShift / OKD 4

If we want to make connoisseur runnable under OpenShift / OKD 4, the easiest way is:
- disable securityContext in deployments completely.

Fixes #

## Description

Create a variable, which can disable securityContexts.

## Checklist

- [X ] Extended README/Documentation (if necessary)

### DIFF
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -85,8 +85,10 @@ spec:
                   fieldPath: metadata.name
           resources:
             {{- toYaml .Values.deployment.resources | nindent 12 }}
+          {{- if .Values.deployment.securityContext.enabled }}
           securityContext:
             {{- toYaml .Values.deployment.securityContext | nindent 12 }}
+          {{- end }}
         {{- if .Values.deployment.extraContainers }}
         {{ tpl (toYaml .Values.deployment.extraContainers) . | nindent 8}}
         {{- end }}
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -38,6 +38,7 @@ deployment:
   #annotations:  # uncomment when using Kubernetes prior v1.19
   #  seccomp.security.alpha.kubernetes.io/pod: runtime/default  # uncomment when using Kubernetes prior v1.19
   securityContext:
+    enabled: true # enable if you want to use it, disable this for OKD / OpenShift 4
     allowPrivilegeEscalation: false
     capabilities:
       drop:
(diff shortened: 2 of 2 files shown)


xopham:
@pflaeging shouldn't it be sufficient to just remove contents of securityContext and pass an empty list?
Or we could allow just removing it by deleting the whole section, including `securityContext` by allowing no securityContext to be configured at the following: https://github.com/sse-secure-systems/connaisseur/blob/master/helm/templates/deployment.yaml#L88

I am not so much in favor of creating a non-native flag simply for one flavor of k8s.

pflaeging:
> @pflaeging shouldn't it be sufficient to just remove contents of securityContext and pass an empty list? Or we could allow just removing it by deleting the whole section, including `securityContext` by allowing no securityContext to be configured at the following: https://github.com/sse-secure-systems/connaisseur/blob/master/helm/templates/deployment.yaml#L88
> 
> I am not so much in favor of creating a non-native flag simply for one flavor of k8s.

I can try this manually and give you feedback. For me it's OK with an empty list, lets see what OPenShift says ;-)

xopham:
@pflaeging is there a way to get direct feedback from openshift? would be interesting to get their take. Happy to support making the overall entry of `securityContext` optional. Let me know if I should contribute.

xopham:
@pflaeging nevermind, I got the point. There is no helm-native way to remove a set value, so this seems like the most at hand solution to the issue. Will move it forward.
Thanks for pointing this out :pray:

xopham:
@pflaeging I proposed a minor rewording. Currently, testing fails because we are setting `securityContext.enabled` with your change and that field is not allowed. A better way maybe as done in the [cloudone](https://github.com/trendmicro/cloudone-container-security-helm/blob/master/values.yaml) but that is a breaking change for us and I'd rather not have that. We could add `disableSecurityContext` field before `securityContext` which is also "not nice" or allow setting it to empty. What do you think?

xopham:
@pflaeging I proposed a fix to simply set `securityContext: null` which should remove the context, is compatible with helm-native configuration and hopefully fixes the problem for OKD/Openshift. Could you test and let me know?

Also please rebase to `develop` so it can be merged or let me know if I should do it (just a bit hesitant since it is your `master` branch) :slightly_smiling_face:

pflaeging:
> @pflaeging any feedback on this? Otherwise, I'd like to go ahead and merge next week 🙂

Sorry, I'm back from Covid right now and check ist back, ...

pflaeging:
OK, we can cancel the pull request. I've tried the following and it works ;-):

In your `values.yaml` set:

~~~yaml
  securityContext: null
  podSecurityPolicy:
    enabled: false
    name: ["connaisseur-psp"] 
~~~

Then I'm rolling the helm chart out with:
`helm install connaisseur connaisseur/helm --atomic --namespace infra-connaisseur -f values.yaml`

And it run's without patching anything.

If you show me the place I can give a small advice in the documentation how to install clean on OKD / OCP (we are using cosign).

Greetings 
:peter

PR #711, opened 2022-07-08
ci: fix webhook unavailable errors

xopham:
Fixes #

## Checklist

### DIFF
--- a/.github/workflows/cicd.yaml
+++ b/.github/workflows/cicd.yaml
@@ -229,10 +229,6 @@ jobs:
         name: Setup k8s cluster
         with:
           k8s-version: v1.22
-      - name: Load Image
-        run: |
-          sudo k3s ctr images import ${GITHUB_SHA}_image.tar
-        shell: bash
       - name: Set environment variables for alerting listener
         run: |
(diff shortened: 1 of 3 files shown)


.github/workflows/cicd.yaml
```diff
@@ -229,10 +229,6 @@ jobs:
         name: Setup k8s cluster
         with:
           k8s-version: v1.22
-      - name: Load Image
-        run: |
-          sudo k3s ctr images import ${GITHUB_SHA}_image.tar
-        shell: bash
```
xopham, line 235: why is this not necessary anymore?

phbelitz: because the `k8s-version-config` already loads the image if the `load-local-image` flag is set to `true`, which it is by default.

https://github.com/sse-secure-systems/connaisseur/blob/master/.github/actions/k8s-version-config/action.yaml#L30-L36

xopham: when was that added?

xopham (commented):
lgtm

phbelitz (approved):

PR #759, opened 2022-08-19
feat: validation only mode

### LINKED ISSUE #270: Build a validation only mode
**Describe the feature**
Connaisseur works as a mutating admission controller and thus replaces tags by signed digest. This also means that requests are modified. A validation only mode would simply check whether images comply with the policy (e.g. only referenced by signed digests or excluded) but fail if the a reference would require mutation.

It was reported that some enterprises are uneasy with implementing a mutating admission controller. (Might be publicly hinted though not explicit [in this blog post](https://medium.com/microsoftazure/validating-content-trust-metadata-for-kubernetes-targeted-container-images-at-deployment-time-360cafc33ef4)).

One could implement Connaisseur as a validating admission controller and simply deny, in case a digest is invalid or mutation  is required.

As a bonus, the validation mode could also be enabled with mutation mode in parallel. As validating controllers run after mutating, this would ensure no other AC introduced an insecure image.

xopham:
relates to #712 

## Description

allow configuration to not mutate image references but simply validate whether signed digest exists

## Checklist

### DIFF
--- a/connaisseur/flask_application.py
+++ b/connaisseur/flask_application.py
@@ -24,7 +24,7 @@ sends its response back.
 """
 CONFIG = Config()
 DETECTION_MODE = os.environ.get("DETECTION_MODE", "0") == "1"
-VALIDATION_MODE_MUTATE = os.environ.get("VALIDATION_MODE_MUTATE", "true") == "true"
+VALIDATION_MODE_MUTATE = os.environ.get("VALIDATION_MODE_MUTATE", "true") != "false"
 
 metrics = PrometheusMetrics(
     APP,
--- a/connaisseur/flask_application.py
+++ b/connaisseur/flask_application.py
@@ -24,6 +24,7 @@ sends its response back.
 """
 CONFIG = Config()
 DETECTION_MODE = os.environ.get("DETECTION_MODE", "0") == "1"
+VALIDATION_MODE_MUTATE = os.environ.get("VALIDATION_MODE_MUTATE", "true") == "true"
 
 metrics = PrometheusMetrics(
     APP,
@@ -144,7 +145,9 @@ async def __admit(admission_request: AdmissionRequest):
     return get_admission_review(
         admission_request.uid,
         True,
-        patch=[patch for patch in patches.result() if patch],
+        patch=[patch for patch in patches.result() if patch]
+        if VALIDATION_MODE_MUTATE
+        else None,
     )
 
 
--- a/docs/features/validation_mode.md
+++ b/docs/features/validation_mode.md
@@ -1,0 +1,32 @@
+# Validation Mode
+
+Validation mode allows configuration whether Connaisseur mutates image references (tags to trusted digests).
+
+In default behavior during validation, Connaisseur uses the image reference (tag or digest) from the admission request to identify a trusted digest (digest with valid signature from a configured trust root) and modifies the reference to the trusted digest if one exists or denies the request otherwise.
+This ensures that the container runtime pulls and spins up a container of a validated image, as the digest is an immutable inherent property of the image.
+
+However, some closed-loop deployment technologies such as [Argo CD](https://argo-cd.readthedocs.io/) verify that the created resources correspond to the requested resources, essentially synchronizing the state from a reference repository.
+In consequence, a mutated image reference causes an error for such tools, as actual and requested resource differ.
+To resolve this, it is possible to configure Connaisseur to only validate but not mutate image references by which the original image reference (tag or digest) remains unchanged and successful admission only indicates that a trusted digest exists.
+However, image tag and digest are only loosly associated which introduces a [*time-of-check to time-of-use* vulnerability](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use): an attacker slips in a malicious image after Connaisseur validated that a given tag exhibits a signed digest but before the runtime resolves the tag for pulling the image.
+This supposedly small time window might be significantly larger if images are re-pulled by the container runtime at a later time for some reason.
+
+To disable image mutation, set the `validationMode.mutateImage` flag to `false` in `helm/values.yaml`.
+
+## Configuration options
+
+`validationMode` in `helm/values.yaml` supports the following keys:
+
+| Key | Default | Required | Description |
+| - | - | - | - |
+| `mutateImage` | true | | `true` or `false`; mutate image reference to trusted digest (`true`) or keep original reference (`false`) |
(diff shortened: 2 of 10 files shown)


phbelitz (approved):
lgtm

connaisseur/flask_application.py
```diff
@@ -24,6 +24,7 @@
 """
 CONFIG = Config()
 DETECTION_MODE = os.environ.get("DETECTION_MODE", "0") == "1"
+VALIDATION_MODE_MUTATE = os.environ.get("VALIDATION_MODE_MUTATE", "true") == "true"
```
Starkteetje, line 27: This falls back silently to other mode if config is different, i.e. if "True". As long as we don't validate that value is `true` or `false`, I'd rather have the fallback in favour of the default, i.e.
```suggestion
VALIDATION_MODE_MUTATE = os.environ.get("VALIDATION_MODE_MUTATE", "true") != "false"
```

helm/values.yaml
```diff
@@ -142,6 +142,10 @@ policy:
 # interrupting operation.
 detectionMode: false
 
+# validation mode allows advanced configuration of image validation such as whether to mutate image references
+validationMode:
+  mutateImage: true  # mutate image references to signed digests ('true', default) or keep original reference ('false')
```
Starkteetje, line 147: Would add a disclaimer that this actually lowers security (significantly I would argue, since the threat model for Connaisseur basically is that the attacker has control over the registry or connection to registry).

Starkteetje (commented):
This does not actually fix #270 as 

> but fail if the a reference would require mutation.

is not actually what happens here. This lowers the security pretty significantly.

helm/values.yaml
xopham: there is a disclaimer in the docs, you think it must be in the helm as well?

Starkteetje: It basically undermines every security guarantee that Conny provides. So yes. I'm also not sure whether it's something we should allow in the first place, because it so significantly undermines it, but if so, there should be as disclaimers at every point this is mentioned

Starkteetje: Thinking about it, I would prefix the config var with `unsafe`

Starkteetje: like `unsafeOnlyValidateTags: false`

connaisseur/flask_application.py
xopham: ah...good catch. Will change

xopham:
@Starkteetje it does fail if there is no trusted digest:
```bash
$ kubectl run unsigned --image=docker.io/securesystemsengineering/testimage:unsigned
Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: Unable to find signed digest for image docker.io/securesystemsengineering/testimage:unsigned.
```
You are correct in that it does not fix #270 or at least a requirement therein. Will highlight more strictly in the docs that it may be better to stick to mutation and use digests. Otherwise, I see no way to solve #712

helm/values.yaml
xopham: I disagree in that it undermines EVERY security guarantee. It significantly complicates a practical attack.
Moreover, `detectionMode` is in some ways worse than that as it gives a warning but that may never be read and we do not denote it as `unsafe`

xopham: also `namespacedValidation` can have the same effect

Starkteetje: detectionMode on the other hand has a very clear interface. It "detects" but doesn't do more. It has very clear applications for testing and thus can improve security by enabling to get the tool into a pre-existing infrastructure without breakage.

validationMode on the other hand doesn't have a super clear-cut interface. It only validates, but somewhat wants to provide security. The distinction is not apparent from naming alone that it will validate existence of a digest, but make no pull guarantees. While it does enable some infrastructures to validate digests, it doesn't give them the actual guarantees that one would assume to get from an admission controller such as Conny

If you feel strongly about not adding the `unsafe` prefix, how about something like

```
security-enforcement (kinda bad name, let's pick another):
  - manifest-security: true # whether to enforce that the manifest corresponds to a validly signed (not signed, because could be allowlisted but you know i mean) (basically this = false equivalent to detectionMode -> unify with next major)
  - runtime-security: true # whether to enforce the pulled image is actually the signed (not signed, because could be allowlisted but you know i mean) image and not another one (this = false is your validation only mode)
```
(if you find the general idea reasonable, let's sit a few minutes for proper names)

docs/features/validation_mode.md
```diff
@@ -0,0 +1,37 @@
...
+> :warning: Be aware that disabling image reference mutation has significant impact on the security guarantees of signature verification.
+> Carefully consider the notes below.
+
+Validation mode allows configuration whether Connaisseur mutates image references (tags to trusted digests).
+
+In default behavior during validation, Connaisseur uses the image reference (tag or digest) from the admission request to identify a trusted digest (digest with valid signature from a configured trust root) and modifies the reference to the trusted digest if one exists or denies the request otherwise.
+This ensures that the container runtime pulls and spins up a container of a validated image, as the digest is an immutable inherent property of the image.
+
+However, some closed-loop deployment technologies such as [Argo CD](https://argo-cd.readthedocs.io/) verify that the created resources correspond to the requested resources, essentially synchronizing the state from a reference repository.
+In consequence, a mutated image reference causes an error for such tools, as actual and requested resource differ.
```
Starkteetje, line 12:
```suggestion
In consequence, a mutated image reference causes an error for such tools, as actual and requested resource references differ.
```
?

docs/features/validation_mode.md
```diff
@@ -0,0 +1,37 @@
...
+To resolve this, it is possible to configure Connaisseur to only validate but not mutate image references by which the original image reference (tag or digest) remains unchanged and successful admission only indicates that a trusted digest exists.
+However, image tag and digest are only loosly associated which introduces a [*time-of-check to time-of-use* vulnerability](https://en.wikipedia.org/wiki/Time-of-check_to_time-of-use): an attacker slips in a malicious image after Connaisseur validated that a given tag exhibits a signed digest but before the runtime resolves the tag for pulling the image.
+This supposedly small time window might be significantly larger if images are re-pulled by the container runtime at a later time for some reason.
```
Starkteetje, line 15:
```suggestion
This supposedly small time window might be significantly larger if images are re-pulled by the container runtime at a later time for some reason or if the attacker completely controls the image registry.
```

phbelitz:
already implemented in golang version

PR #763, opened 2022-08-22
fix: handle image references with both a tag and digest present

### LINKED ISSUE #749: Incorrectly parsed image references if they contain both a tag and digest
**Describe the bug**

I use image references in my manifests that contain both a tag and a sha256 digest, e.g.: `example.com/library/image:v1.2.3@sha256:abcd...`. This causes the admission webhook to fail with:

```console
Error: INSTALLATION FAILED: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: Unexpected Cosign exception for image "example.com/library/image:v1.2.3@sha256:1d076650cbf2f364235e49203a98015d35443d72e74f462085ea3bcedd25cb4c:v2.15.5": Error: parsing reference: could not parse reference: example.com/library/image:v1.2.3@sha256:1d076650cbf2f364235e49203a98015d35443d72e74f462085ea3bcedd25cb4c:v2.15.5
main.go:62: error during command execution: parsing reference: could not parse reference: example.com/library/image:v1.2.3@sha256:1d076650cbf2f364235e49203a98015d35443d72e74f462085ea3bcedd25cb4c:v2.15.5
```

So it appears that these images are incorrectly parsed. I believe this is caused in https://github.com/sse-secure-systems/connaisseur/blob/v2.6.2/connaisseur/image.py#L30-L65 as adding the following new test data to https://github.com/sse-secure-systems/connaisseur/blob/v2.6.2/tests/test_image.py#L115 causes the tests to fail:

```python
        (
            "docker.io/library/image:v1.2.3@sha256:f8816ada742348e1adfcec5c2a180b675bf6e4a294e0feb68bd70179451e1242",
            "image",
            "v1.2.3",
            "f8816ada742348e1adfcec5c2a180b675bf6e4a294e0feb68bd70179451e1242",
            "library",
            "docker.io",
            fix.no_exc(),
        ),
```

https://github.com/sse-secure-systems/connaisseur/blob/v2.6.2/connaisseur/image.py#L84 has to be adapted to either return the full tag+digest or just the digest as it is passed to cosign verify later on (https://github.com/sse-secure-systems/connaisseur/blob/v2.6.2/connaisseur/validators/cosign/cosign_validator.py#L304).

Cosign itself does support both tags+digest when verifying, but drops the tag if both a tag+digest are specified e.g.

```sh
$ COSIGN_EXPERIMENTAL=1 cosign verify ghcr.io/chgl/fhir-server-exporter:v2.0.2@sha256:0d42bd5680994b2f3e376ef1159ec89ede6b1d59892fcd0f8079672304f99f7f

Verification for ghcr.io/chgl/fhir-server-exporter@sha256:0d42bd5680994b2f3e376ef1159ec89ede6b1d59892fcd0f8079672304f99f7f --
...
```

**Expected behavior**

It should be possible to verify image references with both a tag and a digest specified. If both are present, from an implementation point of view it's probably easiest to drop the tag and just use the digest when passing the image reference.

**Optional: To reproduce**

1. Install connaisseur
2. try deploying a manifest containing an image reference with both a tag and digest
3. fails with the above error message as the image being passed to cosign is incorrectly constructed

 - OS: <OS and version>
 - Kubernetes Cluster: <cluster type and version>
 - Notary Server: <notary and version>
 - Container registry: <container registry and version>
 - Connaisseur: <connaisseur version>
 - Other: <other relevant tools and versions>

As a tiny side note, in https://github.com/sse-secure-systems/connaisseur/blob/v2.6.2/connaisseur/image.py#L84 hard-coding the digest to sha256 could possibly be an issue some time in the future as technically various types of digests are possible, see https://github.com/opencontainers/image-spec/blob/main/descriptor.md#digests - although admittedly I've never seen anything but sha256 in the wild.

chgl:
Fixes #749

## Description

It is possible for image references to contain both a tag and a digest at the same time. Previously, this was not handled causing the validators to fail.

I'm not sure if what I did is the most idiomatic or even elegant Python, but the changes are at least fairly small.

I've built and pushed the image as `ghcr.io/chgl/connaisseur:fix-image-tag-and-digest` and testet it using the helm chart by setting

```yaml
deployment:
  image: ghcr.io/chgl/connaisseur:fix-image-tag-and-digest
```

The images that were previously causing issues were now correctly validated and accepted.

## Checklist


chgl:
Draft as long as there is no consensus on whether or not this is actually the right approach in https://github.com/sse-secure-systems/connaisseur/issues/749

phbelitz:
@richgerrard with this the tag is preserved and should be visible in the k8s pod description. additionally, what might be possible to do is add the tag and digest as annotations to the mutated resource. that might be more convenient to get information about the resource, without the need to do additional parsing. is that something you deem as useful?

connaisseur/image.py
```diff
@@ -27,61 +27,63 @@ class Image:
...
-
-        name_tag = image.split("/")[-1]
-        search = re.search(tag_re, name_tag)
-        self.digest, self.tag = search.groups() if search else (None, "latest")
-        self.name = name_tag.removesuffix(":" + str(self.tag)).removesuffix(
-            "@sha256:" + str(self.digest)
+    def __init__(self, image: str):  # pylint: disable=too-many-locals
+
+        # implements https://github.com/distribution/distribution/blob/main/reference/regexp.go
+        digest_hex = r"[0-9a-fA-F]{32,}"
```
Starkteetje, line 33: That's for sha256 and not generally digests.

From [spec](https://docs.docker.com/registry/spec/api/#content-digests) this should be more general. Also algorithm and separator do not seem to match the spec

connaisseur/image.py
```diff
@@ -27,61 +27,63 @@ class Image:
...
-        if not match:
-            msg = "{image} is not a valid image reference."
-            raise InvalidImageFormatError(message=msg, image=image)
-
-        name_tag = image.split("/")[-1]
-        search = re.search(tag_re, name_tag)
-        self.digest, self.tag = search.groups() if search else (None, "latest")
-        self.name = name_tag.removesuffix(":" + str(self.tag)).removesuffix(
-            "@sha256:" + str(self.digest)
+    def __init__(self, image: str):  # pylint: disable=too-many-locals
```
Starkteetje, line 30: This seems like a larger rework. Would add tests for the changed/expected behaviour. E.g. afai can see, you added the length check but never tested it

connaisseur/image.py
```diff
@@ -27,61 +27,63 @@ class Image:
...
+        self.name = components[-1]
+        registry_repo = components[:-1]
+
+        try:
+            registry = registry_repo[0]
+            self.registry = (
+                registry
+                if re.search(r"[.:]", registry)
+                or registry == "localhost"
+                or any(ele.isupper() for ele in registry)
```
Starkteetje, line 72: What is the reason here?

connaisseur/image.py
```diff
@@ -27,61 +27,63 @@ class Image:
...
 
     def __str__(self):
-        repo_reg = "".join(
-            f"{item}/" for item in [self.registry, self.repository] if item
-        )
-        tag = f":{self.tag}" if not self.digest else f"@sha256:{self.digest}"
-        return f"{repo_reg}{self.name}{tag}"
+        repo_reg = "/".join(item for item in [self.registry, self.repository] if item)
+        tag = f":{self.tag}" if self.tag else ""
+        digest = f"@sha256:{self.digest}" if self.digest else ""
```
Starkteetje, line 85: This again doesn't work if there were a different hash algo

connaisseur/image.py
Starkteetje: Also `docker pull securesystemsengineering/testimage@sha256:2d2c82da46687f6f390389c28f64ad00a21a077e6ded496e2e37e6a15b42682E` gives me `invalid checksum digest format` in comparison to lower-case `e` at the end. Not sure why, whether I looked at the wrong spec or docker does more validation that should be

In the latter case, not sure what to do. Maybe log with info?

tests/validators/notaryv1/test_notaryv1_validator.py
```diff
@@ -85,6 +85,20 @@ def test_init(m_notary, val_config):
...
         ),
+        (
+            "securesystemsengineering/alice-image:test@sha:ac904c9b191d14faf54b7952f2650a4bb21c201bf34131388b851e8ce992a652",
+            None,
+            [],
+            "ac904c9b191d14faf54b7952f2650a4bb21c201bf34131388b851e8ce992a652",
+            fix.no_exc(),
+        ),
+        (
+            "securesystemsengineering/alice-image:test@sha:13333333333333333333333333333337",
```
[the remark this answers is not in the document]
phbelitz: ok, I combined the `__search_image_targets_for_tag` and `search_image_targets_for_digest` into one and added a check, whther the searched for digest and actual are the same.

connaisseur/image.py
phbelitz: ok, welp. we pretty much only support `sha256` at this point. it's hardcoded at multiple places in the code and refactoring those would XPLODE the whole PR. so if it's fine with you, i would leave things as the are right now.

connaisseur/image.py
phbelitz: i added tests for the to long name and non supported digest algo.

connaisseur/image.py
phbelitz: that's remnants from a spec we previously implemented. i just kept it that way.

connaisseur/image.py
phbelitz: see above. let's stay with `sha256` for now.

connaisseur/image.py
Starkteetje: If you open an issue I'm fine with postponing that.
What about the other issue with format?

connaisseur/image.py
```diff
@@ -26,62 +26,71 @@ class Image:
...
-        return self.digest is not None
+        name, tag, digest = match.groups()
+        components = name.split("/")
+        self.name = components[-1]
+        self.digest_algo, self.digest = digest.split(":") if digest else (None, None)
+        self.tag = tag or ("latest" if not self.digest else None)
+
+        if self.digest_algo and self.digest_algo != "sha256":
+            raise InvalidImageFormatError(
+                message="A digest algo of {digest_algo} is not supported. Use sha256 instead.",
```
Starkteetje, line 68:
```suggestion
                message="A digest algorithm of {digest_algo} is not supported. Use sha256 instead.",
```

connaisseur/image.py
```diff
@@ -26,62 +26,71 @@ class Image:
...
-        Return `True` if the image has a digest, `False` otherwise.
-        """
-        return self.digest is not None
+        name, tag, digest = match.groups()
+        components = name.split("/")
+        self.name = components[-1]
+        self.digest_algo, self.digest = digest.split(":") if digest else (None, None)
+        self.tag = tag or ("latest" if not self.digest else None)
+
+        if self.digest_algo and self.digest_algo != "sha256":
```
Starkteetje, line 66: Since its used multiple times by now, would add as constant

connaisseur/image.py
phbelitz: Created #843. For the format ... I don't konw, I think I'd rather implement the OCI standard (roughly) then try to comply with funky things docker is doing. Agree?

Starkteetje (approved):

PR #812, opened 2022-11-03
Fix/fix issues

phbelitz:
There is a transitiv dependency on frozenlist, which updated their package for python 3.11, but didn't add the package to pypi wheels. Thus, building on python >= 3.11 fails the whole docker building process ([issue](https://github.com/aio-libs/frozenlist/issues/342)). Switching to python 3.10. Secondly, there are two more vulnerabilities due to cosign, which cannot be fixed on our side. Ignore listing them until cosign publishes a new version.

## Checklist


docker/Dockerfile
```diff
@@ -1,13 +1,17 @@
...
+FROM python:alpine as base
 
 # Build dependencies
 FROM base as builder
 
 RUN mkdir /install
 WORKDIR /install
 COPY requirements.txt /requirements.txt
+
+# Have to upgrade pip and install extra packages due to error when installing frozen list
```
Starkteetje, line 10:
```suggestion
# Have to upgrade pip and install extra packages due to error when installing frozenlist
```

Starkteetje (approved):

PR #814, opened 2022-11-04
Log image name consistently

### LINKED ISSUE #811: Error logs do not contain image information, only pod name
**Describe the feature**
Error logs contain the image that was denied, not just the pod name/id

It should be.  images > pod ids in this context.

Here's the idea, instead of:
```ERROR: {'message': 'Static deny.', 'context': {'detection_mode': True, 'user': 'admin', 'operation': 'CREATE', 'kind': 'Pod', 'name': 'nsenter-n6exa6', 'namespace': 'default'}}```
We try this:
```ERROR: {'message': 'Static deny.', 'context': {'detection_mode': True, 'user': 'admin', 'operation': 'CREATE', 'kind': 'Pod', 'name': 'nsenter-n6exa6', 'namespace': 'default', 'image': 'relevantinformation:tag@sha:blahblah}}```

N/A

### PR DESCRIPTION
**Pull request by Starkteetje**
Fixes #811 

## Checklist

### CODE REVIEW

File: connaisseur/validators/cosign/cosign_validator.py [hunk 1]
```diff
@@ -48,36 +49,43 @@ def __init__(
...
+            # await results (output dropped as `vals` is updated within function)
+            for future in futures:
+                vals.update(future.result())
+
+            return CosignValidator.__apply_policy(
+                vals=vals, threshold=threshold, required=required
+            )
+        # enrich all errors with some contextual information
+        except BaseConnaisseurException as err:
+            err.update_context(**self.generate_logging_context(image, **kwargs))
```
phbelitz, line 87: https://github.com/sse-secure-systems/connaisseur/blob/48088e78498c526be6254fef896c6e2b6c0af58f/connaisseur/flask_application.py#L186-L188

doesn't it make sens to add this in `flask_application.py` instead of the individual validators? then `static_validator.py` just needs to add the image in its error message.

Starkteetje: I was thinking along the lines of this enabling tackling the issue for better logging on each validator, but you're right that this is something that can be handled at the higher level. I'll change it

connaisseur/flask_application.py
```diff
@@ -138,6 +136,7 @@ async def __admit(admission_request: AdmissionRequest):
     try:
         await patches
     except BaseConnaisseurException as err:
+        logging_context = dict(admission_request.context)
```
phbelitz, line 139: :pray:

phbelitz (approved):

PR #800, opened 2022-10-18
Don't clear image tag

richgerrard:
Just because you've augmented an image with a new property ("digest") doesn't mean you get to clear a different value ("tag") in the setter.

Fixes: N/A (or, more correctly: future pain when tag preservation in patch is implemented)

## Description

Image tags are valid properties of the image.  You shouldn't arbitrarily throw the tag away once you've set a digest.
I intend for you to preserve this image tag in the patch based on a config setting in a subsequent PR, but first I need to test your willingness to adapt here with this change.

I can't keep using your solution if we can't take these steps together to preserve my image tags.

## Checklist

- [X] PR is rebased to/aimed at branch `develop`
- [X] PR follows [Contributing Guide](https://github.com/sse-secure-systems/connaisseur/blob/master/docs/CONTRIBUTING.md)
- [X] Added tests (if necessary)
- [X] Extended README/Documentation (if necessary)
- [ ] Adjusted versions of image and Helm chart in `values.yaml` and `Chart.yaml` (if necessary)

PR #833, opened 2022-11-28
Fix: add tag labels to patch output

richgerrard:
Fixes # n/a
Extends: #800

## Description

The mutating controller is stripping the human-readable element from the patch, and therefore the deployment artifacts that are mutated.
This results in pain and frustration for humans and tools that are relying on the human-readable element.  This includes `kubectl` output, metrics tools like `grafana` dashboards that have suddenly lost tag elements, security monitoring tools that can no longer see the deployed version by parsing a tag, and many more.
Essentially, we will have to select a different tool and move away from `connaisseur` if the image tags can not be preserved.
`Chainguard` allegedly handles tags the way that I desire.

## Checklist

- [X] Added tests (if necessary)

### DIFF
--- a/connaisseur/image.py
+++ b/connaisseur/image.py
@@ -80,7 +80,7 @@ class Image:
         repo_reg = "".join(
             f"{item}/" for item in [self.registry, self.repository] if item
         )
-        tag = f":{self.tag}" if not self.digest else f":{self.tag}@sha256:{self.digest}"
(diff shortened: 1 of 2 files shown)


aravandi (approved):

phbelitz:
Closing in favor of https://github.com/sse-secure-systems/connaisseur/pull/763
May add you as co-author if desired.

PR #840, opened 2022-12-16
Feat: Allow configuration of Connaisseur TLS certificate

### LINKED ISSUE #225: Make Connaisseur TLS certificate configurable
**Describe the feature**
Currently, Connaisseur generates a self-signed certificate for communication. This should be made configurable via `helm/values.yaml` to e.g. provide a cert or reference a corresponding secret.

There are situations in which the self-signed certificate may not work and a cert might have to be provided externally. Also, how would Connaisseur work with solutions such as istio and 2-way TLS?

There should be a configuration option in `helm/values.yaml` under `deployment`. That might allow several configurations:
- configure certificate or use self-signed (probably use self-signed if not provided otherwise)
- should cert be used at all (considering the istio case) and further configuration options
- configure cert directly or via secret (probably the latter would be best)

- see also #224 for more information.
- make sure to also check implications for #181 related issues

Starkteetje:
This commit allows for configuration of the TLS paramters used by Connaisseur when being called on one of its endpoints. This allows configuring both publicly trusted or self-signed certificates.

Fixes #225

## Open Questions
- Configures private key in yaml file. That's meh. We can't really use a scoped secret as the namespace doesn't exist prior to installation. Reading secret from different namespace sounds dumber than config secret. Can we work around this?
- I implemented it the way that the config doesn't override prior installations. This would've been valid for current installations as the self-signed cert is valid for 100years. However, the new ones may even be publicly trusted certs, which then have a smaller validity. Thus the question of rotating the certificate comes up. How do we handle this?

## Checklist

- [x] Extended README/Documentation


helm/templates/certificate_webhook-conf.yaml
```diff
@@ -1,10 +1,32 @@
...
+{{- if hasKey .Values.deployment "tls" -}}
+  {{- if and (not (hasKey .Values.deployment.tls "cert")) (hasKey .Values.deployment.tls "key")}}
+    {{ fail "Helm configuration has a 'deployment.tls' section with a 'key' attribute, but is missing the 'cert' attribute." -}}
+  {{- end -}}
+  {{- if and (not (hasKey .Values.deployment.tls "key")) (hasKey .Values.deployment.tls "cert")}}
+    {{ fail "Helm configuration has a 'deployment.tls' section with a 'cert' attribute, but is missing the 'key' attribute." -}}
+  {{- end -}}
+{{- end -}}
+# If present, use the configured certificate instead
+{{- if hasKey .Values.deployment "tls" -}}
```
phbelitz, line 19: second check check for visibility? if not you can pull it in the condition clause above

helm/templates/certificate_webhook-conf.yaml
```diff
@@ -1,10 +1,32 @@
...
+  # Helm 'and' executes both even if first fails, so we cannot use it to check both conditions at the same time without failing
+  {{- if hasKey .Values.deployment.tls "key" -}}
+    {{- $certByHelmConfig = buildCustomCert (.Values.deployment.tls.cert | b64enc) (.Values.deployment.tls.key | b64enc) -}}
+  {{- end -}}
+{{- end -}}
+
+# For whatever reason Helm concatenates the comments with the apiVersion field below
+# unless there's a comment below the Helm commands or there's no comments at all.
+# As I wanted to have comments, here's this comment commenting whis this comment is here.
```
phbelitz, line 28:
```suggestion
# As I wanted to have comments, here's this comment commenting this comment is here.
```

phbelitz:
Helm also offers [reading files](https://helm.sh/docs/chart_template_guide/accessing_files/), so we could make the file name configurable in the yaml file and then look on disk for the name. But I think helm restricts access to files that need to be withing the chart itself (so nothing which is above the directory with the `Chart.yaml`).

Starkteetje:
> Helm also offers [reading files](https://helm.sh/docs/chart_template_guide/accessing_files/), so we could make the file name configurable in the yaml file and then look on disk for the name. But I think helm restricts access to files that need to be withing the chart itself (so nothing which is above the directory with the `Chart.yaml`).

So basically, "Put a file with the creds into the directory when calling make/helm install"? I think I like it better than my alternatives. I mean, it's somewhat cumbersome, but then again I don't really think this will be used super often and if so, it handles TLS keys, so we should take care

Starkteetje:
We talked about this again. For rotation, if we simply reference a file name, the config doesn't change from Helm's PoV, so an upgrade won't do anything. As such, we said for now we'll stick with secrets in the helm.yaml (the way of preparing a Connaisseur namespace and pre-provisioning secrets in it is always possible anyways if one wants to go done that route) and may think about a cert-manager integration or the like in the long run.

In the same vein, we need to make sure that changing the state of the tls variables actually "changes" the config, such that helm knows it needs updates

tests/integration/cases.yaml
```diff
@@ -273,3 +273,18 @@ test_cases:
     namespace: default
     expected_msg: pod/pod-poff created
     expected_result: null
+  certificate:
+  - id: cu
```
phbelitz, line 277: that's the same id as cosign unsigned test case. let's use somethign different like `cert-u`. same below for the `cs` id -> `cert-s`

tests/integration/integration-test.sh
```diff
@@ -404,6 +427,17 @@ case $1 in
   make_install
   load_test
   ;;
+"configured-cert")
+  echo "Testing deployment of Connaisseur using a pre-configured TLS certificate. See issue https://github.com/sse-secure-systems/connaisseur/issues/225"
+  update_via_env_vars
+  make install
+  certificate_int_test
+  kubectl delete pod pod-cs # Clean up such that next test doesn't run into existing pod
```
phbelitz, line 435: needs to change, should the id of the testcase be change

tests/integration/cases.yaml
```diff
@@ -284,3 +284,18 @@ test_cases:
     namespace: default
     expected_msg: pod/pod-poff created
     expected_result: null
+  certificate:
+  - id: x509u
```
phbelitz, line 288: sexy

phbelitz (approved):

Issue #516, opened 2022-02-04
Test for readiness of workload objects


Starkteetje:
**Describe the feature**
https://github.com/sse-secure-systems/connaisseur/pull/514 fixes a bug we had with parent objects not being accessible due to wrong API call. The integration test worked though, since it allowed the parent object when children were signed, though the children didn't work and rejected missing/invalid signatures. Hence, we didn't have a security problem, but we prevented valid deployments from working. The workload tests should cover this case and fail in case we prevent a WO from becoming ready.

**Optional: Implementation ideas**
Since it's probably ugly to check for in the integration tests, implementing it in only the WO test should suffice

PR #868, opened 2023-01-23
test: Test workload readiness

Starkteetje:
This MR introduces checks that not only are workloads in our workload tests deployed, but that they actually become ready at some point. This is related to https://github.com/sse-secure-systems/connaisseur/pull/514, which reports that we successfully deployed a parent resource, which never became ready as its child resources were invalid. Introducing a readiness check ensures that any resources we check are actually functional, in turn preventing regression alike to https://github.com/sse-secure-systems/connaisseur/pull/514.

Fix https://github.com/sse-secure-systems/connaisseur/issues/516

See https://github.com/sse-secure-systems/connaisseur/actions/runs/3988111613 for a run of the pipeline, afterwards only  removed the commit that enabled these checks on dev so that our default testing time isn't blown out of proportion.

## Checklist


xopham (commented):
one minor issue

xopham (approved):
lgtm

PR #869, opened 2023-01-26
feat: auto update approval

### LINKED ISSUE #820: Connaisseur prevents downscaling of a previous replicaset during an update of a deployment.
**Describe the bug**
Connaisseur prevents downscaling of a previous replicaset during an update of a deployment.

We run into a scenario, where after a deployment update, we have two active replicasets running at the same time.

**Expected behavior**
It is expected that older replicaset is downscaled when the new replicaset is ready.

Log message 
ERROR: {'message': 'Static deny.', 'context': {'detection_mode': False, 'user': 'system:serviceaccount:kube-system:deployment-controller', 'operation': 'UPDATE', 'kind': 'ReplicaSet', 'name': 'p1o-fe-cms-55957f46d4', 'namespace': 'o5'}}

Configuration 
  validators:
    # static validator that allows each image
    - name: allow
      type: static
      approve: true
    - name: deny
      type: static
      approve: false
    - name: p1o
      type: cosign
      trust_roots:
        - name: p1o-pipeline
          key: |
            -----BEGIN PUBLIC KEY-----
            -----END PUBLIC KEY-----
      auth:
        secret_name: harbor-registry-secret
  policy:
    - pattern: "*:*"
      validator: deny
    - pattern: "docker.xxx.xx/p1o/p1o-fe:*"
      validator: p1o
      with:
        trust_root: p1o-pipeline

phbelitz:
Added a feature flag that enables auto update approval. When active, Connaisseur automatically allows updates to resources in which no changes to image references were made.

fixes  #820

## Checklist

### DIFF
--- a/connaisseur/flask_application.py
+++ b/connaisseur/flask_application.py
@@ -153,6 +153,17 @@ async def __validate_image(type_index, image, admission_request):
     type_, index = type_index
     logging_context.update(image=original_image)
 
+    # if automatic_unchanged_approval is enabled, admit resource updates
+    # if they do not alter the resource's image reference(s)
+    # https://github.com/sse-secure-systems/connaisseur/issues/820
+    unchanged_approval_on = os.environ.get("AUTOMATIC_UNCHANGED_APPROVAL", "0") == "1"
+    if unchanged_approval_on and admission_request.operation.upper() == "UPDATE":
+        old_images = admission_request.old_wl_object.containers.values()
+        if image in old_images:
+            msg = f'automatic approval for unchanged image "{original_image}".'
+            logging.info(__create_logging_msg(msg, **logging_context))
+            return
+
     # child resources have mutated image names, as their parents got mutated
     # before their creation. this may result in mismatch of rules or duplicate
     # lookups for already approved images. so child resources are automatically
--- a/connaisseur/workload_object.py
+++ b/connaisseur/workload_object.py
@@ -1,3 +1,5 @@
+from functools import cached_property
+
 import connaisseur.kube_api as k_api
 from connaisseur.exceptions import ParentNotFoundError, UnknownAPIVersionError
 from connaisseur.image import Image
@@ -48,7 +50,7 @@ class WorkloadObject:
                 wl_obj_name=self.name,
             )
 
-    @property
+    @cached_property
     def parent_containers(self):
         parent_containers = {}
         for owner in self._owner:
@@ -91,7 +93,7 @@ class WorkloadObject:
     def spec(self):
         return self._spec["template"]["spec"]
 
-    @property
+    @cached_property
     def containers(self):
         return {
             (container_type, index): Image(container["image"])
--- a/docs/features/automatic_update_approval.md
+++ b/docs/features/automatic_update_approval.md
@@ -1,0 +1,34 @@
+# Automatic Unchanged Approval
+
+With the feature *automatic unchanged approval* enabled, Connaisseur automatically approves any resource that is updated and doesn't change its image references.
+This is especially useful when handling long living resources, with potentially out-of-sync signature data, that still need to be scaled up and down.
+
+An example: When dealing with a deployment that has an image reference `image:tag`, this reference is updated by Connaisseur during signature validation to `image@sha256:123...`, to ensure the correct image is used by the deployment.
+When scaling up or down the deployment, the image reference `image@sha256:123...` is presented to Connaisseur, due to the updated definition.
+Over time the signature of the original `image:tag` may change and a new "correct" image is available at `image@sha256:456...`.
+If afterwards the deployment in scaled up or down, Connaisseur will try to validate the image reference `image@sha256:123...`, by looking for it inside the signature data it receives.
+Unfortunately this reference may no longer be present, due to signature updates and thus the whole scaling will be denied.
+
+With automatic unchanged approval enabled, this is no longer the case.
+The validation of `image@sha256:123...` will be skipped, as no different image is used.
+
+This obviously has security implications, since it's no longer guaranteed that resources that are updated, have fresh and up-to-date signatures.
+So use it with caution.
+For that reason the feature is also disabled by default.
+The creation of resources on the other hand remains unchanged and will enforce validation.
+
+## Configuration options
+
+`automaticUnchangedApproval` in `helm/value.yaml` supports the following keys:
+
+| Key | Default | Required | Description |
+| - | - | - | - |
+| `automaticUnchangedApproval` | false |  | `true` or `false` ; when `true`, Connaisseur will enable automatic unchanged approval |
+
+## Example
+
+In `helm/values.yaml`:
+
+```yaml
+automaticUnchangedApproval: true
+```
--- a/helm/templates/env.yaml
+++ b/helm/templates/env.yaml
@@ -27,6 +27,9 @@ data:
   {{- if .Values.alerting }}
   CLUSTER_NAME: {{ default "not specified" .Values.alerting.cluster_identifier }}
   {{- end }}
+  {{- if .Values.automaticUnchangedApproval  }}
+  AUTOMATIC_UNCHANGED_APPROVAL: "1"
+  {{- end }}
   KUBE_VERSION: {{ .Capabilities.KubeVersion }}
 ---
 apiVersion: v1
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -1,7 +1,7 @@
 # configure Connaisseur deployment
 deployment:
   replicasCount: 3
-  image: securesystemsengineering/connaisseur:v2.7.0
+  image: securesystemsengineering/connaisseur:v2.8.0
   imagePullPolicy: IfNotPresent
   # imagePullSecrets contains an optional list of Kubernetes Secrets, in Connaisseur namespace,
   # that are needed to access the registry containing Connaisseur image.
@@ -175,6 +175,13 @@ namespacedValidation:
 # NOTE: configuration of automatic child approval is in EXPERIMENTAL state.
 automaticChildApproval:
   enabled: true
+
+# automatic unchanged approval determines how certain updates of Kubernetes resources are handled
+# by Connaisseur. when disabled (default), Connaisseur validates and mutates resources whenever they 
+# are created or updated. when enabled, Connaisseur will skip validation on any update request that
+# does not change the image references used by the resource. check the docs for more
+# information.
+automaticUnchangedApproval: false
 
 # debug: true
 
--- a/tests/data/sample_admission_requests/ad_request_auto_update_approval.json
+++ b/tests/data/sample_admission_requests/ad_request_auto_update_approval.json
@@ -1,0 +1,220 @@
+{
+    "kind": "AdmissionReview",
+    "apiVersion": "admission.k8s.io/v1",
+    "request": {
+        "uid": "6418f614-b7f1-4c57-94bf-14fc30b2d791",
+        "kind": {
+            "group": "apps",
+            "version": "v1",
+            "kind": "ReplicaSet"
+        },
+        "resource": {
+            "group": "apps",
+            "version": "v1",
+            "resource": "replicasets"
+        },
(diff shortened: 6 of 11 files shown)


connaisseur/flask_application.py
```diff
@@ -153,6 +153,17 @@ async def __validate_image(type_index, image, admission_request):
     type_, index = type_index
     logging_context.update(image=original_image)
 
+    # a resource that is being updated, without any image reference changes,
+    # can be automatically skipped, as image validation happened at some
+    # point prior
+    update_approval_on = os.environ.get("AUTOMATIC_UPDATE_APPROVAL", "0") == "1"
```
Starkteetje, line 159: Guess we didn't do it prior, but maybe it's time to namespace our env vars and make it `CONNAISSEUR_*`?

connaisseur/flask_application.py
```diff
@@ -153,6 +153,17 @@ async def __validate_image(type_index, image, admission_request):
     type_, index = type_index
     logging_context.update(image=original_image)
 
+    # a resource that is being updated, without any image reference changes,
+    # can be automatically skipped, as image validation happened at some
```
Starkteetje, line 157: That's not necessarily true. Deployment could be from before Connaisseur deployment. Would remove that reasoning and maybe replace with link or number of issue that pursuaded you to implement

connaisseur/flask_application.py
```diff
@@ -153,6 +153,17 @@ async def __validate_image(type_index, image, admission_request):
...
+    # point prior
+    update_approval_on = os.environ.get("AUTOMATIC_UPDATE_APPROVAL", "0") == "1"
+    if update_approval_on and admission_request.operation.upper() == "UPDATE":
+        old_images = admission_request.old_wl_object.containers.values()
+        if image in old_images:
+            msg = f'automatic update approval for "{original_image}".'
```
Starkteetje, line 163: If I'd read that log msg the auto update would trigger different (and wrong) associations in my mind. None of them make particular sense with Conny, but maybe we can find a better name for either the feature or at least the log msg

For log should be easy to be bit more verbose, for feature maybe sth like auto_approve_unchanged

connaisseur/workload_object.py
```diff
@@ -48,7 +50,7 @@ def __init__(self, request_object: dict, namespace: str):
                 wl_obj_name=self.name,
             )
 
-    @property
+    @cached_property
```
Starkteetje, line 53: Out of curiosity, how much does that save?

Starkteetje: And nit: should be its own commit to improve performance of other feature

helm/values.yaml
```diff
@@ -176,6 +176,11 @@ namespacedValidation:
 automaticChildApproval:
   enabled: true
 
+# when updating a resource, without changing the image reference (e.g. scaling up a deployment), the
```
Starkteetje, line 179:
```suggestion
# when updating a resource without changing the image reference (e.g. scaling up a deployment), the
```
I think(?)

Starkteetje, line 179: I think I'd make comment more similar to above and bit more straight-forward like `feature_name/config_name determines how resource updates that do not alter the image are handled during admission. If set to true, no validation will occur and the update will be admitted. If set to false (default), validation is performend, ensuring that the image has an up-to-date signature.

helm/values.yaml
```diff
@@ -176,6 +176,11 @@ namespacedValidation:
 automaticChildApproval:
```
Starkteetje, line 176: Unrelated note: This still experimental?

helm/values.yaml
```diff
@@ -176,6 +176,11 @@ namespacedValidation:
...
   enabled: true
 
+# when updating a resource, without changing the image reference (e.g. scaling up a deployment), the
+# validation process can be skipped, as the image was validated at some point prior. set to false,
```
Starkteetje, line 180: not necessarily true, see above

tests/data/sample_admission_requests/ad_request_auto_update_approval.json
```diff
@@ -0,0 +1,220 @@
...
+                        "labels": {
+                            "app": "test",
+                            "pod-template-hash": "58d59c69bf"
+                        }
+                    },
+                    "spec": {
+                        "containers": [
+                            {
+                                "name": "nginx",
+                                "image": "docker.io/library/redis:latest@sha256:325d5a448d8f6c1d30a0a0fb26090343279d4cf23258b26b1745862f332e9479",
```
Starkteetje, line 88: Is that an image that would normally fail validation? Would probably take sth non-existing to make sure this actually tests the feature's approval and not behind the scenes signature approval or test the same thing without monkeypatching env and test that it fails

docs/features/automatic_update_approval.md
```diff
@@ -0,0 +1,25 @@
+# Automatic Update Approval
+
+With the feature *automatic update approval* enabled, Connaisseur now may automatically approve any resource that is updated and doesn't change its image references. This is especially useful when handling long living resources, with potentially out of sync signature data, that still need to be scaled up and down.
```
Starkteetje, line 3:
```suggestion
With the feature *automatic update approval* enabled, Connaisseur automatically approves any resource that is updated and doesn't change its image references. This is especially useful when handling long living resources, with potentially out-of-sync signature data, that still need to be scaled up and down.
```

Starkteetje, line 3: Think we had policy of newline after each sentence for better updates (?) 🤷‍♂️

docs/features/automatic_update_approval.md
```diff
@@ -0,0 +1,25 @@
...
+
+An example: When dealing with a deployment that has an image reference `image:tag`, this reference is updated by Connaisseur during signature validation to `image@sha256:123...`, to ensure the correct and secure image is used by the deployment. When scaling up or down the deployment, the image reference `image@sha256:123...` is presented to Connaisseur, due to the updated. Now over the time the signature of the original `image:tag` may change and a new "secure" image is available at `image@sha256:456...`. If at that time the deployment in scaled up or down, Connaisseur will try to validate the image reference `image@sha256:123...`, by looking for it inside the signature data it receives. Unfortunately this reference may no longer be present, due to signature updates and thus the whole up or down scaling will be denied.
```
Starkteetje, line 5:
```suggestion
An example: When dealing with a deployment that has an image reference `image:tag`, this reference is updated by Connaisseur during signature validation to `image@sha256:123...`, to ensure the correct image is used by the deployment. When scaling up or down the deployment, the image reference `image@sha256:123...` is presented to Connaisseur, due to the updated definition. Over time the signature of the original `image:tag` may change and a new "correct" image is available at `image@sha256:456...`. If afterwards the deployment in scaled up or down, Connaisseur will try to validate the image reference `image@sha256:123...`, by looking for it inside the signature data it receives. Unfortunately this reference may no longer be present, due to signature updates and thus the whole scaling will be denied.
```

docs/features/automatic_update_approval.md
```diff
@@ -0,0 +1,25 @@
...
+
+With automatic update approval, this is no longer the case. The validation of `image@sha256:123...` will be skipped, as only the number of pods is changed and no different image is used.
```
Starkteetje, line 7:
```suggestion
With automatic update approval enabled, this is no longer the case. The validation of `image@sha256:123...` will be skipped, as no different image is used.
```

docs/features/automatic_update_approval.md
```diff
@@ -0,0 +1,25 @@
...
+
+This obviously has security implications, since it's no longer guaranteed that resources that are updated, have fresh and up-to-date signatures. So use it with caution.
```
Starkteetje, line 9: Would add we default to disable the feature

Starkteetje, line 9:
```suggestion
This obviously has security implications, since it's no longer guaranteed that resources that are updated, have fresh and up-to-date signatures. So use it with caution. The creation of resources on the other hand remains unchanged and will enforce validation.
```
or sth like that?

docs/features/automatic_update_approval.md
```diff
@@ -0,0 +1,25 @@
...
+| Key | Default | Required | Description |
+| - | - | - | - |
+| `automaticUpdateApproval` | false |  | `true` or `false` ; when `true`, Connaisseur will enable automatic update approval |
+
+## Example
+
+In `helm/values.yaml`:
+
+```yaml
+automaticUpdateApproval: true
```
Starkteetje, line 24: I think for v3, we should change these to be below a common thingy

admissionStrategies:
 - autoChild
 - autoUpdate

phbelitz: i thought about a unified `feature` field in which `detectionMode`, `namespacedValidation` and `autoChild` `autoUpdate` reside.

Starkteetje: Or sth like that 👍 
out of scope here anyway, but good to know you're feeling the same

connaisseur/workload_object.py
phbelitz: haven't done a test, but previously, if there where multiple containers defined, conny would have made a network request for the parent for each container ... so for the standard case barly any changes, but for handcraft one substantial i'd assume.

so i added it for each "expensive" property.

connaisseur/flask_application.py
phbelitz: will do in another commit, but will have to do that after you reviewed the chnages and I can do some rebasing and shuffeling of commits.

connaisseur/flask_application.py
phbelitz: did go for `automatic unchanged approval` to stay consistent with `automatic child approval` + changed log message.

connaisseur/workload_object.py
phbelitz: will do another commit after review + rebasing

helm/values.yaml
phbelitz: :shrug:  was it ever really experimental?

docs/features/automatic_update_approval.md
```diff
@@ -0,0 +1,34 @@
...
+For that reason the feature is also disabled by default.
+The creation of resources on the other hand remains unchanged and will enforce validation.
+
+## Configuration options
+
+`automaticUnchangedApproval` in `helm/value.yaml` supports the following keys:
+
+| Key | Default | Required | Description |
+| - | - | - | - |
+| `automaticUnchangedApproval` | false |  | `true` or `false` ; when `true`, Connaisseur will enable automatic update approval |
```
Starkteetje, line 26: still update approval

helm/templates/env.yaml
```diff
@@ -27,6 +27,9 @@ data:
   {{- if .Values.alerting }}
   CLUSTER_NAME: {{ default "not specified" .Values.alerting.cluster_identifier }}
   {{- end }}
+  {{- if .Values.automaticUnchangedApproval  }}
+  AUTOMATIC_UNCHNAGED_APPROVAL: "1"
```
Starkteetje, line 31:
```suggestion
  AUTOMATIC_UNCHANGED_APPROVAL: "1"
```

Starkteetje: nit: This would've been caught by an integration test ;)

connaisseur/flask_application.py
```diff
@@ -153,6 +153,15 @@ async def __validate_image(type_index, image, admission_request):
     type_, index = type_index
     logging_context.update(image=original_image)
 
+    # https://github.com/sse-secure-systems/connaisseur/issues/820
```
Starkteetje, line 156: I'd still add a bit of context, just the explanation of why and the justification of why it could be ok, shouldn't go here
```suggestion
    # if automatic_unchanged_approval is enabled, admit resource updates
    # if they do not alter the resource's image reference(s)
    # https://github.com/sse-secure-systems/connaisseur/issues/820
```

connaisseur/flask_application.py
phbelitz: will do in a different MR

Starkteetje (approved):

PR #873, opened 2023-01-30
Fix mutate timeout

### LINKED ISSUE #448: Adjust aiohttp timeouts to 30s
**Describe the bug**
Admission requests timeout after 30s. The http requests of connaisseur using aiohttp apply the default timeout of 300s: https://docs.aiohttp.org/en/stable/client_quickstart.html#timeouts
As a consequence, requests that fail due to badly reachable external resources are reported as internal error "failed calling webhook (..)" which is misleading as this is not an internal error reaching Connaisseur, but a timeout of Connaisseur on requesting external resources, e.g. a bad connection and should be reported as an error by Connaisseur as such.
Furthermore, the 300s timeout results in logs showing no information until 300s(-30s) after the request failed which complicates triangulation of such issues.

**Expected behavior**
http requests should timeout after 30s as there is no need to wait for a response later than that. A bit more difficult would be to enforce below 30s to ensure Connaisseur reports the error and not that the error is reported as inability to reach connaisseur.

Related to #447

Starkteetje:
Admission control is capped to 30s. As such Connaisseur must finish within these 30s or it will fail generically. To add information to this failure, this MR adds a 29s timeout to all validation activities and fails with a dedicated error message if the timeout is surpassed
The MR also contains minor fixes to docs and build I noticed when touching it

Fixes #448 

## Checklist

### DIFF
--- a/connaisseur/constants.py
+++ b/connaisseur/constants.py
@@ -1,1 +1,8 @@
+# We use aiohttp to connect to external sources. Webhooks time out after 30s,
+# so we need to respond before that or the Connaisseur webhook will fail with a
+# generic error message, see https://github.com/sse-secure-systems/connaisseur/issues/448
+# Since the timeouts in aiohttp are ceiled to full seconds
+# (https://docs.aiohttp.org/en/stable/client_quickstart.html#timeouts),
+# we can't get closer to 30 than 29 without failing to respond within 30 a lot more often
+AIO_TIMEOUT_SECONDS = 29
 SHA256 = "sha256"
--- a/connaisseur/flask_application.py
+++ b/connaisseur/flask_application.py
@@ -3,12 +3,13 @@ import logging
 import os
 import traceback
 
+import aiohttp
 from flask import Flask, jsonify, request
 from prometheus_flask_exporter import NO_PREFIX, PrometheusMetrics
 
 import connaisseur.constants as const
 from connaisseur.admission_request import AdmissionRequest
-from connaisseur.alert import send_alerts
+from connaisseur.alert import dispatch_alerts
 from connaisseur.config import Config
 from connaisseur.exceptions import (
     AlertSendingError,
@@ -64,27 +65,7 @@ def mutate():
     Handle the '/mutate' path and accept CREATE and UPDATE requests.
     Send a response back, which either denies or allows the request.
     """
-    admission_request = None
-    try:
-        logging.debug(request.json)
-        admission_request = AdmissionRequest(request.json)
-        response = asyncio.run(__admit(admission_request))
-    except Exception as err:
-        if isinstance(err, BaseConnaisseurException):
-            err_log = str(err)
-            msg = err.user_msg  # pylint: disable=no-member
-        else:
-            err_log = str(traceback.format_exc())
-            msg = "unknown error. please check the logs."
-        send_alerts(admission_request, False, msg)
-        logging.error(err_log)
-        uid = admission_request.uid if admission_request else ""
-        return jsonify(
-            get_admission_review(uid, False, msg=msg, detection_mode=DETECTION_MODE)
-        )
-
-    send_alerts(admission_request, True)
-    return jsonify(response)
+    return asyncio.run(__async_mutate())
 
 
 def metrics_label(response, label):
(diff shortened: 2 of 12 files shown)


tests/test_flask_application.py
```diff
@@ -220,31 +222,60 @@ async def test_admit(
...
-    "function, err",
+    "function, err, status",
     [
         (
             {
                 "target": "connaisseur.flask_application.send_alerts",
                 "side_effect": exc.AlertSendingError(""),
             },
             "Alert could not be sent. Check the logs for more details!",
+            500,
```
Starkteetje, line 246: Currently, if alerting fails and Connaisseur is configured to require alerts to be sent, Connaisseur fails. We could try to catch this and respond with an error that explains, we deny the image because of failed alert. Thoughts?

connaisseur/constants.py
```diff
@@ -1 +1,2 @@
+AIO_TIMEOUT_SECONDS = 29
```
xopham, line 1: think we could go closer to 30s? Ultimately, we'll return even if k8s timed out. It's just a bit unfortunate to loose a full second

connaisseur/flask_application.py
```diff
@@ -116,14 +98,62 @@ def readyz():
...
+    # thus we timeout slightly earlier
+    # https://github.com/sse-secure-systems/connaisseur/issues/448
+    timeout = aiohttp.ClientTimeout(total=const.AIO_TIMEOUT_SECONDS)
+    async with aiohttp.ClientSession(timeout=timeout) as session:
+        admission_request = None
+        try:
+            logging.debug(request.json)
+            admission_request = AdmissionRequest(request.json)
+            response = await __admit(admission_request, session)
+            # Depending on whether alerting must succeed, run it synchronously or in the background
```
xopham, line 114: wow...smooth

connaisseur/flask_application.py
```diff
@@ -116,14 +98,62 @@ def readyz():
...
+            if alerting_must_succeed(True):
```
xopham, line 115: I fail to see why you call with `True`

connaisseur/flask_application.py
```diff
@@ -116,14 +98,62 @@ def readyz():
...
+            return jsonify(response)
+        except Exception as err:
+            if isinstance(err, BaseConnaisseurException):
+                err_log = str(err)
+                msg = err.user_msg  # pylint: disable=no-member
+            elif isinstance(err, asyncio.TimeoutError):
+                err_log = str(traceback.format_exc())
+                msg = (
+                    "couldn't retrieve the necessary trust data for verification within 30s. "
+                    + "most likely there was a temporary network hiccup."
```
xopham, line 131: slang, provide more useful info like "bad connection" etc

connaisseur/validators/cosign/cosign_validator.py
```diff
@@ -1,4 +1,3 @@
-# import asyncio
```
xopham, line 1: I believe my fix was released and consequently, the asyncio parallelization might work now. So one might check, whether it's fast or less consuming. Bit hesitant to simply remove it

connaisseur/validators/notaryv1/notary.py
```diff
@@ -89,49 +89,58 @@ def healthy(self):
...
+                    message=msg, notary_name=self.name, tuf_role=str(role)
+                )
+
+            response.raise_for_status()
+            data = await response.text()
+            return TrustData(json.loads(data), str(role))
 
     async def get_delegation_trust_data(
-        self, image: Image, role: TUFRole, token: str = None
+        self,
```
xopham, line 136: this might be an old version of black

connaisseur/constants.py
Starkteetje: https://docs.aiohttp.org/en/stable/client_quickstart.html#timeouts

> aiohttp ceils timeout if the value is equal or greater than 5 seconds. The timeout expires at the next integer second greater than current_time + timeout.

Which means 29 will on average be 29,5s already and might be 29.9999 (in which case we'll probably not return in time. We could go down to 28 to fix that, but then again I agree with you on the point that we shouldn't "waste" too much time). If we went to 29,5, we'd basically always fail to return in time

connaisseur/validators/cosign/cosign_validator.py
Starkteetje: Well, either we want to have and use it or not. We shouldn't have random commented out code. If we want it back, we can simply git checkout an old version

connaisseur/flask_application.py
Starkteetje: `event_category = "admit_request" if admit_event else "reject_request"` in `alerting_must_succeed`. Could make it more specific and call it `alerting_must_succeed_if_admitted`, but didn't really see the benefit here of making it more specifc. Thoughts?

Starkteetje: Though I shall rename the function to must_... to reflect the bool answer

connaisseur/validators/notaryv1/notary.py
Starkteetje: It doesn't fail black validation, so I don't think so. Also it got bigger with another argument, so it makes sense it's no longer a single line

connaisseur/constants.py
xopham: good point. thanks for clarifying

connaisseur/validators/cosign/cosign_validator.py
xopham: well, ok, than it is gone. fine

connaisseur/validators/notaryv1/notary.py
xopham: ok, then let's keep it

connaisseur/constants.py
Starkteetje: Think I'll add a comment outlining this. Thanks for asking

connaisseur/flask_application.py
Starkteetje: Indeed, it is never called with False. Writing it to answer to any bool was easier (copy pasta from other parts), so I didn't feel the need to make the function less generic.

However, thinking about it, you're right. It really doesn't make sense to be called with False ever as the question is whether Conny should reject the image if alert sending failed and if Conny rejected in the first place, this is irrelevant.

Another question that came up while looking at this is whether we'd want to dispatch the alerts in the "rejected" case in the background as well. Because even in the rejected case we may run into a timeout (though it's less critical in the sense that we'd block the deployment anyways whereas a blocked deployment due to alerts being sent too slow would change behaviour). So I think I'll fix that up as well and send alerts in the background here too

xopham: sounds good

tests/test_flask_application.py
xopham: would consider that to be important if it happens

xopham:
why did you reduce code coverage? could you check the lines?

connaisseur/flask_application.py
```diff
@@ -116,14 +97,56 @@ def readyz():
...
+            return jsonify(response)
+        except Exception as err:
+            if isinstance(err, BaseConnaisseurException):
+                err_log = str(err)
+                msg = err.user_msg  # pylint: disable=no-member
+            elif isinstance(err, asyncio.TimeoutError):
+                err_log = str(traceback.format_exc())
+                msg = (
+                    "couldn't retrieve the necessary trust data for verification within 30s. "
+                    + "most likely there was a temporary network failure."
```
xopham, line 124: you are guessing here. It maybe that the server cannot be reached. Would be a bit more instructive like "check connection..."

tests/test_flask_application.py
Starkteetje: Ok, will open an issue

xopham (approved):
lgtm

PR #890, opened 2023-02-17
Fix: Repair script for getting root key

Starkteetje:
## Description

The root key script was broken with https://github.com/sse-secure-systems/connaisseur/pull/873 due to a new argument added to an internal function, which the root key script uses. This commit adds that missing argument to the call inside the root key script

## Checklist


xopham (approved):
lgtm

PR #921, opened 2023-03-03
Ci/new ci 2

xopham:
Fixes #

## Checklist


.github/actions/context/action.yaml
```diff
@@ -93,21 +87,33 @@ runs:
...
       id: show_context
       run: |
-        BRANCH_TAG=$(echo "${{ steps.meta.outputs.tags }}" | cut -d ':' -f 2)
-        echo BRANCH_TAG=${BRANCH_TAG} >> ${GITHUB_OUTPUT}
+        PREFIX=$(echo "${{ steps.get_context.outputs.BUILD_REGISTRY }}/${{ steps.get_context.outputs.BUILD_REPO }}:" | sed 's%/%\/%g')
+        TAGS="${{ steps.meta.outputs.tags }}"
+        REF_TAGS="${TAGS//${PREFIX}/}"
+        BUILD_IMAGE=$(echo "${TAGS}" | tail -2 | head -1)
+        BUILD_TAG="${BUILD_IMAGE//${PREFIX}/}"
+        [[ ${BUILD_TAG} == "sha-"* ]] || exit 1 # check as parsing of the BUILD_TAG maybe fragile and dependent on docker/metadata-action priorities
```
phbelitz, line 112: This breaks the pipeline right? How do we recover should the parsing be fragile ...?

.github/actions/safety/action.yaml
```diff
@@ -1,24 +1,35 @@
...
+      if: inputs.registry != ''
+      uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a # v2.1.0
+      with:
+        registry: ${{ inputs.registry }}
+        username: ${{ inputs.repo_owner }}
+        password: ${{ inputs.repo_token }}
     - name: Scan
-      run: safety check -r ./actual_package_versions.txt --full-report -o screen --save-json safety-report.json
+      run: |
+        docker run --user 0 -v ${PWD}:/results ${{ inputs.image }} /bin/ash -c \
```
phbelitz, line 28: do we actually expect different results from the images, that this complicated approach now is necessary?

.github/workflows/.reusable-docs.yaml
```diff
@@ -17,13 +15,22 @@ jobs:
...
         run: |
           pip install -r docs/requirements_docs.txt
       - name: Deploy
         run: |
-          mike deploy --push develop
+          if [[ "${GHREF}" == "refs/tags/v"* ]];
+          then
+            mike deploy --push --update-aliases ${RELEASE_VERSION} latest
+          elif [[ "${GHREF}" == "refs/heads/develop" ]]; then
+            mike deploy --push ${RELEASE_VERSION}
```
phbelitz, line 33: `mkdocs build` builds the docs and `mike deploy` builds and pushes ?

.github/actions/grype/action.yaml
```diff
@@ -0,0 +1,49 @@
...
+    - name: Login with registry
+      if: inputs.registry != ''
+      uses: docker/login-action@f4ef78c080cd8ba55a85445d5b36e214a81df20a # v2.1.0
+      with:
+        registry: ${{ inputs.registry }}
+        username: ${{ inputs.repo_owner }}
+        password: ${{ inputs.repo_token }}
+    - name: Scan
+      if: inputs.output == 'table'
+      uses: anchore/scan-action@v3
```
phbelitz, line 32: sha pinning?

.github/actions/context/action.yaml
xopham: I certainly do hope it is not. If that were to happen, I would above distinctly grep the one that is tagged with `sha-...` which could be done either way. Currently, sort of trust it to be the second to last which is supported by the metadata action prioritization. As they mention it explicitly, I guess they are aware that it should not simply be re-ordered

.github/actions/safety/action.yaml
xopham: the approach is actually the [recommended one by safety](https://github.com/pyupio/safety/#scan-a-python-based-docker-image). Now, there is 2 reasons to use this one:
- scanning from within the image scans the actually installed versions not the "presumably" installed versions
- scanning from within scans the actual python environment not only the pinned versions from `requirements.txt`
- only in this way can we actually scan the automated builds and the released version separately
that was my rationale for the switch

.github/workflows/.reusable-docs.yaml
xopham: yes, that is also how it worked in the past if you check out the corresponding deleted files

xopham: e.g. `.github/workflows/docs.yaml`

.github/actions/grype/action.yaml
xopham: correct, I missed that. Will however at some point add the sha pins for all external actions. Still fixing here as well

phbelitz (approved):

PR #922, opened 2023-03-03
ci: fix codecov upload

phbelitz:
## Description

Fixes codecov upload.

## Checklist

### DIFF
--- a/.github/workflows/.reusable-unit-test.yml
+++ b/.github/workflows/.reusable-unit-test.yml
@@ -9,7 +9,7 @@ jobs:
   pytest:
     runs-on: ubuntu-latest
     container:
-      image: python:slim
+      image: python:3
     steps:
       - name: Checkout code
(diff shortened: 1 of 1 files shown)


.github/workflows/.reusable-unit-test.yml
```diff
@@ -15,6 +15,7 @@ jobs:
         uses: actions/checkout@93ea575cb5d8a053eaa0ac8fa3b40d7e05a33cc8 # v3.1.0
       - name: Install
         run: |
+          apt update && apt install -y git
```
xopham, line 18: you should be able to use the non-slim python container above and not have to do an update/install here, which may be faster

xopham: also, I do believe that "ubuntu-latest" does contain python, so I am not certain why we chose the container to begin with.

phbelitz: tried the `ubuntu:22.04` container ... doesn't have python or git installed ... :(

xopham (approved):
lgtm

PR #927, opened 2023-03-04
ci: merge upgrade test into integration script

xopham:
Fixes no issue

## Description

- moves upgrade test into integration test script
- full refactoring only in followup MR

## Checklist


tests/integration/integration-test.sh
```diff
@@ -21,21 +21,24 @@ COSIGN_PUBLIC_KEY="$(printf -- "${COSIGN_PUBLIC_KEY//<br>/\\n          }")"
...
 fi
 
 ### SINGLE TEST CASE ####################################
 single_test() { # ID TXT TYP REF NS MSG RES
   echo -n "[$1] $2"
-  i=0 # intialize iterator
-  while : ; do
-    i=$((i+1))
+  i=0                                                              # intialize iterator
+  export RAND=$(head -c 5 /dev/urandom | hexdump -ve '1/1 "%.2x"') # creating a random index to label the pods and avoid name collision for repeated runs
```
Starkteetje: Would it be a better solution to simply set CTR=0 at start of file and inside single_test CTR+=1? Then we cant have random collisions fail the test and it works for arbitrarily many tests

.github/workflows/.reusable-integration-test.yml
```diff
@@ -173,88 +174,3 @@ jobs:
...
-          yq e '.' helm/values.yaml
-      - name: Install Connaisseur (master)
-        run: |
-          make install
-      - name: Get image name & version (master)
-        run: |
-          kubectl get pods -n connaisseur -o jsonpath="{.items[*].spec.containers[*].image}"
-      - name: Run integration tests (master)
-        run: |
-          bash tests/integration/upgrade-integration-test.sh
```
Starkteetje, line 232: Is this then no longer used and can be deleted?

tests/integration/integration-test.sh
xopham: that would actually only work as long we are worried about collisions within this single run of the script. However, if we re-run a different test later, we will definitely get collisions. Therefore, I'd prefer to stick to the random one

Starkteetje: 👍

Issue #910, opened 2023-02-24
Wrong filename for Unchaged Approval


phbelitz:
**Describe the bug**
The filename for the unchanged update approval feature has a wrong filename. Fix it.

PR #957, opened 2023-03-20
ci/fix/refactor: Check commit messages have identifier and small fixes

### LINKED ISSUE #905: Check commit messages conform to desired format
**Describe the feature**
Since we want to enforce commit messages of a certain format, e.g. `fix: Fix leThing`, it would be nice to actually test/enforce this in the pipeline, so we get at least a warning if a commit doesn't conform

https://github.com/sse-secure-systems/connaisseur/pull/900/commits contains a commit without proper format and we only noticed later

Extra job in pipeline that fails if format is not met

Starkteetje:
Adds a CI step that checks whether all added commits have an identifier like fix, ci or test in front of them as specified in CONTRIBUTING.MD

See e.g. https://github.com/sse-secure-systems/connaisseur/actions/runs/4467696912/jobs/7847485473 for output

Fixes wrongly named docs file and removes some whitespace

Fixes #905 #910 

## Checklist


.github/workflows/cicd.yaml
```diff
@@ -94,3 +94,32 @@ jobs:
       build_image: ${{ needs.build.outputs.build_image }}
       skip_integration_tests: ${{ needs.conditionals.outputs.skip_integration_tests }}
       cosign_public_key: ${{ needs.build.outputs.cosign_public_key }}
+
+  check-commit-message:
```
phbelitz, line 98: nit: maybe group together the job, like all other jobs are groups together aswell ...? thinking about something like a `meta` group in which `check-commit-message` and `conditionals` reside. 

might be a lot of refactoring, since `conditionals` is used everywhere and needs to be relabeled ...? ignore if to much effort.

phbelitz (approved):

.github/workflows/cicd.yaml
xopham, line 98: Can you make that a step in `.reusable_compliance`?

xopham (changes requested):
Think we should foster the new structure and maintain grouping

.github/workflows/cicd.yaml
Starkteetje: Given @xopham s comment, wouldn't mix it with conditionals as it would be in group compliance, but can have a look whether that makes a reasonable group

xopham: Conditionals would be the other option

Starkteetje: Wouldn't put it in conditionals as that would then fail build and would necessitate us making conformant commits even during testing, as opposed to at the end

.github/workflows/.reusable-compliance.yml
```diff
@@ -40,3 +40,32 @@ jobs:
...
+          echo "Commits between dev branch and current SHA:"
+          COMMITS=$(git log --pretty=%H)
+          echo "${COMMITS}"
+          EXIT=0
+          COMMIT_MSGS=$(git log --pretty=%s) # show subject only
+          for commit in ${COMMITS}; do
+              MSG=$(git log ${commit} -n1 --pretty=%s)
+              TYPE=$(echo ${MSG} | awk '{{ print $1 }}')
+              if ! [[ "${TYPE}" =~ ^(build|ci|docs|feat|fix|refactor|test|update):$ ]]; then
+                EXIT=1
```
xopham, line 65: did you test that? `EXIT` is not scoped within for loop only, right?

xopham: ok

xopham (approved):
lgtm

PR #984, opened 2023-03-30
Refactoring of helm/values.yaml

Starkteetje:
## Description
Fix further instances of new version definition and refactor syntax of helm/values.yaml

## Checklist


phbelitz:
be aware that there was a reason for using snake_case as variable names in the values.yaml. the variables are taken as they are from the values.yaml to the python functions as parameters and the python linter wants snake_case for it's variables. 

so we either have to ignore the pylint issues, or do a mapping from camelCase values.yaml variables to snake_case python variables.

phbelitz:
please change for the cosign validator the `host` field to `host.rekor` (referring to https://github.com/sse-secure-systems/connaisseur/pull/969#discussion_r1154433235)

helm/templates/env.yaml
```diff
@@ -9,25 +9,19 @@ data:
   CONNAISSEUR_NAMESPACE: {{ .Release.Namespace }}
```
phbelitz, line 9: i have a vision:

```suggestion
  CONNAISSEUR_NAMESPACE: {{ .Release.Namespace }}
  KUBE_API_TOKEN_PATH: /var/run/secrets/kubernetes.io/serviceaccount/token
  KUBE_API_CA_PATH: /var/run/secrets/kubernetes.io/serviceaccount/ca.crt
  LOG_LEVEL: {{.Values.application.logLevel | default "INFO"}}
  {{- with .Values.application.features }}
  DETECTION_MODE: {{ default false .detectionMode | quote }}
  AUTOMATIC_CHILD_APPROVAL_ENABLED: {{ default true .automaticChildApproval | quote }}
  AUTOMATIC_UNCHANGED_APPROVAL: {{ default true .automaticUnchangedApproval | quote }}
  {{- end }}
  {{- if .Values.alerting }}
  CLUSTER_NAME: {{ default "not specified" .Values.alerting.clusterIdentifier }}
  {{- end }}
  KUBE_VERSION: {{ .Capabilities.KubeVersion }}
```

then have a new function in `utils.py`, something like `feature_flag_on(key: str)` which can parse the string bool values into actual bools. also put all the feature flag keys into `constants.py`.

or keep things as they are and i'll make a seperate PR with these changes 🤷

test.yml
```diff
@@ -0,0 +1,23 @@
+application:
```
phbelitz, line 1: this file needs to be removed?

helm/templates/certificate_webhook-conf.yaml
```diff
@@ -100,15 +100,11 @@ webhooks:
...
+        {{- if .Values.application.features.automaticChildApproval }}
         resources: ["pods", "deployments", "replicationcontrollers", "replicasets", "daemonsets", "statefulsets", "jobs", "cronjobs"]
         {{- else }}
         resources: ["pods"]
         {{- end }}
-        {{- else }}
-        resources: ["pods", "deployments", "replicationcontrollers", "replicasets", "daemonsets", "statefulsets", "jobs", "cronjobs"]
-        {{- end }}
     sideEffects: None
     {{- if gt ($k8sMinor | int) 13 }}
```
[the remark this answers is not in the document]
Starkteetje: Wouldn't that be purposefully breaking old versions? Imo, I would make it a policy to expand no effort to fixing old versions, but to not break them on purpose (see also my [comment here](https://github.com/sse-secure-systems/connaisseur/issues/496#issuecomment-1496102226)). I'm not super strongly opinionated here, though

docs/migrating_to_version_3.md
```diff
@@ -0,0 +1,54 @@
...
+- [Changes](#api-changes) to the Helm `values.yaml` file
+    - :boom: This is a breaking change as we touched quite a bunch of configuration keys to make the configuration API more consistent and more intuitive
+    - :robot: We've prepared a [script](TODO) that migrates your existing Connaisseur configuration to the new format
+
+## API changes
+
+Here's the list of changes we made to the Helm `values.yaml`:
+
+- Top-level `kubernetes` key
+    - Previously, there were top-level `deployment` and `service` keys, both of which concerned themselves with the respective<sup>*</sup> Kubernetes resources.
```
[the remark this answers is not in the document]
Starkteetje: It doesn't currently lead anywhere and is more or less a "be careful, this is nit quite correct" due to the fact, that deployment also had config for the webhook, so it containing config for the corresponding resource isn't quite true. I can link the change below or add a footnote instead if you like that more

helm/templates/env.yaml
Starkteetje: Since that's no longer refactoring of the helm/values.yaml and this MR is already quite big, I'd say it'd be better to move it to a dedicated MR instead

docs/migrating_to_version_3.md
```diff
@@ -0,0 +1,56 @@
...
+It's been a while since our last major update, but it is time again :tada:
+Connaisseur version 3.0 is out and brings along many new features, but also breaking changes :boom:
+For those breaking changes, we've set up a [script](../scripts/upgrade.py) that migrates your existing Connaisseur configuration.
+Read on for the list of most interesting changes.
+
+## Major changes
+
+- NEW: Support for Cosign 2.0
+    - :boom: This is a breaking change, because Connaisseur now supports transparency log verification by default.
+    If all your Cosign signature artifcats are already part of Rekor's default transparency log (or part of your configured Rekor host) or you don't use Cosign, this is not breaking for you. TODO check phbelitz
```
Starkteetje, line 12: 👀 @phbelitz

phbelitz:
```suggestion
    If all your Cosign signature artifacts are already part of Rekor's default transparency log (or part of your configured Rekor host) or you don't use Cosign, this is not breaking for you.
```
absolutely correct.

scripts/upgrade.py
```diff
@@ -0,0 +1,141 @@
...
+    with open("helm/values.yaml", "r") as f:
+        config = yaml.safe_load(f)
+
+    ### Create backup copy
+    with open("helm/values.yaml.old", "w") as f:
+        yaml.dump(config, f)
+
+    ### Remove image tag
+    image = config["deployment"]["image"]
+    repo, _ = image.split(":")
```
phbelitz, line 14:
```suggestion
    repo = "".join(image.split(":")[:-1])
```
just in case someone pushed the connaisseur image into the own private registry at `private-registry:3000/library/connaisseur:v2`

helm/templates/env.yaml
phbelitz: i will take care of this.

phbelitz (approved):

PR #1002, opened 2023-04-13
refactor: feature flag utility

phbelitz:
Checks for feature flags have been unified into a single function.

## Checklist

- [x] PR is rebased to/aimed at branch develop`


connaisseur/util.py
```diff
@@ -130,3 +131,8 @@ def get_kube_version():
     regex = r"v(\d)\.(\d{1,2})\.(.*)"
     match = re.match(regex, version)
     return match.groups() if match else ("0", "0", "0")
+
+
+def feature_flag_on(flag: str, default: bool = False):
```
Starkteetje, line 136: Maybe this is a nice time to make defaults a single thing. I think we're sure that the env var must exist due to defaults in `helm/templates/env.yaml`, right? Otherwise we could also not set defaults in Helm and instead have a feature_flag.py (or the like), which would get the feature value from environment OR the appropriate default. 

Both solutions would prevent multiple instances of defaults being spread around the code, which could hamper development if we changed it in the future

Starkteetje (changes requested):

tests/test_util.py
```diff
@@ -145,3 +143,22 @@ def test_validate_schema(
...
+
+@pytest.mark.parametrize(
+    "flag, value, default, expected_value",
+    [
+        ("FLAG", True, None, True),
+        ("FLAG", "1", None, True),
+        ("FLAG", "True", None, True),
+        ("FLAG", "False", None, False),
+        ("FLAG", "0", None, False),
+        ("FLAG", None, None, False),
```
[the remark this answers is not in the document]
phbelitz: fixed

connaisseur/exceptions.py
```diff
@@ -13,7 +15,8 @@ class BaseConnaisseurException(Exception):
 
     def __init__(self, message: str = default_message, **kwargs):
         self.message = message.format(**kwargs)
-        self.detection_mode = os.environ.get("DETECTION_MODE", "0") == "1"
+        # cant use feature flag, because of circular import
```
Starkteetje, line 18:
```suggestion
        # can't use feature flag, because of circular import
```

connaisseur/util.py
```diff
@@ -32,7 +33,6 @@ def get_admission_review(
     allowed: bool,
     patch: Optional[list] = None,
     msg: Optional[str] = None,
-    detection_mode: bool = False,
```
Starkteetje, line 35: Need to update method description too

connaisseur/util.py
```diff
@@ -130,3 +131,13 @@ def get_kube_version():
     regex = r"v(\d)\.(\d{1,2})\.(.*)"
     match = re.match(regex, version)
     return match.groups() if match else ("0", "0", "0")
+
+
+def feature_flag_on(flag: str):
+    env = os.environ.get(flag)
+
+    # only false if env == None
+    if not env:
```
Starkteetje, line 140: Before there were defaults for some flags, i.e. `os.environ.get("AUTOMATIC_CHILD_APPROVAL_ENABLED", "1")`. Why is this not a breaking change for these situations?

connaisseur/util.py
```diff
@@ -130,3 +131,13 @@ def get_kube_version():
     regex = r"v(\d)\.(\d{1,2})\.(.*)"
     match = re.match(regex, version)
     return match.groups() if match else ("0", "0", "0")
+
+
+def feature_flag_on(flag: str):
+    env = os.environ.get(flag)
+
+    # only false if env == None
```
Starkteetje, line 139: `not ""` is also True

connaisseur/util.py
Starkteetje: Both in the sense of "before undefined defaulted to something instead of boom" and technically for "before 'y' meant False, now it's True". The later should be fine, if you highlight it in the v3 changelog as breaking. The first seems unnecessary

tests/test_util.py
```diff
@@ -145,3 +143,28 @@ def test_validate_schema(
...
+
+
+@pytest.mark.parametrize(
+    "flag, value, expected_value, exception",
+    [
+        ("FLAG", True, True, fix.no_exc()),
+        ("FLAG", "1", True, fix.no_exc()),
+        ("FLAG", "True", True, fix.no_exc()),
+        ("FLAG", "False", False, fix.no_exc()),
+        ("FLAG", "0", False, fix.no_exc()),
```
Starkteetje, line 155: The list is short enough, would test all values

Starkteetje: And add test for empty string as regression of above comment

Starkteetje (changes requested):

docs/migrating_to_version_3.md
```diff
@@ -18,6 +18,8 @@ Read on for the list of most interesting changes.
 
 - Changed the log format to JSON :sunglasses:
     - The logging format of Connaisseur is now fully transformed into JSON. It includes a `timestamp` and a `message` field, next to additional fields depending on the context. Logs of the HTTP request do not contain a `message` field.
+- Different behavior on reading feature flags
```
Starkteetje, line 21: Would put to both major changes and API changes and add migration script
Would formulate more from an end user perspective, instead of implementation details. "Before any truthy value ..." or the like

docs/migrating_to_version_3.md
```diff
@@ -51,9 +49,10 @@ Here's the list of changes we made to the Helm `values.yaml`:
     The idea was based around whether the feature would have further configuration or not.
     However, having configuration for a disabled feature doesn't really make sense, so we changed it to be the following:
         - If a feature has configuration, i.e. `<featureName>.<someConfigKey>` is set, it is enabled.
-        - If a feature is set, i.e. `<featureName>: true`  it is enabled and an error is thrown if it needed configuration but doesn't have any.
-        - If a feature is false, i.e. `<featureName>: false`  it is disabled.
+        - If a feature is set, i.e. `<featureName>: true` it is enabled and an error is thrown if it needed configuration but doesn't have any.
```
Starkteetje, line 52:
```suggestion
        - If a feature is set, i.e. `<featureName>: true` it is enabled and an error is thrown if it needs configuration but doesn't have any.
```

docs/migrating_to_version_3.md
```diff
@@ -51,9 +49,10 @@ Here's the list of changes we made to the Helm `values.yaml`:
...
+        - If a feature is false, i.e. `<featureName>: false` it is disabled.
         - If a feature is not set at all, i.e. `features` doesn't have a `<featureName>` key, its default configuration is chosen.
+        - If a feature has no configuration, i.e. `<featureName>: true`, the only allowed values are `true` or `false`. Otherwise an error is thrown.
```
Starkteetje, line 55:
```suggestion
        - If a feature has no configuration, i.e. `<featureName>: true`, the only allowed values are `true` or `false`.
        Otherwise an error is thrown at time of deployment.
```

Starkteetje (approved):

PR #1126, opened 2023-06-29
feat: add serviceaccount annotations to allow IRSA

### LINKED ISSUE #1124: Need the ability to add annotations to serviceaccount in helm chart
**Describe the feature**

We need the ability to add annotations to the serviceaccount in the helm chart, which will allow people to enable IRSA for EKS, which will let things like cosign be able to access ECR or use KMS keys or other activities that require IAM permissions.

There is no way to add annotations to the serviceaccount in the helm chart, so we cannot use IRSA.

https://github.com/sse-secure-systems/connaisseur/pull/1121 implements this feature.

phbelitz:
This allows us to add annotations to the serviceaccount in the helm chart, which will allow people to enable IRSA for EKS, which will let things like cosign be able to access ECR or use KMS keys or other activities that require IAM permissions.

fixes #1124

## Checklist


helm/values.yaml
```diff
@@ -78,6 +78,11 @@ kubernetes:
     type: ClusterIP
     port: 443
 
+  # configure Connaisseur serviceaccount (so you can do IRSA, for example)
+  serviceaccount:
```
Starkteetje, line 82: This looks like something that we'd previously have commented out in the default helm because it's not really neccessary and can be ignored by most people.

Would be fine with approving, if you want it this way though

Starkteetje (approved):

PR #1144, opened 2023-07-12
fix: make failures visible via metrics in detection mode

FalacerSelene:
When monitoring Connaissuer via the prometheus metrics endpoint, there was previously no way to detect failures when detection mode was turned on, as it changed the status code and 'allowed' field to True (as designed). This change adds a boolean label to the metric that indicates whether warnings are present - and warnings are only present when the operation would have failed but detection mode is on.

## Description

I am running Connaissuer in detection mode and feeding the metrics into Grafana to monitor image signing in my cluster. As-is, there's no way for me to detect image signing failures - I don't want to turn detection mode off as that would risk causing an outage if there's some problem. 

This change adds a label to the metric to indicate if there were any warnings present in the response. From `util.py`, you can see warnings are only added to the response if detection mode is on and the request would otherwise be disallowed. This means checking for the warnings in the response will cover exactly the case I want. 

This repo doesn't have any tests or documentation of the metrics, so I've not changed anything there. 

## Checklist

(tests & adjusted versions not necessary)

### DIFF
--- a/connaisseur/flask_application.py
+++ b/connaisseur/flask_application.py
@@ -56,6 +56,7 @@ def handle_alert_config_error(err):
     labels={
         "allowed": lambda r: metrics_label(r, "allowed"),
         "status_code": lambda r: metrics_label(r, "status_code"),
+        "warnings": lambda r: metrics_label(r, "warnings"),
     },
 )
 def mutate():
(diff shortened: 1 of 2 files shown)


phbelitz:
thank you very much! because of permission issues in our pipeline, I have to recreate the PR under my name, but i pulled in your commit. so closing this in favor of the newly created one (#1148). cheers.

PR #1148, opened 2023-07-13
Make failures visible via metrics in detection mode

phbelitz:
When monitoring Connaissuer via the prometheus metrics endpoint, there was previously no way to detect failures when detection mode was turned on, as it changed the status code and 'allowed' field to True (as designed). This change adds a boolean label to the metric that indicates whether warnings are present - and warnings are only present when the operation would have failed but detection mode is on.

## Checklist


peterthomassen (approved):
lgtm!

PR #1242, opened 2023-09-18
feat: Add pod securityContext to helm chart

### LINKED ISSUE #1240: Helm Chart: Allow to configure the pod security context
**Describe the feature**
Add the possibility to the connaisseur helm chart to configure the _PodSecurityContext_ for the pods in the connaisseur deployment (in addition to configuring the container-level _SecurityContext_, which is already possible today).

Currently, the helm chart allows to define the container-level SecurityContext [here](https://github.com/sse-secure-systems/connaisseur/blob/develop/helm/templates/deployment.yaml#L93), but not the pod-level PodSecurityContext. This is a problem in environments that require security policies on the pod level (e.g., when working in a cloud environment where policies forbid the deployment of pods without strong enough PodSecurityContext).

The helm chart should allow to specify both pod-level and container-level security context separately, because they follow different schemas, permitting to set different policies. (Some of the policies can be specified both on pod- and on container-level like `runAsNonRoot`, but there are some that can only be specified on pod-level like `supplementalGroups` and some that can only be specified on container-level like `allowPrivilegeEscalation`.)

The template for the deployment [here](https://github.com/sse-secure-systems/connaisseur/blob/develop/helm/templates/deployment.yaml) could be extended by a PodSecurityContext, whose contents are taken from Helm values. The default value would be to not have a PodSecurityContext (to keep the current behavior of the chart in case the Helm values do not specify a PodSecurityContext).

hsudbrock:
Fixes #1240

## Description

The current helm chart does not permit to add a pod-level security context, only a container-level security context. With this pull request, the helm chart is extended with an optional pod-level security context.

The change is such that pod-level and container-level security context can be specified separately. Why? Because there are configuration options in the pod-level security context that do not apply to the container-level security context and vice versa (examples are given in the fixed ticket #1240).

The new key in the values.yaml file for the pod-level security context is `podSecurityContext` (to distinguish it from the existing `securityContext`, which applies to the container). Renaming `securityContext` to `containerSecurityContext` to make things more explicit sounds nice, but would be breaking, which is why it is not changed. (But maybe it should be changed? Not sure.)

The default value for the `podSecurityContext` is empty, in which case the generated `Deployment` yaml will not contain a pod-level security context.

I have increased the chart version from 2.0.0 to 2.0.1 with the change -  I a not sure if this is the correct version bump, if not I would be happy about advice.

## Checklist

- [ X] PR is rebased to/aimed at branch `develop`
- [ X] PR follows [Contributing Guide](https://github.com/sse-secure-systems/connaisseur/blob/master/docs/CONTRIBUTING.md)
- [ X] Added tests (if necessary)
- [ X] Extended README/Documentation (if necessary)
- [ X] Adjusted versions of image and Helm chart in `Chart.yaml` (if necessary)

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: connaisseur
 description: Helm chart for Connaisseur - a Kubernetes admission controller to integrate container image signature verification and trust pinning into a cluster.
 type: application
-version: 2.0.0
+version: 2.0.1
 appVersion: 3.0.0
 keywords:
   - container image
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -106,6 +106,12 @@ spec:
       tolerations:
         {{- toYaml . | nindent 8 }}
     {{- end }}
+    {{- if .Values.kubernetes.deployment.podSecurityContext }}
+      {{- with .Values.kubernetes.deployment.podSecurityContext }}
+      securityContext:
+          {{- toYaml . | nindent 8 }}
+      {{- end }}
+    {{- end }}
       volumes:
         - name: {{ .Chart.Name }}-certs
           secret:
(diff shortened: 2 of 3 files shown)


phbelitz:
was megred as part of #1264

PR #1264, opened 2023-09-29
feat: Add pod securityContext to helm chart

### LINKED ISSUE #1240: Helm Chart: Allow to configure the pod security context

phbelitz:
The current helm chart does not permit to add a pod-level security context, only a container-level security context. The helm chart is extended with an optional pod-level security context.

Fix #1240

duplicate of #1242

## Checklist

### DIFF
--- a/helm/values.yaml
+++ b/helm/values.yaml
@@ -36,6 +36,7 @@ kubernetes:
             weight: 100
     #annotations:  # uncomment when using Kubernetes prior v1.19
     #  seccomp.security.alpha.kubernetes.io/pod: runtime/default  # uncomment when using Kubernetes prior v1.19
+    # container wide security context
     securityContext:
       allowPrivilegeEscalation: false
       capabilities:
@@ -48,6 +49,8 @@ kubernetes:
       runAsGroup: 20001 # remove when using openshift or OKD 4
       seccompProfile: # remove when using Kubernetes prior v1.19, openshift or OKD 4
         type: RuntimeDefault # remove when using Kubernetes prior v1.19, openshift or OKD 4
+    # pod wide security context
+    podSecurityContext: {}
     # PodSecurityPolicy is deprecated as of Kubernetes v1.21, and will be removed in v1.25
     podSecurityPolicy:
       enabled: false
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: connaisseur
 description: Helm chart for Connaisseur - a Kubernetes admission controller to integrate container image signature verification and trust pinning into a cluster.
 type: application
-version: 2.0.0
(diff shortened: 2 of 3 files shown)


helm/values.yaml
```diff
@@ -48,6 +48,7 @@ kubernetes:
       runAsGroup: 20001 # remove when using openshift or OKD 4
       seccompProfile: # remove when using Kubernetes prior v1.19, openshift or OKD 4
         type: RuntimeDefault # remove when using Kubernetes prior v1.19, openshift or OKD 4
+    podSecurityContext: {}
```
xopham, line 51: Think you now need to comment that the above is container level sec context or at least make this transparent why there is two

xopham: Is there an exclusion to havin both?

xopham (commented):
Generall lgtm, just 1 minor comment

helm/values.yaml
xopham: https://kubernetes.io/docs/tasks/configure-pod-container/security-context/#set-the-security-context-for-a-container

> Security settings that you specify for a Container apply only to the individual Container, and they override settings made at the Pod level when there is overlap. Container settings do not affect the Pod's Volumes.

phbelitz: added comment.

> Is there an exclusion to havin both?
i don't understand. there are some setting that can be done in A(container) alone, B(pod) alone and A + B together.

xopham (approved):
Lgtm. Good stuff. Thx @hsudbrock

hsudbrock:
Thanks @xopham and @phbelitz!

PR #1274, opened 2023-10-04
Update flask_application.py

phbelitz:
Fix for: issue #1205, loop close exception

duplicate of #1269 

## Checklist

- [x] PR is rebased to/aimed at branch `develop`
- [x] PR follows [Contributing Guide](https://github.com/sse-secure-systems/connaisseur/blob/master/docs/CONTRIBUTING.md)
- [x] Added tests (if necessary)
- [x] Extended README/Documentation (if necessary)
- [x] Adjusted versions of image and Helm chart in `Chart.yaml` (if necessary)

PR #1280, opened 2023-10-06
fix: overloaded requests

phbelitz:
The changes in https://github.com/sse-secure-systems/connaisseur/pull/1274 tried to fix the memory leak problem by directly using an event loop. This unfortunately caused performance issues, so that the load tests regularly failed. Using nest-asyncio fixed the performance issue.

## Checklist


xopham (approved):
lgtm

Issue #709, opened 2022-07-06
Prometheus cannot connect to `/metrics` HTTPS endpoint, throws "cannot validate certificate <IP> because it doesn't contain any IP SANs"


d4l-adrian-castro:
**Describe the bug**
Prometheus cannot connect to `/metrics` HTTPS endpoint, throws "cannot validate certificate <IP> because it doesn't contain any IP SANs"

**Expected behavior**
Prometheus can connect to `/metrics` HTTPS endpoint.

**Optional: To reproduce**
- Have Prometheus running on the same cluster as Connaisseur
- Add the following configuration to Connaisseur:
```
connaisseur:
  deployment:
    annotations:
      prometheus.io/scrape: "true"
      prometheus.io/path: "/metrics"
      prometheus.io/port: "5000"
      prometheus.io/scheme: "https"
  ...
```

**Optional: Versions (please complete the following information as relevant):**
 - OS: <OS and version>
 - Kubernetes Cluster: `v1.23.4`
 - Notary Server: `0.6.1`
 - Connaisseur: 2.6.1
 - Prometheus: `quay.io/prometheus/prometheus:v2.34.0`

**Optional: Additional context**
![image](https://user-images.githubusercontent.com/90603637/177520571-970e570a-9a6b-49f5-8200-47845af57a05.png)

d4l-adrian-castro:
Probably related to https://github.com/sse-secure-systems/connaisseur/issues/225

xopham:
@d4l-adrian-castro interesting issue. I never tested the full integration with prometheus myself, but considered it would be working. We'll need to test it.
Could also be related or resolvable by #572

xopham:
would you have any further details on cluster flavor (kind, k3s, aks, eks, gke, ...) or other information that may help reproducing?

d4l-adrian-castro:
Just k8s, direct upstream

(sorry for the delay, forgot about the notification)

xopham:
@phbelitz thoughts on this?

phbelitz:
The more I look into it, the less trivial this issue seems to be ... let me outline you the problem as much as I have grasped it by now:

[...]

PR #1306, opened 2023-10-12
feat: explicitly specify containerPort in helm chart

### LINKED ISSUE #1305: Helm chart: Explicitly describe the exposed port
**Describe the feature**
In the Helm chart for connaisseur, explicitly specify the port exposed by the container port (cf. https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.28/#containerport-v1-core).

As described in #709, prometheus has issues scraping metrics from connaisseur, because the metrics are exposed via HTTPS, and prometheus cannot validate the HTTPS certificate. To circumvent this problem, one option is to tell prometheus to skip certificate validation (which is, for some, ok for  metrics scraping), e.g., using a [PodMonitor](https://github.com/prometheus-operator/prometheus-operator/blob/main/Documentation/api.md#podmonitor) from the prometheus operator. But: PodMonitors require the port they scrape to be specified explicitly for the container.

Note that using the already provided service to scrape metrics is not an option, because in case of a HA deployment of connaisseur each scrape call might go to different pods.

I will provide a MR for this, it's a small change.

hsudbrock:
Fixes #1305

## Description

Currently, the deployment from the connaisseur helm chart does not explicitly specify the containerPort. This is usually not a problem, because the port can still be used, even if it is not explicitly named. 

Explicitly specifying the port has two benefits:
* It documents the available ports for a user  of the chart.
* It enables to configure automatic scraping of metrics by prometheus using a `PodMonitor` from the prometheus operator (which requires the port to be explicitly specified). This helps in working around the issue described in #709, to scrape metrics even when they are provided via https only.

In principle, one could also name the port, but since there is only one exposed port and I did not want to use a generic name like `thePort`, I did not add a port name.

## Checklist

- [ x] PR is rebased to/aimed at branch `develop`
- [ x] PR follows [Contributing Guide](https://github.com/sse-secure-systems/connaisseur/blob/master/docs/CONTRIBUTING.md)
- [ x] Added tests (if necessary)
- [ x] Extended README/Documentation (if necessary)
- [ x] Adjusted versions of image and Helm chart in `Chart.yaml` (if necessary)

### DIFF
--- a/helm/Chart.yaml
+++ b/helm/Chart.yaml
@@ -2,7 +2,7 @@ apiVersion: v2
 name: connaisseur
 description: Helm chart for Connaisseur - a Kubernetes admission controller to integrate container image signature verification and trust pinning into a cluster.
 type: application
-version: 2.1.0
+version: 2.2.0
 appVersion: 3.1.1
 keywords:
   - container image
--- a/helm/templates/deployment.yaml
+++ b/helm/templates/deployment.yaml
@@ -33,6 +33,9 @@ spec:
         - name: {{ .Chart.Name }}
           image: "{{ .Values.kubernetes.deployment.image.repository }}:{{ default (print "v" .Chart.AppVersion) .Values.kubernetes.deployment.image.tag }}"
           imagePullPolicy: {{ .Values.kubernetes.deployment.imagePullPolicy }}
+          ports:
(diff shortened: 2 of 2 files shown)


phbelitz:
merged in #1308, thank you @hsudbrock

PR #1321, opened 2023-10-19
feat: add functional labels

xopham:
Fixes no issue

## Description

* adds functional labels to kubernetes resources
* facilitates and describes how to deploy Connaisseur via kubernetes manifests

## Checklist

### DIFF
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -74,7 +74,7 @@ However, validating your own images requires additional configuration.
 
 ### Deploy
 
-Install Connaisseur via Helm:
+Install Connaisseur via Helm or Kubernetes manifests:
 
 === "Git repo"
 
@@ -96,6 +96,32 @@ Install Connaisseur via Helm:
 
     ```bash
     helm install connaisseur connaisseur/connaisseur --atomic --create-namespace --namespace connaisseur -f values.yaml
+    ```
+
+=== "Kubernetes manifests"
+
+    Installing Conaisseur via Kubernetes manifests requires to first render the respecitve resources.
+    If the repo was cloned, simply render remplates via:
+    ```bash
+    helm template helm -n connaisseur > deploy.yaml
+    ```
+    Next, the admission controller is deployed step-wise:
+
+    0. Create target namespace:
+    ```bash
+    kubectl create namespace connaisseur
+    ```
+    1. Setup the preliminary hook:
+    ```bash
+    kubectl apply -f deploy.yaml -l 'app.kubernetes.io/component=connaisseur-init' -n connaisseur
+    ```
+    2. Deploy core resources
+    ```bash
+    kubectl apply -f deploy.yaml -l 'app.kubernetes.io/component=connaisseur-core' -n connaisseur
+    ```
+    3. Arm the webhook
+    ```bash
+    kubectl apply -f deploy.yaml -l 'app.kubernetes.io/component=connaisseur-webhook' -n connaisseur
     ```
 
 This deploys Connaisseur to its own namespace called `connaisseur`.
@@ -156,15 +182,48 @@ A running Connaisseur instance can be updated by a Helm upgrade of the current r
     helm upgrade connaisseur connaisseur/connaisseur -n connaisseur --wait -f values.yaml
     ```
 
+=== "Kubernetes manifests"
 
+    Adjust your local Kubernetes manifests (e.g. `deploy.yaml`) as required and upgrade via delete and reinstall:
+
+    ```bash
+    kubectl delete -f deploy.yaml -n connaisseur
+    kubectl apply -f deploy.yaml -l 'app.kubernetes.io/component=connaisseur-init' -n connaisseur
+    kubectl apply -f deploy.yaml -l 'app.kubernetes.io/component=connaisseur-core' -n connaisseur
+    kubectl apply -f deploy.yaml -l 'app.kubernetes.io/component=connaisseur-webhook' -n connaisseur
+    ```
+    !!! note
+        Rolling upgrades as with Helm might also be possible, but might require further configuration.
+        Insights are welcome :pray:
 
 ### Delete
 
 Just like for installation, Helm can also be used to delete Connaisseur from your cluster:
 
-```bash
-helm uninstall connaisseur -n connaisseur
-```
+
+=== "Git repo"
+
+    Uninstall via Helm:
+
+    ```bash
+    helm uninstall connaisseur -n connaisseur
+    ```
+
+=== "Helm chart"
+
+    Uninstall via Helm:
+
+    ```bash
+    helm uninstall connaisseur -n connaisseur
+    ```
+
+=== "Kubernetes manifests"
+
+    Delete via manifests:
+
+    ```bash
+    kubectl delete -f deploy.yaml -n connaisseur
+    ```
 
 In case uninstallation fails or problems occur during subsequent installation, you can manually remove all resources:
 
--- a/docs/basics.md
+++ b/docs/basics.md
@@ -102,7 +102,7 @@ Install Connaisseur via Helm or Kubernetes manifests:
 === "Kubernetes manifests"
 
     Installing Conaisseur via Kubernetes manifests requires to first render the respecitve resources.
-    If the repo was cloned, simply render remplates via:
+    If the repo was cloned, simply render templates via:
     ```bash
     helm template helm -n connaisseur > deploy.yaml
     ```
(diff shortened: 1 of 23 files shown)


docs/basics.md
```diff
@@ -17,8 +17,9 @@ The image policy defines a set of rules which maps different images to those val
 This is done via glob matching of the image name which for example allows to use different validators for different registries, repositories, images or even tags.
 This is specifically useful when using public or external images from other entities like Docker's official images or different keys in a more complex development team.
 
-> :octicons-light-bulb-16: **Note**: Typically, the public key of a known entity is used to validate the signature over an image's content in order to ensure integrity and provenance.
```
Starkteetje, line 20: Given the previous lightbulb, this might be type `tip` instead of `note`?

docs/basics.md
```diff
@@ -98,6 +99,32 @@ Install Connaisseur via Helm:
     helm install connaisseur connaisseur/connaisseur --atomic --create-namespace --namespace connaisseur -f values.yaml
     ```
 
+=== "Kubernetes manifests"
+
+    Installing Conaisseur via Kubernetes manifests requires to first render the respecitve resources.
+    If the repo was cloned, simply render remplates via:
```
Starkteetje, line 105:
```suggestion
    If the repo was cloned, simply render templates via:
```

docs/basics.md
```diff
@@ -156,15 +187,46 @@ A running Connaisseur instance can be updated by a Helm upgrade of the current r
...
+=== "Kubernetes manifests"
 
+    Adjust your local Kubernetes manifests (e.g. `deploy.yaml`) as required and upgrade via delete and reinstall:
+
+    ```bash
+    kubectl delete -f deploy.yaml -n connaisseur
+    kubectl apply -f deploy.yaml -l 'app.kubernetes.io/component=core' -n connaisseur
+    ```
+    !!! note
+        Rolling upgrades as with Helm might also be possible, but might require further configuration.
```
Starkteetje, line 199: Double might sounds meh

docs/basics.md
```diff
@@ -156,15 +187,46 @@ A running Connaisseur instance can be updated by a Helm upgrade of the current r
...
-```bash
-helm uninstall connaisseur -n connaisseur
-```
+
+=== "Git repo"
+
+    Uninstall via Helm:
+
+    ```bash
+    helm uninstall connaisseur -n connaisseur
```
Starkteetje, line 212: Would this be a place where we use the simpler versions of `make uninstall`? (Would then need the same during install)

docs/basics.md
```diff
@@ -312,7 +376,7 @@ The reference to validator and exact trust root is resolved in the following way
 Let's review the pattern and validator matching at a minimal example.
 We consider the following validator and policy configuration (most fields have been omitted for clarity):
 
-```yaml
+```yaml title="helm/values.yaml"
```
Starkteetje, line 379: Not sure how this is rendered, but to keep things consistent, wouldn't it be better to use the same syntax above and below with the

> ??? abstract "helm/values.yaml"
    ```yaml

boxes?

docs/features/README.md
```diff
@@ -1,11 +1,10 @@
 # Overview
 
-Besides Connaisseur's central functionality, several additional features are available:
+Besides Connaisseur's central functionality, several additional features are available, such as:
 
+- [Alerting](./alerting.md): *send alerts based on verification result*
 - [Detection Mode](./detection_mode.md): *warn but do not block invalid images*
 - [Namespaced Validation](./namespaced_validation.md): *restrict validation to dedicated namespaces*
-- [Alerting](./alerting.md): *send alerts based on verification result*
-- [Automatic Child Approval](automatic_child_approval.md): *configure approval of Kubernetes child resources*
```
Starkteetje, line 8: Why no more ACA? If due to experimental, wouldn't an experimental feature list be the better solution than to drop?

docs/getting_started.md
```diff
@@ -22,7 +22,8 @@ For more advanced usage in more complex cases (e.g. authentication, multiple reg
 
 In case you need help, feel free to reach out via [GitHub Discussions](https://github.com/sse-secure-systems/connaisseur/discussions) :wave:
 
-> :octicons-light-bulb-16: **Note**: As more than only public keys can be used to validate integrity and provenance of an image, we refer to these trust anchors in a generalized form as *trust roots*.
+!!! note
```
Starkteetje, line 25: Due to lihtbulb, maybe box type tip?

helm/templates/alertconfig.yaml
```diff
@@ -5,10 +5,8 @@ metadata:
   name: {{ .Chart.Name }}-alert-templates
   namespace: {{ .Release.Namespace }}
   labels:
-    app.kubernetes.io/name: {{ include "helm.name" . }}
-    helm.sh/chart: {{ include "helm.chart" . }}
-    app.kubernetes.io/instance: {{ .Chart.Name }}
-    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/component: core
```
Starkteetje, line 8: imo it might be a good idea to add some form of `connaisseur` to either label name or value to disambiguate it from other such labels

Starkteetje (changes requested):

docs/basics.md
xopham: `mkdocs serve` to see how `!!! `, `???` and `title="helm/values.yaml"` are rendered. I was uncertain as well and decided to go by "is the output part of the point here or just FYI. Feel free to share your thoughts for a general strategy

docs/features/README.md
xopham: Thought I switched from "this is the list" (where the list was incomplete) to "feature such as" (which is a selective list of interesting features. But needs adjustment either way. Central readme as well

docs/getting_started.md
xopham: it's not really a tip though

xopham: using `info`

docs/validators/sigstore_cosign.md
```diff
@@ -160,7 +162,9 @@ kubectl create secret docker-registry my-secret \
   -n connaisseur
 ```
 
-> :octicons-light-bulb-16: **Note**: At present, it [seems to be necessary](https://github.com/sigstore/cosign/issues/587#issuecomment-1062510930) to suffix your registry server URL with `/v1/`. This may become unnecessary in the future.
+!!! note
```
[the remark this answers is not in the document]
xopham: also not a `tip` though. will make it `info` as well

docs/basics.md
```diff
@@ -98,6 +99,32 @@ Install Connaisseur via Helm:
     helm install connaisseur connaisseur/connaisseur --atomic --create-namespace --namespace connaisseur -f values.yaml
     ```
 
+=== "Kubernetes manifests"
```
Starkteetje, line 102: Should the `### Get the code/chart` section also get a Kubernetes manifests content tab such that you can use the same content tab for everything?

helm/templates/alertconfig.yaml
Starkteetje: I mean, if I do `kubectl delete all -l app.kubernetes.io/component=core` I might very well delete something else that also has poor naming, which is both on them and us. Running `kubectl delete all -l app.kubernetes.io/component=connaisseur-core`, if I delete something else then it's totally on them, not on us

docs/basics.md
Starkteetje: I think it's inconsistent. Would vote for adding title="helm/values.yaml" everywhere and making the box title something like "Example configuration" or similar

helm/templates/alertconfig.yaml
xopham: hm...gotcha...think I would still question your abilities, but might be willing to agree and will adjust

docs/basics.md
xopham: thought about it, would be nice, but needs makefile support 🤷

docs/basics.md
xopham: then it becomes "title example, box example, ...". If inconsistency is the fear, will duplicate

docs/basics.md
```diff
@@ -98,6 +99,32 @@ Install Connaisseur via Helm:
...
+    ```
+    Next, the admission controller is deployed step-wise:
+
+    0. Create target namespace:
+    ```bash
+    kubectl create namespace connaisseur
+    ```
+    1. Setup the pre-hook:
+    ```bash
+    kubectl apply -f deploy.yaml -l 'app.kubernetes.io/component=prehook' -n connaisseur
```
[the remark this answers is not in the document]
xopham: not sure, did not investigate whether it needs to happen prior to core

docs/basics.md
xopham: thought about it, but don't see what to say there, as it is literally just doing the same as for `git` and in that case we need to rename the "git repo" part to "git repo (helm)" or something and we start changing more and more for minimal added consistency value

docs/features/README.md
xopham: indeed I did add a "such as" above and thus it is merely a list of "some featured features"

docs/basics.md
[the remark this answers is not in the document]
Starkteetje: Idk, it seems only the Webhook configuration for "dont do anything" is in there, which I think isn't even deployed during usual installations, only during deinstallation, but maybe I'm wrong 🤷‍♂️ 
But that only reinforces my feeling that the name is off

Starkteetje (approved):

PR #1414, opened 2023-11-30
test: fix local integration testing and add script for ease of use

annekebr:
Fixes #

## Checklist

### DIFF
--- a/tests/integration/Dockerfile.populate_notary
+++ b/tests/integration/Dockerfile.populate_notary
@@ -1,0 +1,16 @@
+FROM debian:bullseye@sha256:ab2b95aa8d7d6d54866b92c322cf0693933c1ae8038652f24ddfda1d1763a45a
+
+COPY ./tests/data/notary_service_container/ tests/data/notary_service_container/
+COPY ./tests/integration/notary_init.sh tests/integration/notary_init.sh
+COPY ./tests/integration/notary_addhash.sh tests/integration/notary_addhash.sh
+
+RUN apt update &&\
+    apt upgrade && \
+    apt install -y docker notary expect ca-certificates && \
+    cp ./tests/data/notary_service_container/server/ca.crt /usr/local/share/ca-certificates/notary_root_ca.crt && \
+    update-ca-certificates
+
+ARG DIGEST
+ENV DIGEST=${DIGEST}
+
+CMD [ "bash", "-c", "./tests/integration/notary_init.sh && ./tests/integration/notary_addhash.sh ${DIGEST}" ]
--- a/tests/integration/README.md
+++ b/tests/integration/README.md
@@ -1,76 +1,24 @@
 # Local Integration Test Setup
 
-## minikube
+Local integration testing should be easy if you're running a minikube cluster with the docker driver or a kind cluster. Just run
 
-To load the required docker images into the minikube docker daemon run
+`./tests/integration/run_integration_tests.sh -h`
 
-```shell
-eval $(minikube docker-env)
-make docker
-```
+from the root folder and start from there!
 
-which will change the environment variables that specify the docker daemon to point to minikube instead of your system docker daemon.
+## Example
 
-### Preparation for minikube with virtualbox driver
+If you have
 
-Assuming you have a running minikube cluster using the virtualbox driver set as current kubernetes context you just need to have the alerting interface running.
+- a running kind cluster and want to run the regular and the cosign integration test and you want to preserve the logs in failure case to `../integration_test_logs` you need to run
 
-Run the docker container that serves as alerting endpoint and retrieve the IP address that it has on the `bridge` network:
+        ./tests/integration/run_integration_tests.sh -c kind -r "regular cosign" -p ../integration_test_logs
 
-```shell
-docker run -d --name alerting-endpoint -p 56243:56243 docker.io/securesystemsengineering/alerting-endpoint:latest
-export ALERTING_ENDPOINT_IP=$(docker container inspect alerting-endpoint | jq -r .[].NetworkSettings.Networks.bridge.IPAddress)
-```
+    specifying the `-c` (cluster), `-r` (run-only) and the `-p` (preserve-logs) flags
 
-### Preparation for minikube with docker driver
+- a running minikube cluster and you want to run everything including the self-hosted-notary test but not the load and the complexity test you need to run
 
-Assuming you have a running minikube cluster using the docker driver set as current kubernetes context you need to have the alerting interface running and attach the container to the network that is used by minikube to facilitate communication. By default, this network is named `minikube`, so change the minikube network name to your specific name if you use a customized one:
 
-```shell
-MINIKUBE_NETWORK=$(docker container inspect minikube | jq -r '.[].NetworkSettings.Networks | to_entries | .[].key')
-docker run -d --name alerting-endpoint -p 56243:56243 docker.io/securesystemsengineering/alerting-endpoint:latest
-docker network connect ${MINIKUBE_NETWORK} alerting-endpoint
-export ALERTING_ENDPOINT_IP=$(docker container inspect alerting-endpoint | jq -r --arg minikube_network ${MINIKUBE_NETWORK} '.[].NetworkSettings.Networks[$minikube_network].IPAddress')
-```
+        ./tests/integration/run_integration_tests.sh -s "load complexity" -e
 
-### Actual test
-
-From the git repository folder run the `tests/integration/integration-test.sh` script with the name of the test you want to run, for example `regular`, `cosign` or `load` (see [the test script for all possible values](integration-test.sh)).
-
-To cleanup the mocked alerting interface container don't forget running
-
-```shell
-docker stop alerting-endpoint
-docker rm alerting-endpoint
-```
-
-## kind
-
-For kind we assume you have a running `kind` cluster set as current kubernetes context. To load the required docker image onto the `kind` nodes, run
-
-```shell
-make docker
-kind load docker-image $(yq e '.kubernetes.deployment.image.repository' helm/values.yaml):v$(yq e '.appVersion' helm/Chart.yaml)
-```
-
-You need to have the alerting interface running and attach it to the docker network that is used by the kind container just as for minikube using the docker driver. By default, it's name is `kind`, so if you renamed the docker network of the kind container, provide your custom name as `KIND_NETWORK`:
-
-```shell
-KIND_NETWORK=kind
-docker run -d --name alerting-endpoint -p 56243:56243 docker.io/securesystemsengineering/alerting-endpoint:latest
-docker network connect ${KIND_NETWORK} alerting-endpoint
-export ALERTING_ENDPOINT_IP=$(docker container inspect alerting-endpoint | jq -r --arg kind_network ${KIND_NETWORK} '.[].NetworkSettings.Networks[$kind_network].IPAddress')
-```
-
-From the git repository folder run the `tests/integration/integration-test.sh` script with the name of the test you want to run, for example `regular`, `cosign` or `load` (see [the test script for all possible values](integration-test.sh)).
-
-Obviously, we don't want to leave unused resources running, so stop and remove the alerting interface docker container as in the minikube case :-)
-
-```shell
-docker stop alerting-endpoint
-docker rm alerting-endpoint
-```
-
-## Note
-
-The integration test only works correctly if the alerting endpoint container is set up freshly as it counts all hits to its endpoints. So cleaning up is not only housekeeping, but also prevents issues caused by double-use of the container.
+    specifying the `-s` (skip) and the `-e` (extended) flag. As the cluster defaults to minkube, it's not necessary to provide the `-c` flag.
--- a/tests/integration/cleanup.sh
+++ b/tests/integration/cleanup.sh
@@ -1,0 +1,28 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+preserve_info_and_cleanup() {
+	kubectl config use-context ${DEFAULT_CTX}
+	echo "Preserving log files..."
+	(kubectl logs -n connaisseur -lapp.kubernetes.io/name=connaisseur --prefix=true --tail=-1 2>&1 || kubectl logs -n security -lapp.kubernetes.io/name=connaisseur --prefix=true --tail=-1 ) > connaisseur.log
+	cat helm/values.yaml > connaisseur.conf
+	( (kubectl describe pods -n connaisseur -lapp.kubernetes.io/name=connaisseur 2>&1 &&
+	  kubectl describe deployments.apps -n connaisseur -lapp.kubernetes.io/name=connaisseur 2>&1) ||
+	  (kubectl describe pods -n security -lapp.kubernetes.io/name=connaisseur &&
+      kubectl describe deployments.apps -n security -lapp.kubernetes.io/name=connaisseur)) > connaisseur.state
+    echo 'Cleaning up installation and test resources...'
+    make uninstall >/dev/null 2>&1 || true
+    helm uninstall connaisseur -n security >/dev/null 2>&1 || true
+    kubectl delete ns security >/dev/null 2>&1 || true
+    kubectl delete clusterrole auxillary >/dev/null 2>&1 || true
+    kubectl delete clusterrolebinding securityadmin >/dev/null 2>&1 || true
+    kubectl config delete-context deployconnaisseur >/dev/null 2>&1 || true
+    kubectl delete all,cronjobs,daemonsets,jobs,replicationcontrollers,statefulsets,namespaces -luse="connaisseur-integration-test" -A >/dev/null 2>&1 || true
+    kubectl delete ns connaisseur >/dev/null 2>&1 || true
+    rm ghcr-values ghcr-validator >/dev/null 2>&1 || true
+    rm release.yaml release_patched.yaml >/dev/null 2>&1 || true
+    rm root.json >/dev/null 2>&1 || true
+    mv values.yaml.backup helm/values.yaml >/dev/null 2>&1 || true
+    mv deployment.yaml.backup helm/templates/deployment.yaml >/dev/null 2>&1 || true
+    echo 'Finished cleanup of integration test leftovers.'
+}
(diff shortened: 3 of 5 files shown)


tests/integration/integration-test.sh
```diff
@@ -634,11 +657,4 @@ fi
...
 	exit $((${EXIT}))
-fi
-
-echo 'Cleaning up installation and test resources...'
-make uninstall >/dev/null 2>&1 || true
-kubectl delete all,cronjobs,daemonsets,jobs,replicationcontrollers,statefulsets,namespaces -luse="connaisseur-integration-test" -A >/dev/null
-rm ghcr-values ghcr-validator
-mv values.yaml.backup helm/values.yaml
-echo 'Finished cleanup.'
+fi
```
Starkteetje, line 660: nit: EOF missing newline

tests/integration/populate_notary.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+./tests/integration/notary_init.sh
+./tests/integration/notary_addhash.sh $1
```
Starkteetje, line 5: nit: EOF missing newline

tests/integration/populate_notary.sh
```diff
@@ -0,0 +1,5 @@
+#!/usr/bin/env bash
```
Starkteetje, line 1: Does this actually need a separate script? Just for reading notary_init && notary_addhash $X sounds similarly reasonable to populate_notary $X

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+
+IS_INTEGRATION_TEST_RUNNING="false"
+EXTENDED="false"
+
+cleanup() {
+    rv=$?
+    echo "Cleanup local resources..."
+    eval $(minikube docker-env --unset) >/dev/null 2>&1 || true
+    docker stop alerting-endpoint notary-signer notary-server >/dev/null  2>&1 || true
+    docker rm alerting-endpoint notary-signer notary-server >/dev/null 2>&1|| true
```
Starkteetje, line 23:
```suggestion
    docker rm alerting-endpoint notary-signer notary-server >/dev/null 2>&1 || true
```

tests/integration/integration-test.sh
```diff
@@ -362,7 +376,7 @@ update_for_self_hosted_notary() {
 	yq '. *+ load("ghcr-values")' -i update
 	yq eval-all --inplace 'select(fileIndex == 0) * select(fileIndex == 1)' helm/values.yaml update
 	rm update
-	curl "https://notary.server:4443/v2/docker.io/securesystemsengineering/testimage/_trust/tuf/root.json" > root.json
+	curl -k "https://${NOTARY_IP}:4443/v2/docker.io/securesystemsengineering/testimage/_trust/tuf/root.json" > root.json
 	SELF_HOSTED_NOTARY_PUBLIC_ROOT_KEY_ID=$(cat root.json | jq -r .signatures[0].keyid)
 	cat root.json | jq -r --arg KEYID $SELF_HOSTED_NOTARY_PUBLIC_ROOT_KEY_ID '.signed.keys | .[$KEYID] | .keyval.public' | base64 -d | openssl x509 -pubkey -noout > self_hosted_notary_root_key.pub
```
Starkteetje, line 381:
```suggestion
	cat root.json | jq -r --arg KEYID ${SELF_HOSTED_NOTARY_PUBLIC_ROOT_KEY_ID} '.signed.keys | .[$KEYID] | .keyval.public' | base64 -d | openssl x509 -pubkey -noout > self_hosted_notary_root_key.pub
```

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+    rm .hosts > /dev/null 2>&1 || true
```
Starkteetje, line 24:
```suggestion
    rm .hosts >/dev/null 2>&1 || true
```

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+
+help()
+{
+   # Display Help
+   echo "This script will run the integration test suite locally. By default, all integration tests besides of the 'self-hosted-notary' test will be run (addable with -e (extended) flag)."
+   echo
+   echo "Flag   Values                       Meaning         Description"
+   echo "s      space-separated list         skip            Provided integration tests will be skipped. Cannot be used with -r (run-only) flag."
+   echo "       of integration test names"
+   echo "       as stings"
```
Starkteetje, line 48:
```suggestion
   echo "       as strings"
```

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+   echo "r      space-separated list         run-only        Only the provided integrations tests will be run. Cannot be used with -s (skip) flag."
```
Starkteetje, line 49:
```suggestion
   echo "r      space-separated list         run-only        Only the provided integration tests will be run. Cannot be used with -s (skip) flag."
```

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+   echo "       of integration test names"
+   echo "       as stings"
```
Starkteetje, line 51:
```suggestion
   echo "       as strings"
```

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+   echo "e                                   extended        Include 'self-hosted-notary' test."
+   echo "c       <'minikube'|'kind'>         cluster         Specifies the cluster that should be used for integration testing. Must be either 'minikube' or 'kind'. (Default: 'minikube')"
+   echo "p       <path>                      preserve-logs   Must be a relative or absolute path. If set, log files of failed tests are copied that path."
```
Starkteetje, line 54:
```suggestion
   echo "p       <path>                      preserve-logs   Must be a relative or absolute path. If set, log files of failed tests are copied to that path."
```

Starkteetje: Also "Must be a relative or absolute path." is somewhat tautological and could be "Must be a (valid?) path"

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+done
+shift
+
+mkdir tmp
+
+export COSIGN_PUBLIC_KEY="-----BEGIN PUBLIC KEY-----<br>MFkwEwYHKoZIzj0CAQYIKoZIzj0DAQcDQgAEQ70fJ/NA109p3/43cTgwnLQ+HsvK<br>jFK/0L1kLUQO+e2cL+3+y2s4IykwTxfSnqOSA0kEy1rTrhrXLCox4uRmSQ==<br>-----END PUBLIC KEY-----"
+
+if [[ "${CLUSTER:-"minikube"}" == "minikube" ]]; then
+  if [[ "$(docker inspect minikube > /dev/null 2>&1 | jq -r .[].State.Status || echo 'container not found')" != "running" ]]; then
+    echo "Minikube cluster configured. You need a running minikube cluster. Exiting..."
```
Starkteetje, line 102: Is there a 'no(t)' missing?

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+    exit 1
+  fi
+  eval $(minikube docker-env)
+  echo "Building docker image..."
+  make docker > /dev/null
+  echo "Done building docker image!"
+  NETWORK=$(docker container inspect minikube | jq -r '.[].NetworkSettings.Networks | to_entries | .[].key')
+else
+  if [[ "$(docker inspect kind-control-plane > /dev/null 2>&1 | jq -r .[].State.Status || echo 'container not found')" != "running" ]]; then
+    echo "Kind cluster configured. You need a running kind cluster. Exiting..."
```
Starkteetje, line 112: Is there a 'no(t)' missing?

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+
+readarray -t FULL_INTEGRATION_TEST_SUITE < <(yq e -r -o=j '.jobs.[].strategy.matrix | select(.integration-test-arg) | .integration-test-arg[]' .github/workflows/.reusable-integration-test.yml)
+
+FAILED_INTEGRATION_TESTS=()
+SUCCESSFUL_INTEGRATION_TESTS=()
+SCRIPT_FAILURES=()
+
+INTEGRATION_TEST_SUITE=( "${INTEGRATION_TESTS_TO_RUN[@]:-${FULL_INTEGRATION_TEST_SUITE[@]}}" )
+
+if [[ ${EXTENDED} == "true" || "${INTEGRATION_TESTS_TO_RUN[@]}" =~ "self-hosted-notary" ]]; then
```
Starkteetje, line 136: Not that it really makes a difference, but semantically, wouldn't it be
```suggestion
if [[ ${EXTENDED} == "true" || "${INTEGRATION_TEST_SUITE[@]}" =~ "self-hosted-notary" ]]; then
```
?

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+else
+  INTEGRATION_TESTS_TO_SKIP+=("self-hosted-notary")
+fi
+
+for INTEGRATION_TEST in ${INTEGRATION_TEST_SUITE[@]}; do
+  if [[ "${INTEGRATION_TESTS_TO_SKIP[@]}" =~ "${INTEGRATION_TEST}" ]]; then
+    echo "Skipping integration test ${INTEGRATION_TEST}"
+  else
+    echo "Running integration test ${INTEGRATION_TEST}..."
+    mkdir ./tmp/${INTEGRATION_TEST}
```
Starkteetje, line 156: super nit as with spaces probably doesn't work due to array, but 
```suggestion
    mkdir "./tmp/${INTEGRATION_TEST}"
```
should generally be safer. Same below

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+    IS_INTEGRATION_TEST_RUNNING="true" #this will be wrong when there is a failure with tee because in that case the integration test won't be started
+    ./tests/integration/integration-test.sh ${INTEGRATION_TEST} | tee ./tmp/${INTEGRATION_TEST}/test.log || true
```
Starkteetje, line 158:
```suggestion
    ./tests/integration/integration-test.sh ${INTEGRATION_TEST} &| tee ./tmp/${INTEGRATION_TEST}/test.log || true
```
?

tests/integration/run_integration_tests.sh
```diff
@@ -0,0 +1,222 @@
...
+  echo -e "\n\n"
+
+  if [[ ${DESTINATION:-""} != "" ]]; then
+    echo -e "Log files for failed integration tests were saved to '${DESTINATION}'."
+  fi
+fi
+
+
+
+echo -e "\n\n\n#############################################\n\n\n"
```
Starkteetje, line 222: nit: EOF missing newline

tests/integration/Dockerfile.populate_notary
```diff
@@ -0,0 +1,18 @@
+FROM debian:bullseye@sha256:ab2b95aa8d7d6d54866b92c322cf0693933c1ae8038652f24ddfda1d1763a45a
+
+ARG DIGEST
+ENV DIGEST=${DIGEST}
+
+COPY ./tests/data/notary_service_container/ tests/data/notary_service_container/
```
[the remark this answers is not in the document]
annekebr: That's because I wanted to reuse the existing scripts, pointing to config files at that path

tests/integration/cleanup.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+preserve_info_and_cleanup() {
+	kubectl config use-context ${DEFAULT_CTX}
+	echo "Preserve log files..."
+	(kubectl logs -n connaisseur -lapp.kubernetes.io/name=connaisseur --prefix=true --tail=-1 2>&1|| kubectl logs -n security -lapp.kubernetes.io/name=connaisseur --prefix=true --tail=-1 ) > connaisseur.log
```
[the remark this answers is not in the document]
annekebr: The other NS is `security` which is the part behing the logical OR. Why do you think this breaks?

tests/integration/run_integration_tests.sh
annekebr: No that's to tell user that what cluster is actually configured. Especially if they hav a kind cluster running, but forgot to provide -c flag, the default here is minikube and the message will be `Minkube configured`. Should I reword that somehow?

tests/integration/run_integration_tests.sh
annekebr: see above.

tests/integration/run_integration_tests.sh
annekebr: Nope, because that should evaluate to true only if the user specifically provided `self-hosted-notary` via `-r`  flag.

annekebr: Do you dislike that default?

tests/integration/run_integration_tests.sh
annekebr: What does it do?

tests/integration/cleanup.sh
```diff
@@ -0,0 +1,24 @@
+#!/usr/bin/env bash
+set -euo pipefail
+
+preserve_info_and_cleanup() {
+	kubectl config use-context ${DEFAULT_CTX}
+	echo "Preserve log files..."
```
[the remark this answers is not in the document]
annekebr: I don't understand that comment 🙊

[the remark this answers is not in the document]
Starkteetje: One is time form present, the other is gerund, so either Preserving and Cleaning or Preserve and Clean

tests/integration/cleanup.sh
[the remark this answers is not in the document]
Starkteetje: You're right, no idea why I missed the || part 🤷

tests/integration/run_integration_tests.sh
Starkteetje: Ah, that's "You configured to run the tests on Minikube, but there's no running Minikube." I read "You have a configured Minikube cluster, but you need a running one". My head would be glad if there were a rewording 😁

tests/integration/cleanup.sh
```diff
@@ -0,0 +1,24 @@
...
+	(kubectl logs -n connaisseur -lapp.kubernetes.io/name=connaisseur --prefix=true --tail=-1 2>&1|| kubectl logs -n security -lapp.kubernetes.io/name=connaisseur --prefix=true --tail=-1 ) > connaisseur.log
+	cat helm/values.yaml > connaisseur.conf
+	( (kubectl describe pods -n connaisseur -lapp.kubernetes.io/name=connaisseur 2>&1 &&
+	  kubectl describe deployments.apps -n connaisseur -lapp.kubernetes.io/name=connaisseur 2>&1) ||
+	  (kubectl describe pods -n security -lapp.kubernetes.io/name=connaisseur &&
+      kubectl describe deployments.apps -n security -lapp.kubernetes.io/name=connaisseur)) > connaisseur.state
+    echo 'Cleaning up installation and test resources...'
+    make uninstall >/dev/null 2>&1|| true
+    helm uninstall connaisseur -n security >/dev/null 2>&1 && kubectl delete ns security && kubectl delete clusterrole auxillary && kubectl delete clusterrolebinding securityadmin && kubectl config delete-context deployconnaisseur || true
```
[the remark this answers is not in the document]
annekebr: First question: At least that was what I wanted, might be stupid 🤷‍♀️ 
Second question: yes
Third question: It seemed to me that helm uninstall does sometimes not work if some resources were cleaned up previously manually (not 100% sure). These ANDs would at least guarantee, that on a second try everything will be cleaned up (at least that was the reasoning behind it)

tests/integration/run_integration_tests.sh
Starkteetje: I don't really care. To me it just looked like `self-hosted-notary` wouldn't be part of the `FULL_INTEGRATION_TEST_SUITE` default anyway, so this way even if we decided to add it to defaults in the CI, this script would be more opinionated on it definitely not being default, which I'm totally fine with, just wasn't sure whether it was intentional

tests/integration/run_integration_tests.sh
Starkteetje: Pipe stderr and stdin through the pipe instead of just stdin, which would catch any missing 2>&1 in the other script

tests/integration/cleanup.sh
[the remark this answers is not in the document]
Starkteetje: To 3) But they wouldn't, would they? If something fails during uninstall, then the remaining commands won't be executed, so the won't be "fallback" cleanups. Basically, what I'm asking whether X && Y || true should be 
X || true
Y || true

tests/integration/README.md
```diff
@@ -1,76 +1,7 @@
...
-
-```shell
-docker stop alerting-endpoint
-docker rm alerting-endpoint
-```
-
-## Note
-
-The integration test only works correctly if the alerting endpoint container is set up freshly as it counts all hits to its endpoints. So cleaning up is not only housekeeping, but also prevents issues caused by double-use of the container.
+from the root folder and start from there!
```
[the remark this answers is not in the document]
annekebr: Added two examples now - you like it?

tests/integration/run_integration_tests.sh
annekebr: Oh, oops, I thought thats `|&`  😁

But yep, I think it makes sense 👍

tests/integration/Dockerfile.populate_notary
```diff
@@ -0,0 +1,16 @@
...
+RUN apt update &&\
+    apt upgrade && \
+    apt install -y docker notary expect ca-certificates && \
+    cp ./tests/data/notary_service_container/server/ca.crt /usr/local/share/ca-certificates/notary_root_ca.crt && \
+    update-ca-certificates
+
+ARG DIGEST
+ENV DIGEST=${DIGEST}
+
+CMD [ "bash", "-c", "./tests/integration/notary_init.sh && ./tests/integration/notary_addhash.sh ${DIGEST}" ]
```
Starkteetje, line 16: nit: EOF missing newline

Starkteetje (approved):

Starkteetje (approved):

PR #1429, opened 2023-12-12
update: Update k8s image registry in default policy

Starkteetje:
Context: https://kubernetes.io/blog/2023/03/10/image-registry-redirect/

## Checklist


docs/basics.md
```diff
@@ -494,7 +494,7 @@ This is likely the most common case in simple settings by which only self-built
 
       policy:
       - pattern: "*:*"
-      - pattern: "k8s.gcr.io/*:*"
+      - pattern: "registry.k8s.io/*:*"
```
annekebr, line 497: How comes the former did work until now? Assuming there were no problems for our users as we got no issues, I wonder whether the old k8s versions still use the old registry in their references and that under the hood only the old registry is redirected to the new one. I assume, we don't have a good integration test for this or is this covered by anything?

Starkteetje: The old registry is indeed redirecting to the new one according to the docs, which is probably why it still works. The redirect will be shut down in January though 😬 
But no, afaik there's no test for it

PR #1467, opened 2024-01-19
fix: no exceptions on automatic child approval

phbelitz:
When automatic child approval is active and can't find the parent resource, even though there is one, the validation should continue, as if there were no parent resources in the first place. This prevents some deployments from failing, where the parent resource was deleted and now the child can never be validated again.

## Checklist


connaisseur/workload_object.py
```diff
@@ -75,18 +78,19 @@ def parent_containers(self):
...
 
             if parent["metadata"]["uid"] != uid:
-                msg = (
-                    "Couldn't find the right parent"
-                    " resource {parent_kind} {parent_name}."
-                )
-                raise ParentNotFoundError(
-                    message=msg, parent_kind=kinds, parent_name=name, parent_uid=uid
-                )
+                msg = f"Couldn't find the right parent" f" resource {kinds} {name}."
```
Starkteetje, line 91: Why not
```suggestion
                msg = f"Couldn't find the right parent resource {kinds} {name}."
```

Starkteetje (approved):

PR #1599, opened 2024-05-03
feat: Configurable cache expiry

### LINKED ISSUE #1587: enable redis as optional feature
**Describe the feature**
Since 3.4.0 the a redis is shipped by default and can not be disabled. 
In my opinion it should be possible to disable the redis completely or at least use an external redis.

The problem is that all cache/db services in my cluster must run cluster external (as a service).

Nothing fancy in helm but a few values to enable/disable the feature.

Starkteetje:
Fixes #1587 in its basic form, though it doesn't allow configuring a remote Redis

## Description
This PR makes the cache expiry time of entries in the Redis configurable. If cache expiry is set to 0, the Redis deployment is cut from Connaisseur's deployment.

## Checklist

### DIFF
--- a/docs/features/caching.md
+++ b/docs/features/caching.md
@@ -3,4 +3,24 @@
 Connaisseur utilizes [Redis](https://github.com/redis/redis) as a cache.
 For each image reference the resolved digest or validation error is cached.
 This drastically boosts the performance of Connaisseur compared to older non-caching variants.
-The expiration for keys in the cache is set to 30 seconds.
+The expiration for keys in the cache defaults to 30 seconds, but can be tweaked.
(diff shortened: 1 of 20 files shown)


docs/features/caching.md
```diff
@@ -3,4 +3,24 @@
...
+
+## Example
+
+In `charts/connaisseur/values.yaml`:
+
+```yaml
+application:
+  features:
+    cache:
+      expirySeconds: true
```
phbelitz, line 25: `true`?

Starkteetje: When you copy example config from testing behaviour in case of misconfig...Good catch

phbelitz (approved):

PR #1609, opened 2024-05-10
fix: Fix handling of undefined values in values.yaml

Starkteetje:
## Description

Before this commit, there were different cases where Connaisseur mishandled default values in its config, if their parent keys were not set in the values.yaml. Resulting errors could've been anything between inconsequential (additional log entry) and fatal (e.g. missing deployment of Role). This commit fixes handling of these values.

## Checklist

### DIFF
--- a/charts/connaisseur/templates/env.yaml
+++ b/charts/connaisseur/templates/env.yaml
@@ -6,17 +6,15 @@ metadata:
   labels:
     {{- include "connaisseur.labels" . | nindent 4 }}
 data:
-  {{- with .Values.application.features }}
-  AUTOMATIC_CHILD_APPROVAL: {{ print .automaticChildApproval | default "true" | quote }}
-  AUTOMATIC_UNCHANGED_APPROVAL: {{ print .automaticUnchangedApproval | default "false" | quote }}
-  DETECTION_MODE: {{ print .detectionMode | default "false" | quote }}
-  RESOURCE_VALIDATION_MODE: {{ print .resourceValidationMode | default "all" | quote }}
-  {{- with .cache }}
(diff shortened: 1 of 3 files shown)


charts/connaisseur/templates/env.yaml
```diff
@@ -7,13 +7,13 @@ metadata:
     {{- include "connaisseur.labels" . | nindent 4 }}
 data:
   {{- with .Values.application.features }}
-  AUTOMATIC_CHILD_APPROVAL: {{ print .automaticChildApproval | default "true" | quote }}
-  AUTOMATIC_UNCHANGED_APPROVAL: {{ print .automaticUnchangedApproval | default "false" | quote }}
-  DETECTION_MODE: {{ print .detectionMode | default "false" | quote }}
-  RESOURCE_VALIDATION_MODE: {{ print .resourceValidationMode | default "all" | quote }}
+  AUTOMATIC_CHILD_APPROVAL: {{ .automaticChildApproval | default "true" | quote }}
```
phbelitz, line 10: are you sure this works? if `.automaticChildApproval` is set to `false` (the bool, not the string), then the default clause will trigger and will set the value to `"true"` eventhough you wanted it the be set to false.

Starkteetje: You're totally right that this introduces another set of mishandlings :(
Without print and if set to false, it's seen as falsy, thus default is applied. If not set but with print, it prints `<nil>`, so also wrong. I now replaced it with dig.
Something like a custom helm function for this would be nice ([dig doesn't work with .Values](https://github.com/helm/helm/issues/11796) unfortunately, but there are workaround one could put into a function like `{{ omit .Values | dig "a" "b" true }}`), but it doesn't look possible if one doesn't want to use a dict for input

phbelitz (approved):

PR #1610, opened 2024-05-10
update: dependabot ignore golang update

phbelitz:
Dependabot should not update the golang version, since this will break Connaisseur quite easily.

## Checklist


Starkteetje (approved):

phbelitz:
let's rather do updates here then ignoring the changes

PR #1630, opened 2024-05-24
fix: remove startup probe

phbelitz:
In some cases the Redis pod isn't reachable by Connaisseur, which results in Connaisseur never starting up even though it could perform its task even without the Redis pod. Thus the startup probe for Connaisseur, trying to reach the Redis pod was removed.

## Checklist


internal/handler/validation/skip.go
```diff
@@ -104,6 +104,10 @@ func getCachedDigest(
 ) (cachedDigest string, cachedError, err error) {
 	cache := ctx.Value(constants.Cache).(caching.Cacher)
 
+	if cache.Ping(ctx) != nil {
```
Starkteetje, line 107: Doesn't this just add latency in the usual case where everything works as intended? If cache isnt up, cache.Get will fail anyways and we have the same behaviour (though with different error)

phbelitz: the thing is, i want to a a regular warning message that says the cahc can't be reached. currently, if the `cache.Get` doesn't go though, the error is only logged on debug level (which is intended). only warn if the ping didn't work once during initialization wont work, as connaisseur starts up faster then redis ...

compromise would be: I only ping when the Get returned an error. then i either had a cache miss, or redis wasn't reachable in the first place, so i recheck with the ping which one it was, okey?

Starkteetje: Sounds better 👍 
However, wouldn't the error message already give away whether the connection worked? i.e. couldn't you parse that instead of doing another roundtrip upon failure?

internal/handler/validation/skip.go
```diff
@@ -106,7 +106,11 @@ func getCachedDigest(
 
 	val, err := cache.Get(ctx, img.OriginalString())
 	if err != nil {
-		return "", nil, fmt.Errorf("cache miss for image %s: %s", img.OriginalString(), err)
+		if cache.Ping(ctx) != nil {
+			return "", nil, fmt.Errorf("error pinging cache")
```
Starkteetje, line 110:
```suggestion
			return "", nil, fmt.Errorf("error reaching cache")
```
maybe?

internal/handler/validation/skip.go
Starkteetje: If you only want to have debug level logging of that error anyway, you could also say "if debug, then ping", which would also make prod deployments not waste a roundtrip

Starkteetje (approved):

PR #1659, opened 2024-06-14
feat: keyless

### LINKED ISSUE #1493: Add keyless support (also with private fulcio/rekor instance)
It would be great to have working keyless support in connaisseur.

We made a strong effort to establish a system to rollout your own instance of fulcio and rekor (look at <https://gitlab.opencode.de/ig-bvc/ag-sig/fulcio-rekor-rollout>). This is a project for european public agencies to share secure container.

I've tried to write down the assets to verify images with cosign (<https://gitlab.opencode.de/ig-bvc/ag-sig/fulcio-rekor-rollout/-/blob/main/WhatIsNeededForVerify.md>).

I've seen the principal hooks inside connaisseurs are there but empty (not implemented yet).

Can we start a discussion to implement it? There's example code for signing and verification in the repo above.

Thanks in advance

peter <peter@pflaeging.net>

phbelitz:
Adds support for keyless cosign verification. This feature allows users to verify images signed with cosign without the need for a public key.

fixes #1493

## Checklist


phbelitz:
Integration tests and documentation are still missing.

internal/utils/validation.go
```diff
@@ -53,3 +54,11 @@ func Validate(s interface{}) error {
 
 	return fmt.Errorf("%s has %d errors:\n%s", ty, len(err.(validation.ValidationErrors)), msg)
 }
+
+func ValidateCertificate(cert string) error {
+	certPool := x509.NewCertPool()
+	if !certPool.AppendCertsFromPEM([]byte(cert)) {
```
Starkteetje, line 60: That's not really what this function is supposed to check. 

> // AppendCertsFromPEM attempts to parse a series of PEM encoded certificates.
// It appends any certificates found to s and reports whether any certificates
// were successfully parsed.

If there is a single valid certificate inside the bytestring and then a bunch of garbage, this succeeds

Also a nit: the return value is never used in its form and could be a bool with as `IsValidCert` or similar
```suggestion
	return certPool.AppendCertsFromPEM([]byte(cert))
```

internal/validator/auth/trust_root.go
```diff
@@ -7,9 +7,18 @@ type TrustRoot struct {
...
+	Cert string `yaml:"cert" validate:"required_without_all=Key Keyless,excluded_with_all=Key Keyless"`
+	// keyless configuration
+	Keyless Keyless `yaml:"keyless" validate:"required_without_all=Key Cert,excluded_with_all=Key Cert"`
+}
+
+type Keyless struct {
+	// issuer of the trust root (e.g. a mail address)
+	Issuer string `yaml:"issuer"`
+	// provider of the trust root (e.g. an oidc provider)
+	Subject string `yaml:"subject"`
```
Starkteetje, line 21: Isn't this the exact opposite of what you implemented?

internal/config/config_test.go
```diff
@@ -478,7 +478,7 @@ func TestValidateErrors(t *testing.T) {
 					},
 				},
 			},
-			"Key must be set if Cert isn't",
+			"Key is a required field",
```
Starkteetje, line 481: That's not true, though?

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -106,16 +115,20 @@ func (cv *CosignValidator) UnmarshalYAML(unmarshal func(interface{}) error) erro
...
+	if valData.Host.FulcioCert != "" && utils.ValidateCertificate(valData.Host.FulcioCert) != nil {
+		return fmt.Errorf("invalid fulcio certificate for validator %s", valData.Name)
 	}
 
 	cv.Name = valData.Name
 	cv.Type = valData.Type
 	cv.Rekor = valData.Host.Rekor
+	cv.RekorPubkey = valData.Host.RekorPubkey
+	cv.FulcioCert = valData.Host.FulcioCert
+	cv.CTLogPubkey = valData.Host.CTLogPubkey
```
Starkteetje, line 131: Is there no way to validate the pub keys?

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -275,15 +314,15 @@ func (cv *CosignValidator) setupOptions(
...
-	} else {
-		ignoreTlog = !*args.VerifyTLog
+	getDefault := func(b *bool) bool {
+		if b == nil {
+			return false
+		}
+		return !*b
 	}
-	opts := &cosign.CheckOpts{SigVerifier: verifier, IgnoreTlog: ignoreTlog}
+	ignoreTlog, ignoreSCT := getDefault(args.VerifyTLog), getDefault(args.VerifySCT)
```
Starkteetje, line 324: A function `getDefault` returns the negated bool of its input (even ignoring the nil stuff)? 🤨 

Maybe along the lines of
```
		boolOrDefault := func(b *bool, default bool) bool {
		if b == nil {
			return default
		}
		return b
		
		verifyTlog,verifySCT := boolOrDefault(args.VerifyTLog, true), boolOrDefault(args.VerifySCT, true)
                opts := &cosign.CheckOpts{IgnoreTlog: !verifyTlog, IgnoreSCT: !verifySCT}
```
?
That has reasonable naming and would work for `false` defaults, too. If you don't like the premature generalization, you could also make it `boolOrTrue` and rm the default param

Starkteetje: As a nice benefit, this also negates all the `if !ignoreX` below back to `if verifyX`

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -316,74 +355,103 @@ func (cv *CosignValidator) setupOptions(
...
+		return &new_pubs, nil
+	}
+	_ = os.Setenv("SIGSTORE_NO_CACHE", "1")
+	return cosign.GetCTLogPubs(ctx)
+}
+
+func (cv *CosignValidator) getFulcioCerts() (*x509.CertPool, *x509.CertPool, error) {
+	if cv.FulcioCert != "" {
+		root := x509.NewCertPool()
+		// certificate was already validated during unmarshalling
```
Starkteetje, line 442: Thinking about it, why do we even put the bytestring into the CosignValidator in the first place? It is unserialized into a CosignValidatorYaml, so we could make the cv.Cert and cv.FulcioCert field the cert pool with the added certificate during validation already, or am I missing something?

Starkteetje (commented):
I don't have the mental capacity today to review this fully, but here are some preliminary comments

internal/validator/cosignvalidator/cosign_validator.go
phbelitz: yea. i changed it.

internal/config/config_test.go
phbelitz: changed the text

internal/validator/cosignvalidator/cosign_validator.go
phbelitz: added tests for rekor pub key and ctlog pubkey.

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -315,53 +378,26 @@ func (cv *CosignValidator) setupOptions(
...
-		if err != nil { // Currently cannot happen as it only fails for invalid URLs, which are already caught during unmarshalling
-			return nil, fmt.Errorf("unable to create Rekor client for %s: %s", cv.Rekor, err)
-		}
-		opts.RekorClient = rekorClient
-
-		// env var is needed, otherwise cosign tries to create files in readonly FS, which will fail
-		_ = os.Setenv("SIGSTORE_NO_CACHE", "1")
-		pubs, err := cosign.GetRekorPubs(ctx)
+	// set rekor options
+	if cv.Rekor != nil && verifyTlog {
```
Starkteetje, line 382: cv.Rekor can never be nil, can it?

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -35,10 +34,16 @@ type CosignValidator struct {
...
-	// url of the Rekor instance
-	Rekor string `validate:"url"`
+	// Rekor instance
+	Rekor *client.Rekor
+	// public key of the Rekor instance
+	RekorPubkey *cosign.TrustedTransparencyLogPubKeys
+	// certificate of the fulcio instance
+	FulcioCert *x509.CertPool
+	// public key of the CT log
+	CTLogPubkey *cosign.TrustedTransparencyLogPubKeys
```
Starkteetje, line 44: You only ever use this in a private function, so it should be private. Also, the description isn't true, see above.

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -146,61 +186,82 @@ func (cv *CosignValidator) ValidateImage(
...
+				verifierOutputChannel <- verifierOutput{nil, nil, err, ""}
 				return
 			}
-			verifierOutputChannel <- verifierOutput{validSignatures, nil, verifier.Name}
+			verifierOutputChannel <- verifierOutput{validSignatures, nil, nil, verifier.Name}
 		}(verifier, verifierOutputChannel)
 	}
 	validatingTrustRoots := []string{}
 	checkedSignatures := []oci.Signature{}
+	propagateErrors := []error{}
```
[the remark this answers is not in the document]
phbelitz: if the image is signed by a different identity then requested, the error would say `no signed digest`, which isn't really helpful

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -35,10 +34,16 @@ type CosignValidator struct {
...
 	// type of the validator (will always be "cosign")
 	Type string `validate:"eq=cosign"`
-	// url of the Rekor instance
-	Rekor string `validate:"url"`
+	// Rekor instance
+	Rekor *client.Rekor
+	// public key of the Rekor instance
+	RekorPubkey *cosign.TrustedTransparencyLogPubKeys
+	// certificate of the fulcio instance
+	FulcioCert *x509.CertPool
```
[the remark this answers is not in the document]
phbelitz: i think i will keep it the way it is, so that connaisseur will always pull "fresh" certificates, should they change while connaisseur is installed somewhere. then there is no need to reinstall.

docs/validators/sigstore_cosign.md
```diff
@@ -92,8 +92,13 @@ kubectl run altsigned --image=docker.io/securesystemsengineering/testimage:co-si
...
+| `trustRoots[*].keyless.subject` | - | :heavy_check_mark: if not using a key | Name of the identity that created the keyless signature. Usually an email address. This is a regex, so either the full email address can be used or only parts of it (e.g. `your.mail@example.org` or `your.mail`). |
 | `host.rekor` | `rekor.sigstore.dev` | - | Rekor URL to use for validation against the transparency log (default sigstore instance is `rekor.sigstore.dev`). Setting `host` enforces successful transparency log check to pass verification. See additional notes [below](#transparency-log-verification). |
+| `host.rekorPubkey` | Public key of `rekor.sigstore.dev` | - | Public key used to verify signature of log entries from Rekor. If none is given the public key for `rekor.sigstore.dev` is being retrieved. |
+| `host.fulcioCert` | Root and intermediate certificates belonging to `fulcio.sigstore.dev` | - | The root certificate belonging the fulcio CA which was used to create keyless signatures. |
```
Starkteetje, line 100: Why certiface**s** in the default when the var name is singular?

docs/validators/sigstore_cosign.md
```diff
@@ -92,8 +92,13 @@ kubectl run altsigned --image=docker.io/securesystemsengineering/testimage:co-si
...
+| `trustRoots[*].keyless.Issuer` | - | :heavy_check_mark: if not using a key | Name of the OIDC provider which attests the identity. This is a regex, so either the full url of the OIDC provider can be used or only parts of it. |
+| `trustRoots[*].keyless.subject` | - | :heavy_check_mark: if not using a key | Name of the identity that created the keyless signature. Usually an email address. This is a regex, so either the full email address can be used or only parts of it (e.g. `your.mail@example.org` or `your.mail`). |
 | `host.rekor` | `rekor.sigstore.dev` | - | Rekor URL to use for validation against the transparency log (default sigstore instance is `rekor.sigstore.dev`). Setting `host` enforces successful transparency log check to pass verification. See additional notes [below](#transparency-log-verification). |
+| `host.rekorPubkey` | Public key of `rekor.sigstore.dev` | - | Public key used to verify signature of log entries from Rekor. If none is given the public key for `rekor.sigstore.dev` is being retrieved. |
```
Starkteetje, line 99:
```suggestion
| `host.rekorPubkey` | Public key of `rekor.sigstore.dev` | - | Public key used to verify signature of log entries from Rekor. |
```
That's covered by the default field

docs/validators/sigstore_cosign.md
```diff
@@ -92,8 +92,13 @@ kubectl run altsigned --image=docker.io/securesystemsengineering/testimage:co-si
...
+| `host.fulcioCert` | Root and intermediate certificates belonging to `fulcio.sigstore.dev` | - | The root certificate belonging the fulcio CA which was used to create keyless signatures. |
+| `host.ctLogPubkey` | Public keys provided by sigstore | - | The public keys needed for verifying Signed Certificate Timestamps (SCT). |
```
Starkteetje, line 101:
```suggestion
| `host.ctLogPubkey` | Public keys for the certificate transparency log provided by Sigstore | - | The public keys needed for verifying Signed Certificate Timestamps (SCT). |
```
?

docs/validators/sigstore_cosign.md
```diff
@@ -349,4 +355,37 @@ To disable this feature the `with.verifyInTransparencyLog` key can be set to `fa
 
 ### Keyless signatures
 
-Keyless signatures have not yet been implemented but are planned in upcoming releases.
+Keyless signatures are a feature of sigstore that allows to sign container images without the need to manage a private key. Instead the signatures are bound to identities, attested by OIDC providers, and use ephemeral keys, short-lived certificates and a transparency log under the hood to provide the same security guarantees. Further information on this topic can be found [here](https://docs.sigstore.dev/signing/overview/).
```
Starkteetje, line 358:
```suggestion
Keyless signatures are a feature of Sigstore that allows to sign container images without the need to manage a private key. Instead the signatures are bound to identities, attested by OIDC providers, and use ephemeral keys, short-lived certificates and a transparency log under the hood to provide similar security guarantees. Further information on this topic can be found [here](https://docs.sigstore.dev/signing/overview/).
```

docs/validators/sigstore_cosign.md
```diff
@@ -349,4 +355,37 @@ To disable this feature the `with.verifyInTransparencyLog` key can be set to `fa
...
+- name: keylessvalidator
+  type: cosign
+  trustRoots:
+  - name: keyless
+    keyless:
+      issuer: "github"
+      subject: "philipp.belitz@securesystems.de
+```
+
+In case the signature was created using the sigstore infrastructure, nothing else needs to be configured since Connaisseur will automatically retrieve all needed public keys and certificates from the sigstore infrastructure. If the signature was created using a private infrastructure, the `host.fulcioCert` field can be used to specify the root certificate belonging to the fulcio CA which was used to create the keyless signatures. The `host.fulcioCert` field should contain the root certificate in PEM format. The same applies to the `host.ctLogPubkey` field which can be used to specify the public keys needed for verifying Signed Certificate Timestamps (SCT) and the `host.rekorPubkey` field which can be used to specify the public key used to verify the signature of log entries from Rekor.
```
Starkteetje, line 372:
```suggestion
In case the signature was created using the Sigstore infrastructure, nothing else needs to be configured since Connaisseur will automatically retrieve all needed public keys and certificates from the Sigstore infrastructure. If the signature was created using a private infrastructure, the `host.fulcioCert` field can be used to specify the root certificate belonging to the Fulcio CA which was used to create the keyless signatures. The `host.fulcioCert` field should contain the root certificate in PEM format. The same applies to the `host.ctLogPubkey` field which can be used to specify the public keys needed for verifying Signed Certificate Timestamps (SCT) and the `host.rekorPubkey` field which can be used to specify the public key used to verify the signature of log entries from Rekor.
```

internal/validator/auth/trust_root.go
```diff
@@ -7,9 +7,18 @@ type TrustRoot struct {
...
+	Cert string `yaml:"cert" validate:"required_without_all=Key Keyless,excluded_with_all=Key Keyless"`
+	// keyless configuration
+	Keyless Keyless `yaml:"keyless" validate:"required_without_all=Key Cert,excluded_with_all=Key Cert"`
+}
+
+type Keyless struct {
+	// issuer of the trust root (e.g. an oidc provider)
+	Issuer string `yaml:"issuer"`
+	// subject of the trust root (e.g. a mail address)
+	Subject string `yaml:"subject"`
```
Starkteetje, line 21: If these are regexes here, it should be noted in their description

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -35,10 +33,16 @@ type CosignValidator struct {
 	Name string `validate:"required"`
 	// type of the validator (will always be "cosign")
 	Type string `validate:"eq=cosign"`
-	// url of the Rekor instance
-	Rekor string `validate:"url"`
+	// Rekor instance
```
Starkteetje, line 36:
```suggestion
	// client for connecting to the Rekor instance
```

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -35,10 +33,16 @@ type CosignValidator struct {
...
+	Rekor *client.Rekor
+	// public key of the Rekor instance or nil if using the sigstore one
+	rekorPubkey *cosign.TrustedTransparencyLogPubKeys
+	// certificate of the fulcio instance or nil if using the sigstore one
+	fulcioCert *x509.CertPool
+	// public key of the CT log or nil if using the sigstore one
```
Starkteetje, line 42: nit
```suggestion
	// public key of the Rekor instance or nil if using the Sigstore one
	rekorPubkey *cosign.TrustedTransparencyLogPubKeys
	// certificate of the Fulcio instance or nil if using the Sigstore one
	fulcioCert *x509.CertPool
	// public key of the CT log or nil if using the Sigstore one
```
or lower-case Rekor, but this way it's inconsistent

docs/validators/sigstore_cosign.md
phbelitz: Because the value you actually enter is singular -> the root certificate.

Starkteetje: So the default is Root + intermediates, but a user could only specify root (which in practice would be the intermediate then)? Kinda weird, but I guess that works

internal/validator/auth/trust_root.go
```diff
@@ -19,6 +19,10 @@ type Keyless struct {
 	Issuer string `yaml:"issuer"`
 	// subject of the trust root (e.g. a mail address)
 	Subject string `yaml:"subject"`
+	// issuer regex of the trust root (e.g. an oidc provider)
+	IssuerRegex string `yaml:"issuerRegex"`
```
Starkteetje, line 23: I think we should validate that either issuer or issuesRegex is set, not both. Same for subject

docs/validators/sigstore_cosign.md
```diff
@@ -92,13 +92,13 @@ kubectl run altsigned --image=docker.io/securesystemsengineering/testimage:co-si
...
-| `trustRoots[*].keyless.subject` | - | :heavy_check_mark: if not using a key | Name of the identity that created the keyless signature. Usually an email address. This is a regex, so either the full email address can be used or only parts of it (e.g. `your.mail@example.org` or `your.mail`). |
+| `trustRoots[*].key` | - | :heavy_check_mark: if not using keyless | See [basics](../basics.md#validators). ECDSA/RSA public key from `cosign.pub` file or [KMS URI](https://docs.sigstore.dev/key_management/overview/#basic-usage). See additional notes [below](#kms-support). |
+| `trustRoots[*].keyless.Issuer` | - | :heavy_check_mark: if not using a key | Regex for the OIDC provider name which attests the identity. So, either the full URL of the OIDC provider can be used or only parts of it. |
+| `trustRoots[*].keyless.subject` | - | :heavy_check_mark: if not using a key | Regex of the identity that created the keyless signature. Usually an email address. So, either the full email address can be used or only parts of it (e.g. `your.mail@example.org` or `your.mail`). |
```
Starkteetje, line 97: This would be the docs for *Regex, while the docs for Issue and subject are missing

docs/validators/sigstore_cosign.md
```diff
@@ -92,13 +92,13 @@ kubectl run altsigned --image=docker.io/securesystemsengineering/testimage:co-si
...
-| `trustRoots[*].keyless.Issuer` | - | :heavy_check_mark: if not using a key | Name of the OIDC provider which attests the identity. This is a regex, so either the full url of the OIDC provider can be used or only parts of it. |
-| `trustRoots[*].keyless.subject` | - | :heavy_check_mark: if not using a key | Name of the identity that created the keyless signature. Usually an email address. This is a regex, so either the full email address can be used or only parts of it (e.g. `your.mail@example.org` or `your.mail`). |
+| `trustRoots[*].key` | - | :heavy_check_mark: if not using keyless | See [basics](../basics.md#validators). ECDSA/RSA public key from `cosign.pub` file or [KMS URI](https://docs.sigstore.dev/key_management/overview/#basic-usage). See additional notes [below](#kms-support). |
+| `trustRoots[*].keyless.Issuer` | - | :heavy_check_mark: if not using a key | Regex for the OIDC provider name which attests the identity. So, either the full URL of the OIDC provider can be used or only parts of it. |
```
Starkteetje, line 96:
```suggestion
| `trustRoots[*].keyless.issuer` | - | :heavy_check_mark: if not using a key | Regex for the OIDC provider name which attests the identity. So, either the full URL of the OIDC provider can be used or only parts of it. |
```

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -146,61 +186,82 @@ func (cv *CosignValidator) ValidateImage(
...
+
+	// do cosign validation on image per trust root in parallel
 	for _, verifier := range verifiers {
-		go func(verifier KeyVerifierTuple, verifierOutputChannel chan<- verifierOutput) {
-			// configure validation options
-			opts, err := cv.setupOptions(ctx, args, img, verifier.Verifier)
-			if err != nil {
-				verifierOutputChannel <- verifierOutput{nil, fmt.Errorf("error setting up cosign options: %s", err), ""}
-				return
+		go func(verifier Verifier, verifierOutputChannel chan<- verifierOutput) {
```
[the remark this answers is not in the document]
phbelitz: another time. if i do that now we need more unit tests. so ... another time.

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -78,6 +86,15 @@ type CosignPayloadImage struct {
 	Digest string `json:"Docker-manifest-digest"`
 }
 
+// verifiertOutput is a struct to hold the output
```
Starkteetje, line 89:
```suggestion
// verifierOutput is a struct to hold the output
```

internal/validator/cosignvalidator/cosign_validator.go
```diff
@@ -264,26 +291,75 @@ func (cv *CosignValidator) ValidateImage(
...
 }
 
+func (cv *CosignValidator) validateWithVerifier(
+	ctx context.Context,
+	imageRef name.Reference,
+	opts *cosign.CheckOpts,
+	verifier Verifier,
+	verifierOutputChannel chan<- verifierOutput,
+) {
+	copts := *opts
```
Starkteetje, line 310: Just from an API point, I think, it would be nicer to handover the value instead of pointer, such that it is clear that the original opts shouldn't (and factually) cannot be changed by the goroutine

docs/validators/sigstore_cosign.md
```diff
@@ -92,8 +92,15 @@ kubectl run altsigned --image=docker.io/securesystemsengineering/testimage:co-si
...
+| `trustRoots[*].keyless.subjectRegex` | - | :heavy_check_mark: if not using a key or subject | Regex of the identity that created the keyless signature. Usually an email address. So, either the full email address can be used or only parts of it (e.g. `your.mail@example.org` or `your.mail`). |
 | `host.rekor` | `rekor.sigstore.dev` | - | Rekor URL to use for validation against the transparency log (default sigstore instance is `rekor.sigstore.dev`). Setting `host` enforces successful transparency log check to pass verification. See additional notes [below](#transparency-log-verification). |
+| `host.rekorPubkey` | Public key of `rekor.sigstore.dev` | - | Public key used to verify signature of log entries from Rekor. |
+| `host.fulcioCert` | Root and intermediate certificates belonging to `fulcio.sigstore.dev` | - | The root certificate belonging the Fulcio CA which was used to create keyless signatures. |
+| `host.ctLogPubkey` | Public keys for the certificate transparency log provided by Sigstore | - | The public keys needed for verifying Signed Certificate Timestamps (SCT). |
```
Starkteetje, line 103: Either this should be `host.ctLogPubkeys` and we should accept multiple keys or the docs should state that only a single public key can be provided, e.g. no key*s*

docs/validators/sigstore_cosign.md
```diff
@@ -349,4 +357,37 @@ To disable this feature the `with.verifyInTransparencyLog` key can be set to `fa
...
+- name: keylessvalidator
+  type: cosign
+  trustRoots:
+  - name: keyless
+    keyless:
+      issuerRegex: "github"
+      subject: "philipp.belitz@securesystems.de
+```
+
+In case the signature was created using the Sigstore infrastructure, nothing else needs to be configured since Connaisseur will automatically retrieve all needed public keys and certificates from the Sigstore infrastructure. If the signature was created using a private infrastructure, the `host.fulcioCert` field can be used to specify the root certificate belonging to the Fulcio CA which was used to create the keyless signatures. The `host.fulcioCert` field should contain the root certificate in PEM format. The same applies to the `host.ctLogPubkey` field which can be used to specify the public keys needed for verifying Signed Certificate Timestamps (SCT) and the `host.rekorPubkey` field which can be used to specify the public key used to verify the signature of log entries from Rekor.
```
Starkteetje, line 374: Same as above for ctLogPubkey and `specify the public keys`

Starkteetje (approved):
🎉

Starkteetje (approved):

PR #1689, opened 2024-07-12
build: unpin ca certificates

phbelitz:
The version for the ca-certificates package, installed during an intermediate build step from an alpine image, is being unpinned. Reason for that is that alpine only ever provides a single version for their packages and once they update the ca-certificates packages, the Connaisseur build automatically fails, since the previous used version is no longer available.

## Checklist

### DIFF
--- a/build/Dockerfile
+++ b/build/Dockerfile
@@ -13,7 +13,8 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
     go build -o /go/bin ./cmd/connaisseur
 
 FROM alpine:3 as certs
-RUN apk --update --no-cache add ca-certificates=20240226-r0
+# hadolint ignore=DL3018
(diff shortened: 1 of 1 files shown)


build/Dockerfile
```diff
@@ -13,7 +13,7 @@ RUN --mount=type=cache,target=/root/.cache/go-build \
     go build -o /go/bin ./cmd/connaisseur
 
 FROM alpine:3 as certs
-RUN apk --update --no-cache add ca-certificates=20240226-r0
+RUN apk --update --no-cache add ca-certificates
```
Starkteetje, line 16: If you think the rule in general is reasonable, maybe
```suggestion
# hadolint ignore=DL3018
RUN apk --update --no-cache add ca-certificates
```
instead of ignoring it everywhere

Starkteetje (approved):

PR #1715, opened 2024-08-02
test: fix flakey redis-cert test

phbelitz:
The redis cert test tested that after upgrading a redis with a custom certificate the caching mechanism still worked. Problem was, the redis instance might have been ready on Kubernetes side, but internally not ready to receive connections. Thus the caching was tested and failed, before the newly upgraded redis was ready. This has been fixed with a wating mechanism.

## Checklist

### DIFF
--- a/test/integration/redis-cert/test.sh
+++ b/test/integration/redis-cert/test.sh
@@ -6,6 +6,9 @@ redis_cert_test() {
     update_with_file "redis-cert/install.yaml"
     install "make"
 
+    # wait for redis to be ready
+    wait_for_redis
+
     # create same image twice
     single_test "redis-cert-01-signed" "Testing signed nv1 image..." "deploy" "securesystemsengineering/testimage:signed" "default" "" "null"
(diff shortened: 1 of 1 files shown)


test/integration/redis-cert/test.sh
```diff
@@ -39,3 +45,12 @@ redis_cert_test() {
...
     rm output.log
 }
+
+wait_for_redis() {
+    timeout 30 bash -c '
+    while true; do
+        kubectl logs -n connaisseur -lapp.kubernetes.io/instance=redis --tail=-1 | grep -q  "Ready to accept connections tls" && break
+        sleep 1
+    done
+    '
```
Starkteetje, line 55: You could fail open here and hope it still works
```suggestion
wait_for_redis() {
    timeout 30 bash -c '
    while true; do
        kubectl logs -n connaisseur -lapp.kubernetes.io/instance=redis --tail=-1 | grep -q  "Ready to accept connections tls" && break
        sleep 1
    done
    ' || true
```
That would allow not breaking the test if the message changes, but also prolong the test by 30s everytime if it changed. Your call (I guess I wouldnt do it, but you're more vocal on test failures; I just had the thought that one could do it)

Starkteetje (approved):

Issue #1734, opened 2024-08-22
Passing in kubernetes.deployment.envs to connaisseur chart values for chart version 2.4 upwards breaks helm deployment 


edison-vflow:
**Describe the bug**
<A clear and concise description of what the bug is.>

From Connaisseur Helm chart version `2.4` onwards it seems when you have custom environment variables `kubernetes.deployment.envs`
eg https://sse-secure-systems.github.io/connaisseur/latest/validators/sigstore_cosign/#kms-support where you need to support KMS and inject AWS environment variables 

```
kubernetes:
  deployment:
    resources:
      limits:
        cpu: 1000m
        memory: 512Mi
      requests:
        cpu: 100m
        memory: 512Mi
    annotations:
      cluster-autoscaler.kubernetes.io/safe-to-evict: "true"
    envs:
      AWS_DEFAULT_REGION: my-region
      AWS_ACCESS_KEY_ID: my-access-key
      AWS_SECRET_ACCESS_KEY: my-secret-key

  service: {}
```

Once you have an `envs` fragment, the helm deployment fails 

`│ Error: unable to build kubernetes objects from release manifest: error validating "": error validating data: [ValidationError(Deployment.spec.template.spec.containers[0].env[1]): unknown field "secretRef" in io.k8s.api.core.v1.EnvVar, ValidationError(Deployment.spec.template.spec.containers[0].env[1]): missing required field "name" in io.k8s.api.core.v1.EnvVar]`

**Expected behavior**
<A clear and concise description of what you expected to happen.>

In version 2.3.4 and downwards, having this `envs` fragment works without failing 

Comparing the default values of chart 2.3.4 and downwards and 2.4 upwards, there was no change to the yaml structure implying the same `kubernetes.deployment.envs` can be passed.

**Optional: To reproduce**
<Stepwise guide or instructions or minimal example to reproduce.>

Deploy the helm chart  using terraform 

```
resource "helm_release" "connaisseur" {
  name              = "connaisseur"
  namespace         = kubernetes_namespace.connaisseur.metadata[0].name
  create_namespace  = true
  atomic            = false
  repository        = "https://sse-secure-systems.github.io/connaisseur/charts"
  chart             = "connaisseur"
  version           = 2.6.0 

  values = [
    file("${abspath(path.module)}/values.yaml")
  ]

}
``` 

* use the default values file `https://artifacthub.io/packages/helm/connaisseur/connaisseur/2.6.0?modal=values` 
[...]

PR #1735, opened 2024-08-23
fix: correctly place connaisseur-env-secret in deployment yaml

edison-vflow:
From chart version 2.4, a refactor was made that places `connaisseur-env-secret` under `env` instead of under `envFrom` in the deployment.This breaks cosign validations that require custom secrets like ECR
Relevant helm template commands were run to ensure chart compiles correctly with `kubernetes.deployment.envs` values populated and not populated 

Related issue: https://github.com/sse-secure-systems/connaisseur/issues/1734


phbelitz:
closing in favor of #1736

PR #1801, opened 2024-10-25
Ci/rework

phbelitz:
rework for ci

Restructures the ci, renames some of the arguments and their values and adds a documentation.

## Checklist


.github/workflows/101_build.yml
```diff
@@ -1,14 +1,12 @@
...
 
-permissions: {} 
-
 on:
   workflow_call:
     inputs:
-      skip:
-        description: "Want to skip running certain jobs 'none', 'non-required', 'all'?"
+      jobs_to_run:
+        description: "Which build jobs should be run: 'all', 'only-required', 'skip'?"
```
Starkteetje, line 7: IMO `skip` doesn't make sense the way you renamed. Which jobs should be run? Skip! What?
I'd keep `none`.

Other than that, I'm a fan of configuring the positive instead of negative

Starkteetje: Obviously the same goes for all the other files

.github/workflows/101_build.yml
```diff
@@ -1,14 +1,12 @@
 name: build
 
-permissions: {}
```
Starkteetje, line 3: Why delete this? It removes default permissions for all jobs that don't explicitly require them, which is a good thing. IMO this should probably prefix every workflow file

.github/workflows/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Pipeline Definitions
+
+The main pipeline definitions are as follows:
+
+- `01_pr.yaml`: Triggered by creating a pull request against master and develop. It runs the reusable CI pipeline.
+- `02_push.yaml`: Triggered by pushing code to master or develop and is otherwise identical to the `00_pr.yaml` pipeline.
+- `03_release.yaml:`: Triggered by pushing a tag prefixed with `v` and builds the Connaisseur image, checks that the image tag matches the git tag, runs the integration tests, publishes the helm chart to ArtifactHub and updates the Connaisseur documentation.
+- `04_publish.yaml`: Triggered manually and publishes the helm chart to ArtifactHub and updates the Connaisseur documentation. This is meant to be used in case the `03_release.yaml` pipeline fails and the helm chart needs to be published manually.
+- `05_nightly.yaml`: Triggered by a scheduled job that runs every night at 01:30, which runs the reusable CI pipeline but without unit tests, sast jobs, documentation publishing and integration tests. This pipeline also cleans up old images from the GitHub container registry.
```
Starkteetje, line 21:
```suggestion
- `05_nightly.yaml`: Triggered by a scheduled job that runs every night, which runs the reusable CI pipeline but without unit tests, sast jobs, documentation publishing and integration tests. This pipeline also cleans up old images from the GitHub container registry.
```
TMI, I'd say

.github/workflows/README.md
```diff
@@ -0,0 +1,49 @@
...
+- `06_dockerhub_check.yaml`: Triggered by a scheduled job that runs every wednesday at 06:37, which pulls all test images from DockerHub to make sure they are still available and don't get deleted because of inactivity.
```
Starkteetje, line 22:
```suggestion
- `06_dockerhub_check.yaml`: Triggered by a scheduled job that runs weekly, which pulls all test images from DockerHub to make sure they are still available and don't get deleted because of inactivity.
```

.github/workflows/README.md
```diff
@@ -0,0 +1,49 @@
...
+
+## Reusable Pipeline Definitions
+
+The reusable pipeline definitions are as follows:
+
+- `100_ci.yaml`: This is the combined pipeline definition that includes all other jobs in this section and can be configured to skip certain jobs.
+- `101_build.yaml`: Builds the Connaisseur image and pushes it to the GitHub container registry. Also prints a summary to the GitHub Actions summary.
+- `102_compliance.yaml`: Does a dependency review, looks for invalid licenses and check that the commit message follows the conventional commit format.
+- `103_unit-tests.yaml`: Runs the unit tests.
+- `104_sast.yaml`: Runs static application security testing, including codeql, golangci-lint, gosec, hadolint, kubelinter and trivy config.
```
Starkteetje, line 32:
```suggestion
- `104_sast.yaml`: Runs static application security testing.
```

.github/workflows/README.md
```diff
@@ -0,0 +1,49 @@
...
+- `105_sca.yaml`: Runs software composition analysis using trivy image.
```
Starkteetje, line 33:
```suggestion
- `105_sca.yaml`: Runs software composition analysis.
```

.github/workflows/05_nightly.yaml
```diff
@@ -0,0 +1,31 @@
...
+      security-events: write
+    secrets: inherit
+    with:
+      unit_tests: skip
+      sast: skip
+      docs: skip
+      integration_tests: skip
+      output_type: table
+
+  cleanup-registry:
```
Starkteetje, line 28: I would assume this job needs more permissions than none, but maybe I'm mistaken

phbelitz: It's the same as before and that works 🤷 I guess since a separate token is passed on, which has enough permissions there is no need to define additional `permissions`.

.github/workflows/101_build.yml
phbelitz: changed to `skip-all`

Starkteetje: I still think that's worse than none, because it's not quite an answer to the question you pose in the description, but won't argue this further

Starkteetje (approved):

PR #1835, opened 2024-11-22
ci: rework build step

phbelitz:
Removed the signing of the build image and creation of SBOM to reduce complexity. Also moved the code of the context and build action into the reusable build workflow.

## Checklist


.github/workflows/.reusable-build.yml
```diff
@@ -79,21 +141,46 @@ jobs:
...
-          repo_token: ${{ secrets.GITHUB_TOKEN }}
-          cosign_private_key: ${{ secrets.COSIGN_PRIVATE_KEY }}
-          cosign_password: ${{ secrets.COSIGN_PASSWORD }}
+          push: true
+          cache-from: type=gha
+          cache-to: type=gha,mode=max
+          file: build/Dockerfile
+          labels: ${{ needs.context.outputs.build_labels }}
+          tags: ${{ steps.tags.outputs.tags }}
+          sbom: false  # Duplicates SBOMs manually created below
```
Starkteetje, line 173: this comment is no longer accurate

test/integration/common_functions.sh
```diff
@@ -180,20 +180,12 @@ ghcr_update() { # $1: file
     yq e -i '.kubernetes.deployment.image.tag = env(TAG)' $1
     # add imagePullSecrets
     yq e -i '.kubernetes.deployment.image.imagePullSecrets[0].name = env(IMAGEPULLSECRET)' $1
-    # add cosign validator:
```
Starkteetje, line 183: This change completely changes the context of this function. It should at least be renamed. Probably, we can get rid of it, too

Starkteetje: Maybe even the whole stuff related to GHCR can be scrubbed from the testcases

phbelitz: the ghcr image is still present, this much hasn't changed. only thing that has changed is that the image is no longer signed. thus the function is still valid.

Starkteetje (approved):

PR #1842, opened 2024-11-25
feat: support single-line json as log format

czenker:
Also introduces a structure to add more log formats should the need arise. Does not change the default of logging prettified json.

## Description

Log Shippers are usually less complex and less error-prone when they only have to handle single-line logs. This PR adds the ability to change the log format to single-line json, but does not change the default multi-line json.

## Checklist


phbelitz:
closing in favor of #1846

PR #1846, opened 2024-11-29
feat: support single-line json as log format

phbelitz:
Also introduces a structure to add more log formats should the need arise. Does not change the default of logging prettified json.

> Duplicates #1842, since pipeline can only be run by repo members (sadly).

## Checklist

### DIFF
--- a/internal/utils/logging.go
+++ b/internal/utils/logging.go
@@ -31,6 +31,22 @@ func ConnaisseurLogLevel() logrus.Level {
 	return LogLevel(os.Getenv(constants.LogLevel))
 }
 
+func logFormat(logFormat string) constants.ConnaisseurLogFormat {
+	switch strings.ToLower(logFormat) {
+	case "json":
+		return constants.LogFormatJson
+	case "json-pretty":
+		return constants.LogFormatJsonPretty
+	default:
+		// default for backwards compatibility
+		return constants.LogFormatJsonPretty
+	}
+}
+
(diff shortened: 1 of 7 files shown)


internal/utils/logging_test.go
```diff
@@ -83,6 +83,27 @@ func TestConnaisseurLogLevel(t *testing.T) {
...
+		want      constants.ConnaisseurLogFormat
+	}{
+		{
+			logFormat: "json",
+			want:      constants.LogFormatJson,
+		},
+		{
+			logFormat: "json-pretty",
+			want:      constants.LogFormatJsonPretty,
+		},
```
Starkteetje, line 98:
```suggestion
		{
			logFormat: "json-pretty",
			want:      constants.LogFormatJsonPretty,
		},
		{
			logFormat: "not valid",
			want:      constants.LogFormatJsonPretty,
		},
```
maybe also default test?

internal/utils/logging.go
```diff
@@ -31,6 +31,22 @@ func ConnaisseurLogLevel() logrus.Level {
 	return LogLevel(os.Getenv(constants.LogLevel))
 }
 
+func LogFormatter(logFormat string) constants.ConnaisseurLogFormat {
```
Starkteetje, line 34: Not really used or useful anywhere else, right? So I guess should be private func
also nit: Probably rather `logFormat` instead of `LogFormatter` because result isn't really a thing that does anything by itself

Starkteetje (approved):

PR #1851, opened 2024-12-06
feat: pod disruption budget

### LINKED ISSUE #1850: Pod disruption budget
**Describe the feature**
When we are upgrading the cluster or adding/removing node pools, we are getting error that new image cannot be pulled because the connaisseur service is not available. The solution is either to uninstall the helm chart before the operation and install it back after, or use pod disruption budget like following:

```yaml
apiVersion: policy/v1
kind: PodDisruptionBudget
metadata:
  name: connaisseur
  namespace: connaisseur
spec:
  minAvailable: 1
  selector:
    matchLabels:
      app.kubernetes.io/instance: connaisseur
```

<A clear and concise description of what the problem is.>

<Maybe you already have an implementation and alternatives in mind. Please share your thoughts.>

<Add any other context or screenshots about the feature request here.>
I would follow this link, but I'm lazy atm. https://awjunaid.com/kubernetes/how-to-implement-pod-disruption-budgets-with-helm-in-kubernetes/

phbelitz:
Added a flag to enable/disable a [pod disruption budget](https://kubernetes.io/docs/tasks/run-application/configure-pdb/), helping Connaisseur to be always available.

fixes #1850

## Checklist


charts/connaisseur/values.yaml
```diff
@@ -102,6 +102,9 @@ kubernetes:
   # (except for pods – see kubernetes.deployment.podLabels for that)
   additionalLabels: {}
 
+  # activate pod disruption budget
+  podDisruptionBudget: false
```
Starkteetje, line 106: How about exposing the minAvailable instead? Then the feature is more powerful and if someone wants to fiddle with it, we don't have to add to the API in a weird way?

Starkteetje (approved):

PR #1878, opened 2025-01-17
docs: updated badges

phbelitz:
Fixed and updated the workflow status badges in the README. Also renamed the workflow name for pushing events from 'push' to 'cicd' since this is represented in the status badge and would sound confusing.

## Checklist


.github/workflows/02_push.yml
```diff
@@ -1,4 +1,4 @@
-name: push
+name: cicd
```
Starkteetje, line 1: Why this rename? It runs on push and that's its distinction from the 01_pr for example. So I don't see `cicd` as a better name

And if it were a better name, the file name should also reflect this

phbelitz: > Fixed and updated the workflow status badges in the README. Also renamed the workflow name for pushing events from 'push' to 'cicd' since this is represented in the status badge and would sound confusing.

Starkteetje: But all other files have same name and name attribute value, so this sticks out, would then at least add a comment here to explain this discrepancy

README.md
```diff
@@ -1,7 +1,7 @@
...
 [![Artifact Hub](https://img.shields.io/endpoint?url=https://artifacthub.io/badge/repository/connaisseur)](https://artifacthub.io/packages/search?repo=connaisseur)
-![pipeline status](https://github.com/sse-secure-systems/connaisseur/workflows/cicd/badge.svg)
-[![nightly-build](https://github.com/sse-secure-systems/connaisseur/actions/workflows/nightly-build.yaml/badge.svg)](https://github.com/sse-secure-systems/connaisseur/actions/workflows/nightly-build.yaml)
+[![cicd](https://github.com/sse-secure-systems/connaisseur/actions/workflows/02_push.yml/badge.svg)](https://github.com/sse-secure-systems/connaisseur/actions/workflows/02_push.yml)
+[![nightly](https://github.com/sse-secure-systems/connaisseur/actions/workflows/05_nightly.yaml/badge.svg)](https://github.com/sse-secure-systems/connaisseur/actions/workflows/05_nightly.yaml)
 [![codecov](https://codecov.io/gh/sse-secure-systems/connaisseur/branch/master/graph/badge.svg)](https://codecov.io/gh/sse-secure-systems/connaisseur)
 [![OpenSSF Scorecard](https://api.securityscorecards.dev/projects/github.com/sse-secure-systems/connaisseur/badge)](https://api.securityscorecards.dev/projects/github.com/sse-secure-systems/connaisseur)
```
Starkteetje, line 6: I think we threw out scorecard, right? If so, we should remove this too

Starkteetje (approved):

PR #1885, opened 2025-01-24
wip: notation implementation

phbelitz:
Implementation for notation

## Checklist


internal/validator/notation/trust_store.go
```diff
@@ -0,0 +1,49 @@
...
+	"encoding/pem"
+	"fmt"
+
+	"github.com/notaryproject/notation-go/verifier/truststore"
+)
+
+type InMemoryTrustStore struct {
+	trustRoots []auth.TrustRoot
+	certs      map[string][]*x509.Certificate
+	truststore.X509TrustStore
```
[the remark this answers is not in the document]
phbelitz: simply to signify that this is implementing the interface.

internal/validator/notation/notation_validator.go
```diff
@@ -0,0 +1,173 @@
+package notation
```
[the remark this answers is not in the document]
phbelitz: yea. so notation can only do verifications on digests. if we get an image with a tag, the tag gets resolved into the latest digest first, which then is used to find a signature. the signature then is verified. if successfull, we take the resolved digest and do our mutation.

docs/validators/notation.md
```diff
@@ -0,0 +1,253 @@
...
+
+`.application.validators[*]` in `charts/connaisseur/values.yaml` supports the following keys for Notation (refer to [basics](../basics.md#validators) for more information on default keys):
+
+| Key | Default | Required | Description |
+| - | - | - | - |
+| `name` | - | :heavy_check_mark: | See [basics](../basics.md#validators). |
+| `type` | - | :heavy_check_mark: | `notation`; the validator type must be set to `notation`. |
+| `trustRoots[*].name` | - | :heavy_check_mark: | See [basics](../basics.md#validators). |
+| `trustRoots[*].cert` | - | :heavy_check_mark: | X.509 certificate in PEM format used for signature verification. This should be the root signing certificate or a trusted CA certificate. |
+| `trustRoots[*].tsCert` | - | - | X.509 certificate in PEM format used for timestamp counter signature verification. |
```
Starkteetje: Are these "regular" certs by default, like those in OS cert pools? Asking, because if you usually have to configure this (as ts verification defaults to always), I'd state that. If on the other hand, you don't I would also highlight that this is more for a case where this is a dedicated PKI by choice

docs/validators/notation.md
```diff
@@ -0,0 +1,253 @@
...
+| `auth.` | - | - | Authentication credentials for registries with restricted access (e.g. private registries or rate limiting). See additional notes [below](#authentication). |
```
Starkteetje:
```suggestion
| `auth.` | - | - | Authentication credentials for registries with restricted access (e.g., private registries or rate limiting). See additional notes [below](#authentication). |
```

docs/validators/notation.md
```diff
@@ -0,0 +1,253 @@
...
+| `auth.secretName` | - | - | Name of a Kubernetes secret in Connaisseur namespace that contains [dockerconfigjson](https://kubernetes.io/docs/concepts/configuration/secret/#docker-config-secrets) for registry authentication. See additional notes [below](#dockerconfigjson). |
+| `cert` | - | - | A TLS certificate in PEM format for private registries with self-signed certificates. |
+
+`.application.policy[*]` in `charts/connaisseur/values.yaml` supports the following additional keys for Notation (refer to [basics](../basics.md#image-policy) for more information on default keys):
+
+| Key | Default | Required | Description |
+| - | - | - | - |
+| `with.trustRoot` | - | :heavy_check_mark: | Name of the trust root to use for verification. |
```
Starkteetje: I think the required checkmark is wrong here and for other validators, as if the trustroot name is default, this is picked, or am I mistaken?

test/integration/notation/cases.yaml
```diff
@@ -0,0 +1,11 @@
+# simple test cases for notation
```
Starkteetje: we should add additional test cases for required timestamp failing due to no timestamp verification

maybe the other verification levels may be interesting, too, but the ts thing is most relevant as it ensures that the highest security setting works

internal/validator/notation/notation_validator_test.go
```diff
@@ -0,0 +1,253 @@
...
+		{ // 6: tesing tsa
+			"08_tsa_cert",
+			"ghcr.io/test/test:latest",
+			policy.RuleOptions{},
+			"ghcr.io/test/test",
+			[]string{"ca:default", "tsa:default"},
+			trustpolicy.LevelStrict.Name,
+			trustpolicy.OptionAlways,
+			"",
+		},
```
Starkteetje: test case for faulty timestamp cert

internal/validator/notation/trust_store.go
```diff
@@ -0,0 +1,91 @@
...
+		}
+	}
+
+	return &InMemoryTrustStore{
+		trustRoots: trustRoots,
+		certs:      certs,
+	}, nil
+}
+
+func parseCertificate(cert string) ([]*x509.Certificate, error) {
```
Starkteetje: nit: name and return type don't really match. maybe just return the cert and build the list in the outer function?

docs/validators/notation.md
phbelitz: so are you asking whether those certs need to be installed into the OS, or wether they can just be added here and be use for some kind of verification?

Starkteetje (approved):

PR #2073, opened 2026-03-06
feat: Allow updates to resources marked for deletion

### LINKED ISSUE #2053: Connaisseur blocks Kubernetes garbage collection causing pods to remain stuck in Terminating state
## Describe the bug

  Kubernetes garbage collection fails for pods with ownerReferences when the parent resource is deleted. Connaisseur blocks the pod UPDATE operation (finalizer removal) by attempting to validate container images against a parent resource that no longer exists, resulting in 404 errors and webhook denial. This causes pods to remain stuck in Terminating state indefinitely.

  ### Root Cause

  When a Job/Deployment is deleted, Kubernetes garbage collector attempts to clean up owned pods by removing finalizers (UPDATE operation). Connaisseur intercepts this UPDATE request and tries to fetch the parent Job to validate container images. Since the parent is already deleted, Connaisseur receives a 404 error and denies the request, preventing garbage collection from completing.

  ### Impact

  - Pods remain stuck in Terminating state indefinitely (observed 138-169 days in production)
  - Manual cleanup is also blocked by the same validation logic
  - Cluster resources are not properly reclaimed
  - Garbage collection process is fundamentally broken for validated namespaces

  ## Expected behavior

  Connaisseur should skip image validation for pods that are being deleted (have `metadata.deletionTimestamp` set). These pods are in the cleanup/termination phase and only require metadata changes (finalizer removal), not container image validation.

  ### Proposed Solution

  Add a check in the admission handler to skip validation when:
  1. Request is an UPDATE operation
  2. Resource is a Pod
  3. Pod has `metadata.deletionTimestamp` set (indicates deletion in progress)

  **Example implementation location:** `internal/handler/mutate.go:163` (mutateReview function)

  ```go
  // Skip validation for terminating pods (garbage collection)
  if ar.Request.Operation == admission.Update &&
     ar.Request.Kind.Kind == "Pod" {
      pod := &core.Pod{}
      if err := json.Unmarshal(ar.Request.Object.Raw, pod); err == nil {
          if pod.ObjectMeta.DeletionTimestamp != nil {
              return &admission.AdmissionResponse{
                  UID:     ar.Request.UID,
                  Allowed: true,
              }, &alerting.NotificationValues{
                  Result: constants.NotificationResultSkip,
              }
          }
      }
  }

  To Reproduce

  1. Deploy Connaisseur with image signature validation enabled
  2. Create a namespace with Connaisseur validation label:
  kubectl create namespace test-gc
  kubectl label namespace test-gc securesystemsengineering.connaisseur/webhook=validate
  3. Create a Job with a signed/allowed image:
  kubectl create job test-job --image=<signed-image> -n test-gc -- sleep 120
  4. Wait for pod to start (verify it's running):
  kubectl get pods -n test-gc
  5. Delete the Job while pod is running:
  kubectl delete job test-job -n test-gc
  6. Observe pod stuck in Terminating state:
  kubectl get pods -n test-gc
  # Pod remains in Terminating indefinitely
  7. Check Connaisseur logs - will show 404 errors attempting to fetch deleted Job:
  HTTPError: 404 Client Error: Not Found for url:
  https://kubernetes.default.svc/apis/batch/v1/namespaces/test-gc/jobs/test-job
  8. Attempt manual cleanup (also fails):
  kubectl patch pod <pod-name> -n test-gc -p '{"metadata":{"finalizers":null}}' --type=merge
  # Error from server: admission webhook "connaisseur-svc.connaisseur.svc" denied the request: unknown error
  9. Even force delete fails (hangs indefinitely):
  kubectl delete pod <pod-name> -n test-gc --force --grace-period=0
  # Command hangs, pod remains in Terminating state

  Environment

  - Kubernetes Cluster: Amazon EKS 1.30
  - Connaisseur: v3.0.0 (Helm chart v2.0.0)
  - Container Registry: GitHub Container Registry (ghcr.io)
  - Image Signing: Cosign

  Note: This issue affects all Connaisseur versions including v3.9.1 (latest) - confirmed no deletionTimestamp handling exists in current codebase.

  Additional Context

  Security Consideration

  This change maintains security because:
  - Validation is skipped only for pods with metadata.deletionTimestamp set (actively being deleted)
  - Running pods without deletionTimestamp still have full UPDATE validation - Connaisseur continues to block unauthorized image changes via kubectl edit pod
  - Terminating pods are in cleanup/removal phase and will be deleted from the cluster
  - Image validation at pod CREATE time prevents unsigned images from initially entering the cluster
  - Parent resource updates (Deployment/Job) remain fully validated on both CREATE and UPDATE

  Note: While Kubernetes allows editing pod specs via kubectl edit pod, Connaisseur's UPDATE validation blocks unauthorized image changes on running pods. This fix only bypasses validation for pods that are already marked for deletion.

  Kubernetes Reference

  This pattern is consistent with Kubernetes garbage collection behavior documented at:
  https://kubernetes.io/docs/concepts/architecture/garbage-collection/

  Production Impact

  Multiple pods observed stuck in Terminating state, unable to be cleaned up due to this issue. Manual intervention attempts (patch, force delete) all fail.

  Alternative Approaches Investigated

  All alternatives have significant drawbacks:

  1. Removing UPDATE validation from pod webhook rules - Tested and confirmed this creates a critical security hole: allows changing running pod images to unsigned/malicious images via kubectl edit pod, completely bypassing Connaisseur's security controls.
  2. Using failurePolicy: Ignore temporarily - Disables all webhook validation during cleanup window, creates temporary security gap for all operations.
  3. Force deleting pods with --grace-period=0 --force - Command hangs indefinitely, does not resolve the issue. Finalizers remain and block deletion.
  4. objectSelector with field selectors - Investigated but Kubernetes objectSelector only supports label selectors, not field selectors like deletionTimestamp. This is a Kubernetes API limitation, not something Connaisseur can work around via configuration.

  Logs Example

  Actual error from Connaisseur when attempting to validate terminating pod UPDATE:

  {
    "timestamp": "2026-02-03 22:43:53.449343",
    "message": "Traceback (most recent call last):\n
    File \"/app/connaisseur/workload_object.py\", line 76, in parent_containers\n
    parent = k_api.request_kube_api(...)\n
    requests.exceptions.HTTPError:
    404 Client Error: Not Found for url:
    https://172.20.0.1:443/apis/batch/v1/namespaces/test-gc/jobs/test-job\n"
  }

Starkteetje:
Allow updates to resources marked for deletion

Fixes #2053

## Description
Previously, we validated admission requests on creation and update. For deletion, we do not block requests as resources should be able to deleted, irrespective of their signing status. However, if resources contained finalizers, these need to be removed through updates before deletion can occur. This is a safe operation from Connaisseur's point of view as an update on a resource marked for deletion cannot meaningfully alter the object, especially not change container images. This PR changes behaviour to allow these patches even if the container images of the resource might be invalid

## Checklist


Starkteetje:
@phbelitz I feel like we should document this, but there's no great place that comes to mind, where it would fit. It's not a "feature" feature, so wouldn't want to create a full page, but also wouldn't see it as important enough to distract from main content in the basic READMEs. Any thoughts?

internal/kubernetes/workload_object.go
```diff
@@ -73,6 +76,7 @@ func NewWorkloadObjectFromBytes(
 		initContainers = pod.Spec.InitContainers
 		ephemeralContainers = pod.Spec.EphemeralContainers
 		owners = pod.OwnerReferences
+		deleted = pod.DeletionTimestamp != nil
```
phbelitz, line 79: can the deleteion timestamp be in the future ...? 🤔

phbelitz (commented):
for the documentation: maybe we can add a "Troubleshooting" section to usage, where we can explained known quirks that happen when using Connaisseur. there we could mention things like this.

internal/kubernetes/workload_object.go
Starkteetje: Yes: https://kubernetes-client.github.io/javascript/classes/V1ObjectMeta.html#deletionTimestamp

phbelitz (dismissed):

docs/advanced.md
```diff
@@ -0,0 +1,27 @@
...
+
+Connaisseur is a Mutating Admission Controller and as such has a `MutatingWebhookConfiguration` which specifies where requests are to be sent for mutation and admission review.
+By default, Connaisseur is configured to be a fail-closed service: If anything goes wrong, including service unavailability, resource admission will be blocked.
+This works well if Connaisseur is installed into a running cluster.
+Then the actual Connaisseur resources are deployed and then the `MutatingWebhookConfiguration` is installed, routing to the Connaisseur pods and subsequent verification can start.
+If however, the cluster is restarted with all of Connaisseur's resources in place, the `MutatingWebhookConfiguration` will require all starting resources to be verified by Connaisseur, which itself hasn't been deployed yet.
+A not-yet-running Connaisseur would need to allow itself to be started, a classical chicken-and-egg problem.
+
+There is unfortunately no good way around it, so you'll have to uninstall and reinstall Connaisseur when restarting the cluster.
+If you want all other resources are properly validated, make sure that Connaisseur is deployed before restoring the remaining resources.
```
phbelitz:
```suggestion
If you want all other resources properly validated, make sure that Connaisseur is deployed before restoring the remaining resources.
```

phbelitz (changes requested):

phbelitz (approved):

PR #2125, opened 2026-06-16
fix: allow overriding and disabling serviceaccount creation

### LINKED ISSUE #2124: ServiceAccount name is hardcoded in Helm chart, preventing overrides
Hi team. I'm trying to install the lastest version of the Connaisseur Helm Chart (v2.10.0 at the moment of writing).

In the Connaisseur Helm chart, the ServiceAccount name is hardcoded in `templates/_helpers.tpl` via the `connaisseur.serviceAccountName` helper:

```yaml
{{- define "connaisseur.serviceAccountName" -}}
{{- include "connaisseur.name" . }}-serviceaccount
{{- end }}
```

This ignores any standard overrides and makes it difficult to integrate Connaisseur with automated infrastructure provisioning (like Terraform EKS Pod Identity modules), which often require specific, deterministic ServiceAccount names (like `connaisseur-sa`).

While `values.yaml` has a `kubernetes.serviceaccount` section for annotations, there is no way to override the `name`.

**Proposed Solution**
Update the `connaisseur.serviceAccountName` helper to check for an explicit name in the values (e.g., `.Values.kubernetes.serviceaccount.name`) and only fall back to the `-serviceaccount` suffix if it's not set.

nicomazzarello:
The Connaisseur Helm chart hardcodes the ServiceAccount name in the helper template and always creates it, which prevents users from overriding it with custom names or using an externally managed ServiceAccount.

This PR adds two new configuration options to the `kubernetes.serviceaccount` block in `values.yaml`:
1.  `name`: Allows overriding the default ServiceAccount name.
2.  `create`: Allows toggling the creation of the ServiceAccount resource.

This change is particularly useful for users integrating Connaisseur with external infrastructure management tools (like Terraform EKS Pod Identity) where the ServiceAccount may be pre-provisioned or require a specific, deterministic name.

Fixes #2124

### DIFF
--- a/charts/connaisseur/templates/_helpers.tpl
+++ b/charts/connaisseur/templates/_helpers.tpl
@@ -57,7 +57,11 @@ Create the name of the service account to use
 Create the name of the service account to use
 */}}
 {{- define "connaisseur.serviceAccountName" -}}
+{{- if .Values.kubernetes.serviceaccount.name }}
+{{- .Values.kubernetes.serviceaccount.name }}
+{{- else }}
 {{- include "connaisseur.name" . }}-serviceaccount
+{{- end }}
 {{- end }}
 
 {{/*
--- a/charts/connaisseur/templates/serviceaccount.yaml
+++ b/charts/connaisseur/templates/serviceaccount.yaml
@@ -1,3 +1,4 @@
+{{- if .Values.kubernetes.serviceaccount.create -}}
 apiVersion: v1
 kind: ServiceAccount
 metadata:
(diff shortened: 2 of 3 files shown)


lmbaschiera:
This would be useful in my use-case as well, since I manage the creation and binding of the pod-identity using a different naming convention (`<workload>-sa`), and not being able to customize the serviceaccount name (which most charts support) makes me have to program for edge cases.

Starkteetje:
Hi @nicomazzarello 
thank you for your contribution. I cherry-picked your commit and did small changes in #2127 as our CI pipelines (currently) dont work with forks.

PR #2128, opened 2026-06-19
feat: add servicemonitor resource template to helm chart

lmbaschiera:
## Description

Adds an optional Prometheus Operator `ServiceMonitor` to the Connaisseur Helm chart so clusters running the operator can scrape Connaisseur's existing `/metrics` endpoint.

- New template `charts/connaisseur/templates/servicemonitor.yaml`, gated by `kubernetes.serviceMonitor.enabled` (disabled by default — no impact on existing installs).
- The endpoint is scraped over **HTTPS** (`scheme: https`) because Connaisseur serves `/metrics` on the same TLS listener as the webhook, using the webhook serving certificate. Scrape behaviour is configurable via new values: `enabled`, `interval`, `scrapeTimeout`, `additionalLabels` (to match the Prometheus instance's `serviceMonitorSelector`), and `tlsConfig` (defaults to `insecureSkipVerify: true`; users can supply a CA + `serverName` to verify).
- The ServiceMonitor reuses the existing `connaisseur.selectorLabels` so it targets the existing Service, and pins `namespaceSelector` to the release namespace.
- Helm unittests written (and passing) for the new resource. To run these, the [unittest plugin](https://github.com/helm-unittest/helm-unittest) needs to be installed.

When enabled without the `monitoring.coreos.com/v1` CRD present, `helm install` fails loudly. This is intentional, as enabling the resource is an explicit opt-in.

## Checklist


lmbaschiera:
@Starkteetje sorry to ping you directly - wondering if there's any way I can get help getting this reviewed? Currently we've been declaring the ServiceMonitor as an extra manifest in a different Helm Release to work around the lack of this feature in Connaisseur, which is not great. 

Thanks,

charts/connaisseur/values.yaml
```diff
@@ -57,6 +57,22 @@ kubernetes:
...
+    interval: 30s
+    scrapeTimeout: 10s
+    # additional labels to attach to the ServiceMonitor; typically used to match
+    # the Prometheus instance's serviceMonitorSelector, e.g.
+    #   release: kube-prometheus-stack
+    additionalLabels: {}
+    # connaisseur serves /metrics over HTTPS using the webhook serving cert.
+    # By default the certificate is not verified; supply ca/serverName here to verify.
+    tlsConfig:
+      insecureSkipVerify: true
```
Starkteetje: I don't quite like this. I totally see that this might be desired, but we try to really do security by default, so I wouldnt want to include this in the default (even if behind an `enabled: false` default). I'm not too familiar with the Prometheus Operator, but a brief look gives me the impression, we could reference Connaisseur's tls.crt in the `connaisseur-tls` secret here and get an authenticated TLS connection

Starkteetje (commented):
Hi @lmbaschiera 

sorry that your PR got lost and thanks for your contribution. I added one comment pertaining to insecureSkipVerify, apart from that the PR lgtm

charts/connaisseur/values.yaml
lmbaschiera: @Starkteetje - Nice catch, thanks. I dropped `insecureSkipVerify` entirely; the ServiceMonitor now verifies the connection by default against Connaisseur's own certificate.

A couple of notes on why the current state works:

- `serverName` is needed because Prometheus scrapes the endpoint's pod IP, which isn't among the certificate's SANs.
- The Prometheus Operator resolves the secret reference in the ServiceMonitor's own namespace, so no certificate has to be copied into the Prometheus namespace.

`kubernetes.serviceMonitor.tlsConfig` now defaults to `{}` and only needs setting if you bring your own certificate through `kubernetes.deployment.tls` that is issued by a separate CA. In that case, the supplied `tlsConfig` replaces the defaults, so it needs to carry `serverName` too. 

The unit tests were also updated accordingly.

Starkteetje: Thank you. I took your changes and opened a separate PR (https://github.com/sse-secure-systems/connaisseur/pull/2175) since our pipeline currently doesn't run for outside contributors 😞 . You're still the commit author, I just squashed the changes into one commit and signed it
I did drop the tests over there. While I do very much like the idea, I think this should be a concerted effort to add helm tests and then run them in the CI, if we want to add this, not a one-off test file

Starkteetje:
Closed in favour of https://github.com/sse-secure-systems/connaisseur/pull/2175

lmbaschiera:
Great stuff, thanks @Starkteetje!

PR #2131, opened 2026-06-27
fix: match registry host case-insensitively in policy rules

### LINKED ISSUE #2130: Policy rules match the registry host case-sensitively
**Describe the bug**

`Config.MatchingRule` (`internal/config/config.go`) matches the image against policy rule patterns case-sensitively on the registry host. Registry hosts are case-insensitive, so `MyRegistry.io/app` and `myregistry.io/app` pull the same image but match different rules, and can be validated by different validators.

**Expected behavior**

The registry host should be matched case-insensitively, so the same image maps to the same rule regardless of host case. The repository path should stay case-sensitive.

**Optional: To reproduce**

With rules `myregistry.io/secure/*` (strict validator) and `*` (a more permissive validator):

```go
c := Config{Rules: []policy.Rule{
    {Pattern: "myregistry.io/secure/*", Validator: "strict"},
    {Pattern: "*", Validator: "permissive"},
}}
r1, _ := c.MatchingRule("myregistry.io/secure/app:1") // -> strict
r2, _ := c.MatchingRule("MyRegistry.io/secure/app:1") // -> permissive
```

The uppercase-host reference skips the specific rule and falls through to the broader one. With the default policy (`*: deny`) this fails closed, so the impact depends on a configuration that pairs a specific rule with a more permissive catch-all.

**Optional: Versions**

- Connaisseur: `develop` at `0c40cab`

The fix is to case-fold the registry host (the component before the first `/`) of both the image and the patterns before matching, leaving the path untouched. I can open a PR with the change and a test.

kanywst:
Fixes #2130

## Description

Policy rules matched the registry host case-sensitively, so the same image referenced with a different-case host matched a different rule. Registry hosts are case-insensitive. This case-folds the registry host of both the image and the patterns before matching, leaving the repository path untouched.

## Checklist

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -76,13 +76,28 @@ func (c Config) Validator(key string) (*validator.Validator, error) {
 	return nil, fmt.Errorf("validator \"%s\" not found", key)
 }
 
+// foldHost lowercases only the registry host of a reference or pattern; the
+// first component counts as a host if it has "." or ":" or is "localhost".
+func foldHost(ref string) string {
+	i := strings.IndexByte(ref, '/')
+	if i < 0 {
+		return ref
+	}
+	host := ref[:i]
+	if !strings.ContainsAny(host, ".:") && host != "localhost" {
+		return ref
+	}
+	return strings.ToLower(host) + ref[i:]
+}
+
 // Gets the most specific matching rule for a given image from the config.
 func (c Config) MatchingRule(image string) (*policy.Rule, error) {
+	image = foldHost(image)
 	// start with empty pattern
 	bestMatch := policy.NewMatch(policy.Rule{}, "")
 
 	for _, rule := range c.Rules {
-		pattern := rule.Pattern
+		pattern := foldHost(rule.Pattern)
 
 		// prepend wildcard if not present
 		if !strings.HasPrefix(pattern, "*") {
--- a/internal/config/config_test.go
+++ b/internal/config/config_test.go
@@ -99,6 +99,8 @@ func TestMatchingRule(t *testing.T) {
 		},
 		{"my.registry/test:tag", "my.registry/test"},
 		{"my.registry/abc:tag", "my.registry/*"},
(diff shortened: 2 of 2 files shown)


Starkteetje:
Hi @kanywst 
thank you for your issue and contribution.

I agree that we can fix the image host and likely also images (IIRC docker repos and names need to be lower-case).
I'm not entirely sure, whether we should meddle with the patterns the user supplied (@phbelitz any thoughts?), but if we do, we definitely need to take care.

There's an issue with the implementation: We allow wildcard patterns, so `foldHost` of pattern `*:myTag` would lowercase the tag which is case-sensitive. Filtering after `:` in additon to `/` also won't work due to IPv6 addresses which could be part of the host, so I guess we'll need to parse the reference properly, likely best inside internal/image/image.go directly

kanywst:
@Starkteetje 
Good catches, fixed both. `foldHost` only folds the host now. Happy to move it into `internal/image` if you prefer; the pattern-normalization call I'll leave to you.

internal/config/config.go
```diff
@@ -76,13 +76,28 @@ func (c Config) Validator(key string) (*validator.Validator, error) {
...
 
+// foldHost lowercases only the registry host of a reference or pattern; the
+// first component counts as a host if it has "." or ":" or is "localhost".
+func foldHost(ref string) string {
+	i := strings.IndexByte(ref, '/')
+	if i < 0 {
+		return ref
+	}
+	host := ref[:i]
+	if !strings.ContainsAny(host, ".:") && host != "localhost" {
```
Starkteetje, line 87: This won't work for LocalHost since the check is case-sensitive

Starkteetje (commented):
I'll add a fix for the localhost thingy and open a new PR (#2137) with the cherrypicked changes, since our pipeline currently doesnt work for outside contributors

Starkteetje:
Merged in #2137 . Hopefully, we get to release today, in which case this would be live later

PR #2139, opened 2026-07-10
build: Remove hidden files from build context

Starkteetje:
## Description

Remove hidden files from build context

## Checklist

### DIFF
--- a/.dockerignore
+++ b/.dockerignore
@@ -13,3 +13,5 @@ helm/
 LICENSE
 Makefile
 mkdocs.yml
+# ignore hidden files/folders
+**/.?*


.dockerignore
```diff
@@ -13,3 +13,5 @@ helm/
 LICENSE
 Makefile
 mkdocs.yml
+# ignore hidden files/folders
+**/.?*
```
phbelitz, line 17: whta does the `?` do here? is it similar to regex, so i could math files with `.` prefixes, but doesn't have to?

Starkteetje: dockerignore regex isnt the same are normal regex. `.` is just the dot character and `?` is any character (same as . in normal regex)

So
`**/` - any directory depth
`.` - file or folder starts with `.`
`?` - then there is any character
`*` - then there are arbitrary many chars following

We want /. do be part of the context and with just `**/.*` our build command doesn't work as it uses current dir `.` as context, which is why we need to add something and that is just any arbitrary char

phbelitz (approved):

